PR #13, opened 2026-05-13
Add streaming timing metrics to generic stream wrappers

### LINKED ISSUE #8: Implement gen_ai.client.operation.time_to_first_chunk for OpenAI v2 streaming
## Problem

The OpenTelemetry GenAI Semantic Conventions define a `gen_ai.client.operation.time_to_first_chunk` histogram metric that records the time (in seconds) from request start to the first output chunk during streaming operations. No Python instrumentation currently implements this metric.

Originally tracked in opentelemetry-python-contrib #3932 with a previous implementation in opentelemetry-python-contrib#4415. That PR was closed when GenAI instrumentations moved to this repository.

## Context

In the old repository, #4415 was designed to build on top of opentelemetry-python-contrib #4500 (streaming ABC refactor by @eternalcuriouslearner), which introduced the generic stream wrapper in `opentelemetry-util-genai`. That refactor is already present here at `util/opentelemetry-util-genai/src/opentelemetry/util/genai/stream.py`, so the time-to-first-chunk metric can now be built directly on top of it.

cc @lmolkova @eternalcuriouslearner

## Proposed solution

1. Add a `gen_ai.client.operation.time_to_first_chunk` histogram to the OpenAI v2 instrumentation and the generic streaming ABC.
2. Record `time_of_first_chunk - start_time` in seconds when the first streaming chunk arrives.
3. Attach standard GenAI metric attributes: `gen_ai.operation.name`, `gen_ai.request.model`, `gen_ai.response.model`, `gen_ai.system`, `server.address`, `server.port`.
4. Use the semconv-defined explicit bucket boundaries.
5. Record only for streaming calls; do not emit when no chunk is ever received.
6. Cover both sync and async streaming paths.

## Acceptance criteria

- Metric recorded only on streaming completions (not non-streaming)
- No data point emitted if the stream errors before the first chunk
- Sync and async streaming paths both instrumented
- Test coverage ported from opentelemetry-python-contrib #4415

## References

- Semantic convention: https://opentelemetry.io/docs/specs/semconv/gen-ai/gen-ai-metrics/
- Original issue: https://github.com/open-telemetry/opentelemetry-python-contrib/issues/3932
- Previous implementation: https://github.com/open-telemetry/opentelemetry-python-contrib/pull/4415
- Streaming ABC refactor: https://github.com/open-telemetry/opentelemetry-python-contrib/pull/4500

I have a previous implementation from opentelemetry-python-contrib#4415 ready to port.

Nik-Reddy:
Closes #8

## What this does

Adds three streaming timing measurements to the shared utils layer:

1. `gen_ai.client.operation.time_to_first_chunk` histogram
2. `gen_ai.response.time_to_first_chunk` span attribute
3. `gen_ai.client.operation.time_per_output_chunk` histogram (one data point per inter-chunk gap)

All timing logic lives in the stream wrapper base classes (`SyncStreamWrapper`/`AsyncStreamWrapper` in `util/opentelemetry-util-genai`). The OpenAI chat wrappers just pass through the invocation start time and copy measured values back, keeping provider-specific code minimal.

## How timing works

- TTFC: measured from invocation start (when the HTTP request was issued) to when the first chunk arrives
- time_per_output_chunk: measures the blocking read duration for each chunk after the first, so user processing time between pulls doesn't inflate the metric
- Uses `timeit.default_timer()` around the blocking `next()`/`anext()` call

## Scope

This wires up timing for OpenAI chat completions streams (which use the shared ABC). Anthropic and the Responses API have their own stream wrappers that don't extend the ABC, so they will need separate follow-up work.

## Testing

- 5 unit tests for the timing logic in stream.py (sync, async, single chunk, error cases)
- 2 integration tests verifying histogram recording in metrics.py
- All existing tests still pass

### DIFF
--- a/instrumentation/opentelemetry-instrumentation-google-genai/tests/generate_content/cassettes/test_upload_hook_non_streaming[SPAN_AND_EVENT-vertexaiapi-sync-gemini-2.5-flash-enable_completion_hook].yaml
+++ b/instrumentation/opentelemetry-instrumentation-google-genai/tests/generate_content/cassettes/test_upload_hook_non_streaming[SPAN_AND_EVENT-vertexaiapi-sync-gemini-2.5-flash-enable_completion_hook].yaml
@@ -1,0 +1,102 @@
+interactions:
+- request:
+    body: |-
+      {
+        "contents": [
+          {
+            "parts": [
+              {
+                "text": "Create a haiku about Open Telemetry."
+              }
+            ],
+            "role": "user"
+          }
+        ]
+      }
+    headers:
+      accept:
+      - '*/*'
+      accept-encoding:
+      - identity
+      connection:
+      - keep-alive
+      content-length:
+      - '93'
+      content-type:
+      - application/json
+      host:
+      - us-central1-aiplatform.googleapis.com
+      user-agent:
+      - google-genai-sdk/1.32.0 gl-python/3.10.18
+      x-goog-api-client:
+      - <REDACTED>
+      x-goog-user-project:
+      - <REDACTED>
+    method: POST
+    uri: https://test-location-aiplatform.googleapis.com/v1beta1/projects/test-project/locations/test-location/publishers/google/models/gemini-2.5-flash:generateContent
+  response:
+    body:
+      string: |-
+        {
+          "candidates": [
+            {
+              "content": {
+                "role": "model",
+                "parts": [
+                  {
+                    "text": "Open data streams,\nMetrics, logs, and traces flow,\nClearly see inside."
+                  }
+                ]
+              },
+              "finishReason": "STOP",
+              "avgLogprobs": -10.948673672146267
+            }
+          ],
+          "usageMetadata": {
+            "promptTokenCount": 8,
+            "candidatesTokenCount": 18,
+            "totalTokenCount": 663,
+            "trafficType": "ON_DEMAND",
+            "promptTokensDetails": [
+              {
+                "modality": "TEXT",
+                "tokenCount": 8
+              }
+            ],
+            "candidatesTokensDetails": [
+              {
+                "modality": "TEXT",
+                "tokenCount": 18
+              }
+            ],
+            "thoughtsTokenCount": 637
+          },
+          "modelVersion": "gemini-2.5-flash",
+          "createTime": "2025-10-10T16:32:22.100847Z",
+          "responseId": "FjXpaO-TBoC4nvgPk77D6Ak"
+        }
+    headers:
+      Accept-Ranges:
+      - none
+      Content-Type:
+      - application/json; charset=UTF-8
+      Date:
+      - Fri, 10 Oct 2025 16:32:25 GMT
+      Server:
+      - scaffolding on HTTPServer2
+      Transfer-Encoding:
+      - chunked
+      Vary:
+      - X-Origin
+      - Referer
+      - Origin,Accept-Encoding
+      X-Content-Type-Options:
+      - nosniff
+      X-Frame-Options:
+      - SAMEORIGIN
+      X-XSS-Protection:
+      - '0'
+    status:
+      code: 200
+      message: OK
+version: 1
--- a/util/opentelemetry-util-genai/src/opentelemetry/util/genai/_inference_invocation.py
+++ b/util/opentelemetry-util-genai/src/opentelemetry/util/genai/_inference_invocation.py
@@ -4,7 +4,7 @@
 from __future__ import annotations
 
 from dataclasses import dataclass, field
-from typing import Any
+from typing import TYPE_CHECKING, Any
 
 from opentelemetry._logs import Logger, LogRecord
 from opentelemetry.semconv._incubating.attributes import (
@@ -18,7 +18,6 @@ from opentelemetry.util.genai._invocation import (
     get_content_attributes,
 )
 from opentelemetry.util.genai.completion_hook import CompletionHook
-from opentelemetry.util.genai.metrics import InvocationMetricsRecorder
 from opentelemetry.util.genai.types import (
     InputMessage,
     MessagePart,
@@ -30,8 +29,12 @@ from opentelemetry.util.genai.utils import (
     should_emit_event,
 )
 
+if TYPE_CHECKING:
+    from opentelemetry.util.genai.metrics import InvocationMetricsRecorder
+
 # TODO: Migrate to GenAI constants once available in semconv package
 _GEN_AI_REASONING_OUTPUT_TOKENS = "gen_ai.usage.reasoning.output_tokens"
+_GEN_AI_RESPONSE_TIME_TO_FIRST_CHUNK = "gen_ai.response.time_to_first_chunk"
 
 
 class InferenceInvocation(GenAIInvocation):
@@ -93,6 +96,9 @@ class InferenceInvocation(GenAIInvocation):
         self.cache_creation_input_tokens: int | None = None
         self.cache_read_input_tokens: int | None = None
         self.tool_definitions: list[ToolDefinition] | None = None
+        # Streaming timing fields (populated by stream wrappers)
+        self.ttfc_seconds: float | None = None
+        self.chunk_gap_seconds: list[float] = []
         self._start(self._get_base_attributes())
 
     def _get_message_attributes(self, *, for_span: bool) -> dict[str, Any]:
@@ -161,6 +167,10 @@ class InferenceInvocation(GenAIInvocation):
                 _GEN_AI_REASONING_OUTPUT_TOKENS,
                 self.thinking_tokens,
             ),
+            (
+                _GEN_AI_RESPONSE_TIME_TO_FIRST_CHUNK,
+                self.ttfc_seconds,
+            ),
         )
         attrs.update({k: v for k, v in optional_attrs if v is not None})
         return attrs
@@ -171,6 +181,29 @@ class InferenceInvocation(GenAIInvocation):
             attrs[GenAI.GEN_AI_RESPONSE_MODEL] = self.response_model_name
         attrs.update(self.metric_attributes)
         return attrs
+
+    def _record_chunk_gap(self, gap: float) -> None:
+        """Buffer a time-per-output-chunk gap (in seconds).
+
+        Called by the stream wrapper as each chunk after the first arrives.
+        Buffered gaps are drained by ``_consume_streaming_timing`` when the
+        invocation stops.
+        """
+        self.chunk_gap_seconds.append(gap)
+
+    def _consume_streaming_timing(
+        self,
+    ) -> tuple[float | None, list[float]]:
+        """Return TTFC and chunk gaps, then reset them on the invocation.
+
+        Called by InvocationMetricsRecorder so the timing values are emitted
+        once and not held past finalization.
+        """
+        ttfc = self.ttfc_seconds
+        gaps = self.chunk_gap_seconds
+        self.ttfc_seconds = None
+        self.chunk_gap_seconds = []
+        return ttfc, gaps
 
     def _get_metric_token_counts(self) -> dict[str, int]:
         counts: dict[str, int] = {}
--- a/util/opentelemetry-util-genai/src/opentelemetry/util/genai/instruments.py
+++ b/util/opentelemetry-util-genai/src/opentelemetry/util/genai/instruments.py
@@ -4,6 +4,14 @@
 from opentelemetry.metrics import Histogram, Meter
 from opentelemetry.semconv._incubating.metrics import gen_ai_metrics
 
+# TODO: Migrate to GenAI constants once available in semconv package
+_GEN_AI_CLIENT_OPERATION_TIME_TO_FIRST_CHUNK = (
+    "gen_ai.client.operation.time_to_first_chunk"
+)
+_GEN_AI_CLIENT_OPERATION_TIME_PER_OUTPUT_CHUNK = (
+    "gen_ai.client.operation.time_per_output_chunk"
+)
+
 _GEN_AI_CLIENT_OPERATION_DURATION_BUCKETS = [
     0.01,
     0.02,
@@ -54,4 +62,30 @@ def create_token_histogram(meter: Meter) -> Histogram:
         description="Number of input and output tokens used by GenAI clients",
         unit="{token}",
         explicit_bucket_boundaries_advisory=_GEN_AI_CLIENT_TOKEN_USAGE_BUCKETS,
+    )
+
+
+def create_ttfc_histogram(meter: Meter) -> Histogram:
+    return meter.create_histogram(
+        name=_GEN_AI_CLIENT_OPERATION_TIME_TO_FIRST_CHUNK,
+        description=(
+            "Time to receive the first chunk, measured from when the client "
+            "issues the generation request to when the first chunk is "
+            "received in the response stream."
+        ),
+        unit="s",
+        explicit_bucket_boundaries_advisory=_GEN_AI_CLIENT_OPERATION_DURATION_BUCKETS,
+    )
+
+
+def create_time_per_chunk_histogram(meter: Meter) -> Histogram:
+    return meter.create_histogram(
+        name=_GEN_AI_CLIENT_OPERATION_TIME_PER_OUTPUT_CHUNK,
+        description=(
+            "Time per output chunk, recorded for each chunk received after "
+            "the first one, measured as the time elapsed from the end of "
+            "the previous chunk to the end of the current chunk."
+        ),
+        unit="s",
+        explicit_bucket_boundaries_advisory=_GEN_AI_CLIENT_OPERATION_DURATION_BUCKETS,
     )
--- a/util/opentelemetry-util-genai/src/opentelemetry/util/genai/metrics.py
+++ b/util/opentelemetry-util-genai/src/opentelemetry/util/genai/metrics.py
@@ -14,9 +14,12 @@ from opentelemetry.semconv._incubating.attributes import (
 )
 from opentelemetry.util.genai.instruments import (
     create_duration_histogram,
+    create_time_per_chunk_histogram,
     create_token_histogram,
+    create_ttfc_histogram,
 )
 
+from ._inference_invocation import InferenceInvocation
 from ._invocation import GenAIInvocation
 
 
@@ -26,6 +29,10 @@ class InvocationMetricsRecorder:
     def __init__(self, meter: Meter):
         self._duration_histogram: Histogram = create_duration_histogram(meter)
         self._token_histogram: Histogram = create_token_histogram(meter)
+        self._ttfc_histogram: Histogram = create_ttfc_histogram(meter)
+        self._time_per_chunk_histogram: Histogram = (
+            create_time_per_chunk_histogram(meter)
+        )
 
     def record(self, invocation: GenAIInvocation) -> None:
         """Record duration and token metrics for an invocation if possible."""
@@ -33,9 +40,9 @@ class InvocationMetricsRecorder:
         token_counts = invocation._get_metric_token_counts()
 
         duration_seconds: Optional[float] = None
-        if invocation._monotonic_start_s is not None:
+        if invocation.monotonic_start_s is not None:
             duration_seconds = max(
-                timeit.default_timer() - invocation._monotonic_start_s,
+                timeit.default_timer() - invocation.monotonic_start_s,
                 0.0,
             )
 
@@ -52,6 +59,22 @@ class InvocationMetricsRecorder:
                 attributes=attributes | {GenAI.GEN_AI_TOKEN_TYPE: token_type},
                 context=invocation._span_context,
             )
+
+        # Streaming timing metrics
+        if isinstance(invocation, InferenceInvocation):
+            ttfc, gaps = invocation._consume_streaming_timing()
+            if ttfc is not None:
+                self._ttfc_histogram.record(
+                    ttfc,
+                    attributes=attributes,
+                    context=invocation._span_context,
+                )
+            for gap in gaps:
+                self._time_per_chunk_histogram.record(
+                    gap,
+                    attributes=attributes,
+                    context=invocation._span_context,
+                )
 
 
 __all__ = ["InvocationMetricsRecorder"]
--- a/util/opentelemetry-util-genai/src/opentelemetry/util/genai/stream.py
+++ b/util/opentelemetry-util-genai/src/opentelemetry/util/genai/stream.py
@@ -4,6 +4,7 @@
 from __future__ import annotations
 
 import logging
+import timeit
 from abc import ABCMeta, abstractmethod
 from types import TracebackType
 from typing import (
@@ -28,6 +29,22 @@ else:
 ChunkT = TypeVar("ChunkT")
 _ChunkT_co = TypeVar("_ChunkT_co", covariant=True)
 _logger = logging.getLogger(__name__)
+
+
+class _TimingTarget(Protocol):
+    """Internal protocol for objects that receive streaming timing.
+
+    This is the wrapper-to-invocation seam inside this package and is not
+    a public extension point; third parties should not implement it
+    directly. ``ttfc_seconds`` is set once by the wrapper at
+    finalization. Each inter-chunk gap is pushed through
+    ``_record_chunk_gap`` as soon as it is measured so the wrapper does
+    not need to buffer measurements.
+    """
+
+    ttfc_seconds: float | None
+
+    def _record_chunk_gap(self, gap: float) -> None: ...
 
 
 class _StreamWrapperMeta(ABCMeta, type(_ObjectProxy)):
@@ -64,11 +81,25 @@ class SyncStreamWrapper(
     internally by the wrapper lifecycle and are not part of the public API.
     """
 
-    def __init__(self, stream: _SyncStream[ChunkT]):
+    def __init__(
+        self,
+        stream: _SyncStream[ChunkT],
+        start_time_s: float | None = None,
+        timing_target: _TimingTarget | None = None,
+    ) -> None:
+        if timing_target is not None and start_time_s is None:
+            raise ValueError(
+                "start_time_s is required when timing_target is provided"
+            )
         super().__init__(stream)
         self._self_stream = stream
         self._self_iterator = iter(stream)
         self._self_finalized = False
+        self._self_start_time_s = start_time_s
+        self._self_timing_target = timing_target
+        self._self_ttfc_seconds: float | None = None
+        self._self_first_chunk_at: float | None = None
+        self._self_last_chunk_at: float | None = None
 
     def __enter__(self):
         return self
@@ -116,22 +147,57 @@ class SyncStreamWrapper(
         except Exception as error:
             self._safe_finalize_failure(error)
             raise
+
+        if (
+            self._self_start_time_s is not None
+            or self._self_timing_target is not None
+        ):
+            t_after_read = timeit.default_timer()
+            if self._self_first_chunk_at is None:
+                self._self_first_chunk_at = t_after_read
+                if self._self_start_time_s is not None:
+                    self._self_ttfc_seconds = (
+                        t_after_read - self._self_start_time_s
+                    )
+            else:
+                previous = self._self_last_chunk_at
+                if previous is not None:
+                    self._safe_record_chunk_gap(t_after_read - previous)
+            self._self_last_chunk_at = t_after_read
+
         try:
             self._process_chunk(chunk)
         except Exception as error:  # pylint: disable=broad-exception-caught
             self._handle_process_chunk_error(error)
         return chunk
 
+    def _safe_record_chunk_gap(self, gap: float) -> None:
+        if self._self_timing_target is None:
+            return
+        try:
+            self._self_timing_target._record_chunk_gap(gap)
+        except Exception:  # pylint: disable=broad-exception-caught
+            _logger.debug(
+                "GenAI stream instrumentation error recording chunk gap",
+                exc_info=True,
+            )
+
+    def _sync_timing_to_target(self) -> None:
+        if self._self_timing_target is not None:
+            self._self_timing_target.ttfc_seconds = self._self_ttfc_seconds
+
     def _finalize_success(self) -> None:
         if self._self_finalized:
             return
         self._self_finalized = True
+        self._sync_timing_to_target()
         self._on_stream_end()
 
     def _finalize_failure(self, error: BaseException) -> None:
         if self._self_finalized:
             return
         self._self_finalized = True
+        self._sync_timing_to_target()
         self._on_stream_error(error)
 
     def _safe_finalize_success(self) -> None:
@@ -191,11 +257,25 @@ class AsyncStreamWrapper(
     are owned by this base class.
     """
 
-    def __init__(self, stream: _AsyncStream[ChunkT]):
+    def __init__(
+        self,
+        stream: _AsyncStream[ChunkT],
+        start_time_s: float | None = None,
+        timing_target: _TimingTarget | None = None,
+    ) -> None:
+        if timing_target is not None and start_time_s is None:
+            raise ValueError(
+                "start_time_s is required when timing_target is provided"
+            )
         super().__init__(stream)
         self._self_stream = stream
         self._self_aiter = aiter(stream)
         self._self_finalized = False
+        self._self_start_time_s = start_time_s
+        self._self_timing_target = timing_target
+        self._self_ttfc_seconds: float | None = None
+        self._self_first_chunk_at: float | None = None
+        self._self_last_chunk_at: float | None = None
 
     async def __aenter__(self):
         return self
@@ -245,22 +325,57 @@ class AsyncStreamWrapper(
         except Exception as error:
             self._safe_finalize_failure(error)
             raise
+
+        if (
+            self._self_start_time_s is not None
+            or self._self_timing_target is not None
+        ):
+            t_after_read = timeit.default_timer()
+            if self._self_first_chunk_at is None:
+                self._self_first_chunk_at = t_after_read
+                if self._self_start_time_s is not None:
+                    self._self_ttfc_seconds = (
+                        t_after_read - self._self_start_time_s
+                    )
+            else:
+                previous = self._self_last_chunk_at
+                if previous is not None:
+                    self._safe_record_chunk_gap(t_after_read - previous)
+            self._self_last_chunk_at = t_after_read
+
         try:
             self._process_chunk(chunk)
         except Exception as error:  # pylint: disable=broad-exception-caught
             self._handle_process_chunk_error(error)
         return chunk
 
+    def _safe_record_chunk_gap(self, gap: float) -> None:
+        if self._self_timing_target is None:
+            return
+        try:
+            self._self_timing_target._record_chunk_gap(gap)
+        except Exception:  # pylint: disable=broad-exception-caught
+            _logger.debug(
+                "GenAI stream instrumentation error recording chunk gap",
+                exc_info=True,
+            )
+
+    def _sync_timing_to_target(self) -> None:
+        if self._self_timing_target is not None:
+            self._self_timing_target.ttfc_seconds = self._self_ttfc_seconds
+
     def _finalize_success(self) -> None:
         if self._self_finalized:
             return
         self._self_finalized = True
+        self._sync_timing_to_target()
         self._on_stream_end()
 
     def _finalize_failure(self, error: BaseException) -> None:
         if self._self_finalized:
             return
         self._self_finalized = True
+        self._sync_timing_to_target()
         self._on_stream_error(error)
 
     def _safe_finalize_success(self) -> None:
--- a/util/opentelemetry-util-genai/tests/test_handler_metrics.py
+++ b/util/opentelemetry-util-genai/tests/test_handler_metrics.py
@@ -393,6 +393,111 @@ class TelemetryHandlerToolMetricsTest(TestBase):
         self.assertAlmostEqual(duration_point.sum, 1.5, places=3)
         self.assertNotIn("gen_ai.client.token.usage", metrics)
 
+    def test_stop_inference_records_ttfc_and_chunk_gaps(self) -> None:
+        """Verify TTFC and time_per_chunk histograms are recorded for streaming."""
+        handler = TelemetryHandler(
+            tracer_provider=self.tracer_provider,
+            meter_provider=self.meter_provider,
+        )
+        with patch("timeit.default_timer", return_value=1000.0):
+            invocation = handler.start_inference("prov", request_model="model")
+
+        # Simulate streaming timing
+        invocation.ttfc_seconds = 0.35
+        invocation.chunk_gap_seconds = [0.05, 0.08, 0.12]
+
+        with patch("timeit.default_timer", return_value=1002.0):
+            invocation.stop()
+
+        # Timing fields are reset after recording so they aren't double-counted
+        self.assertIsNone(invocation.ttfc_seconds)
+        self.assertEqual(invocation.chunk_gap_seconds, [])
+
+        metrics = self._harvest_metrics()
+
+        # TTFC histogram
+        self.assertIn("gen_ai.client.operation.time_to_first_chunk", metrics)
+        ttfc_points = metrics["gen_ai.client.operation.time_to_first_chunk"]
+        self.assertEqual(len(ttfc_points), 1)
+        self.assertAlmostEqual(ttfc_points[0].sum, 0.35, places=3)
+        self.assertEqual(
+            ttfc_points[0].attributes[GenAI.GEN_AI_OPERATION_NAME],
+            GenAI.GenAiOperationNameValues.CHAT.value,
+        )
+        self.assertEqual(
+            ttfc_points[0].attributes[GenAI.GEN_AI_PROVIDER_NAME], "prov"
+        )
+        self.assertEqual(
+            ttfc_points[0].attributes[GenAI.GEN_AI_REQUEST_MODEL], "model"
+        )
+
+        # Time per chunk histogram (one record per gap)
+        self.assertIn("gen_ai.client.operation.time_per_output_chunk", metrics)
+        chunk_points = metrics["gen_ai.client.operation.time_per_output_chunk"]
+        self.assertEqual(len(chunk_points), 1)
+        # Sum of all gaps: 0.05 + 0.08 + 0.12 = 0.25
+        self.assertAlmostEqual(chunk_points[0].sum, 0.25, places=3)
+        # Count should be 3 (one per gap)
+        self.assertEqual(chunk_points[0].count, 3)
+        self.assertEqual(
+            chunk_points[0].attributes[GenAI.GEN_AI_OPERATION_NAME],
+            GenAI.GenAiOperationNameValues.CHAT.value,
+        )
+        self.assertEqual(
+            chunk_points[0].attributes[GenAI.GEN_AI_PROVIDER_NAME], "prov"
+        )
+        self.assertEqual(
+            chunk_points[0].attributes[GenAI.GEN_AI_REQUEST_MODEL], "model"
+        )
+
+    def test_ttfc_span_attribute_recorded(self) -> None:
+        """Verify gen_ai.response.time_to_first_chunk lands on the span."""
+        handler = TelemetryHandler(
+            tracer_provider=self.tracer_provider,
+            meter_provider=self.meter_provider,
+        )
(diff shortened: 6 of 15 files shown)


Nik-Reddy:
@lmolkova This implements the three items you mentioned in https://github.com/open-telemetry/opentelemetry-python-genai/issues/8:
- gen_ai.response.time_to_first_chunk span attribute
- gen_ai.client.operation.time_to_first_chunk histogram
- gen_ai.client.operation.time_per_output_chunk histogram (one record per inter-chunk gap)

For chunk gaps I'm measuring blocking read time rather than wall clock between returns, so user-side processing doesn't inflate it. Let me know if that's not what you had in mind.

Anthropic/Responses API don't use the ABC yet so they'll need follow-up work in a separate PR.

util/opentelemetry-util-genai/src/opentelemetry/util/genai/_inference_invocation.py
```diff
@@ -162,6 +166,10 @@ def _get_attributes(self) -> dict[str, Any]:
                 _GEN_AI_REASONING_OUTPUT_TOKENS,
                 self.thinking_tokens,
             ),
+            (
+                _GEN_AI_RESPONSE_TIME_TO_FIRST_CHUNK,
+                self.ttfc_seconds,
```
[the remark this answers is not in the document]
Nik-Reddy: Added in `2dbb54d`

util/opentelemetry-util-genai/src/opentelemetry/util/genai/stream.py
```diff
@@ -64,11 +65,15 @@ class SyncStreamWrapper(
...
-    def __init__(self, stream: _SyncStream[ChunkT]):
+    def __init__(self, stream: _SyncStream[ChunkT], start_time_s: float | None = None):
         super().__init__(stream)
         self._self_stream = stream
         self._self_iterator = iter(stream)
         self._self_finalized = False
+        self._self_start_time_s = start_time_s
+        self._self_ttfc_seconds: float | None = None
+        self._self_first_chunk_at: float | None = None
+        self._self_chunk_gaps: list[float] = []
```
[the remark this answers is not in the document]
Nik-Reddy: Should be an acceptable tradeoff for the simplicity it gives us.

instrumentation/opentelemetry-instrumentation-openai-v2/src/opentelemetry/instrumentation/openai_v2/chat_wrappers.py
```diff
@@ -164,6 +164,10 @@ def _cleanup(self, error: Optional[BaseException] = None) -> None:
                 },
             )
 
+        # Copy streaming timing from base wrapper to invocation
+        self._self_invocation.ttfc_seconds = self._self_ttfc_seconds
```
lmolkova: we should try to find means to do it inside stream helpers in utils and not inside individual instrumentation libraries.

Nik-Reddy: @lmolkova moved this into the stream helpers. The ABC now accepts an optional timing_target and syncs timing values before calling `_on_stream_end/_on_stream_error`, so providers just pass it in the constructor and don't touch timing.

util/opentelemetry-util-genai/src/opentelemetry/util/genai/instruments.py
```diff
@@ -55,3 +63,21 @@ def create_token_histogram(meter: Meter) -> Histogram:
...
+        description="Time from request start to first output chunk",
+        unit="s",
+        explicit_bucket_boundaries_advisory=_GEN_AI_CLIENT_OPERATION_DURATION_BUCKETS,
+    )
+
+
+def create_time_per_chunk_histogram(meter: Meter) -> Histogram:
+    return meter.create_histogram(
+        name=_GEN_AI_CLIENT_OPERATION_TIME_PER_OUTPUT_CHUNK,
+        description="Time between consecutive output chunks after the first",
```
MikeGoldsmith: I don't think this correctly describes the [semconv metric](https://github.com/open-telemetry/semantic-conventions-genai/blob/e68d2d3114ed89f56313f2d3c316d68f3c757fbc/model/gen-ai/metrics.yaml#L60). This says the time between output chunks, but semconv describes it as time between output chunks specifically end of the last to end of the current one.

util/opentelemetry-util-genai/src/opentelemetry/util/genai/metrics.py
```diff
@@ -53,5 +59,23 @@ def record(self, invocation: GenAIInvocation) -> None:
                 context=invocation._span_context,
             )
 
+        # Streaming timing metrics
```
MikeGoldsmith, line 63: Why do we need this intermediate values? Can we just pass the recordings directly into the histogram so we're not holding them separately? Also, it doesn't look like we clean this up after recording, directly recording measurements would solve this.

instrumentation/opentelemetry-instrumentation-openai-v2/src/opentelemetry/instrumentation/openai_v2/chat_wrappers.py
```diff
@@ -203,7 +203,7 @@ def __init__(
         invocation: InferenceInvocation,
         capture_content: bool,
     ) -> None:
-        super().__init__(stream)
+        super().__init__(stream, start_time_s=invocation._monotonic_start_s, timing_target=invocation)
```
MikeGoldsmith: `invocation._monotonic_start_s` is private - we should either expose it publicly or have the wrapper read it from the timing target instead

MikeGoldsmith (changes requested):
Looks mostly okay, I've left a few of comments we should resolve before accepting.

util/opentelemetry-util-genai/src/opentelemetry/util/genai/instruments.py
Nik-Reddy: Updated the description to "**Duration from the end of the previous output chunk to the end of the current output chunk**" to align with the semconv definition.

instrumentation/opentelemetry-instrumentation-openai-v2/src/opentelemetry/instrumentation/openai_v2/chat_wrappers.py
Nik-Reddy: Added a read-only `monotonic_start_s` property on `GenAIInvocation` and switched both chat wrappers and the metrics recorder over to it.

util/opentelemetry-util-genai/src/opentelemetry/util/genai/metrics.py
Nik-Reddy: Added **`.clear()`** on the list right after recording so nothing lingers past finalization. 

For recording directly per-chunk, the stream ABC doesn't have access to the histogram or metric attributes, so wiring that in would mean coupling the generic stream layer to OTel metrics. Felt like a bigger change than this PR warrants given the list is bounded by stream lifetime (10's to low 100's of entries for typical LLM responses). 

Open to revisiting if you'd prefer that approach though

Nik-Reddy:
@MikeGoldsmith resolved all review thread, can you re-review when you have a moment.

util/opentelemetry-util-genai/src/opentelemetry/util/genai/metrics.py
```diff
@@ -53,5 +59,25 @@ def record(self, invocation: GenAIInvocation) -> None:
...
+        ttfc_seconds = getattr(invocation, "ttfc_seconds", None)
```
eternalcuriouslearner: question(blocking): Can we use is instance here to improve duck typing? Something like 

```
   if isinstance(invocation, InferenceInvocation):                                                                                                                                                     
          if invocation.ttfc_seconds is not None:
              self._ttfc_histogram.record(
                  invocation.ttfc_seconds,                                                                                                                                                                
                  attributes=attributes,
                  context=invocation._span_context,                                                                                                                                                       
              )   
          if invocation.chunk_gap_seconds:
              for gap in invocation.chunk_gap_seconds:
                  self._time_per_chunk_histogram.record(                                                                                                                                                  
                      gap,
                      attributes=attributes,                                                                                                                                                              
                      context=invocation._span_context,
                  )                                                                                                                                                                                       
              invocation.chunk_gap_seconds.clear()
```

instrumentation/opentelemetry-instrumentation-openai-v2/src/opentelemetry/instrumentation/openai_v2/chat_wrappers.py
```diff
@@ -182,7 +182,7 @@ def __init__(
         invocation: InferenceInvocation,
         capture_content: bool,
     ) -> None:
-        super().__init__(stream)
+        super().__init__(stream, start_time_s=invocation.monotonic_start_s, timing_target=invocation)
```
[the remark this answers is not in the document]
Nik-Reddy: Added **`test_chat_completion_streaming_metrics`** covering the sync **ChatStreamWrapper** through a VCR cassette. Asserts TTFC and per-output-chunk get recorded with the full attribute set (operation_name, provider_name,
 request_model, response_model, server_address). Dropping `timing_target=invocation` here would now fail this test.
 
 Skipped on the legacy semconv path, same pattern as the existing stream-wrapper-only tests.

instrumentation/opentelemetry-instrumentation-openai-v2/src/opentelemetry/instrumentation/openai_v2/chat_wrappers.py
```diff
@@ -203,7 +203,7 @@ def __init__(
...
```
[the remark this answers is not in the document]
Nik-Reddy: Added `test_async_chat_completion_streaming_metrics` for the async wrapper. Same shape as the sync one but the **async__init__** has separate wiring so it gets its own test

util/opentelemetry-util-genai/tests/test_handler_metrics.py
```diff
@@ -392,3 +392,96 @@ def test_fail_tool_records_duration_with_error(self) -> None:
         )
         self.assertAlmostEqual(duration_point.sum, 1.5, places=3)
         self.assertNotIn("gen_ai.client.token.usage", metrics)
+
+    def test_stop_inference_records_ttfc_and_chunk_gaps(self) -> None:
+        """Verify TTFC and time_per_chunk histograms are recorded for streaming."""
```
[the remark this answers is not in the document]
Nik-Reddy: Expanded the existing util test to assert provider_name and request_model on the TTFC point, plus the full operation_name/provider_name/request_model set on the per-chunk point. Was checking too little before

util/opentelemetry-util-genai/src/opentelemetry/util/genai/stream.py
```diff
@@ -116,22 +135,38 @@ def __next__(self) -> ChunkT:
...
             self._safe_finalize_failure(error)
             raise
+        t_after_read = timeit.default_timer()
+
+        if self._self_first_chunk_at is None:
+            self._self_first_chunk_at = t_after_read
+            if self._self_start_time_s is not None:
+                self._self_ttfc_seconds = t_after_read - self._self_start_time_s
+        else:
+            self._self_chunk_gaps.append(t_after_read - t_before_read)
```
eternalcuriouslearner: question (blocking): `time_per_output_chunk` measures blocking-read time, not inter-chunk gap — is that intentional?   I believe `time_per_output_chunk` should measure inter-chunk gap. @lmolkova @MikeGoldsmith can you please confirm if I am wrong?

util/opentelemetry-util-genai/src/opentelemetry/util/genai/stream.py
```diff
@@ -64,11 +72,21 @@ class SyncStreamWrapper(
     internally by the wrapper lifecycle and are not part of the public API.
     """
 
-    def __init__(self, stream: _SyncStream[ChunkT]):
+    def __init__(
+        self,
+        stream: _SyncStream[ChunkT],
+        start_time_s: float | None = None,
+        timing_target: _TimingTarget | None = None,
+    ):
```
eternalcuriouslearner: nit(blocking): Can we add a return signature here? say something like:

```
    def __init__(
        self,
        stream: _SyncStream[ChunkT],
        start_time_s: float | None = None,
        timing_target: _TimingTarget | None = None,
    ):
```

instrumentation/opentelemetry-instrumentation-openai-v2/src/opentelemetry/instrumentation/openai_v2/chat_wrappers.py
eternalcuriouslearner: question(blocking):  should TTFC start from invocation creation or from the SDK call site?

util/opentelemetry-util-genai/src/opentelemetry/util/genai/metrics.py
```diff
@@ -53,5 +59,25 @@ def record(self, invocation: GenAIInvocation) -> None:
...
+        chunk_gap_seconds = getattr(invocation, "chunk_gap_seconds", None)
+        if chunk_gap_seconds:
+            for gap in chunk_gap_seconds:
+                self._time_per_chunk_histogram.record(
+                    gap,
+                    attributes=attributes,
+                    context=invocation._span_context,
+                )
+            # Clear after recording to avoid holding timing data beyond finalization
+            chunk_gap_seconds.clear()
```
eternalcuriouslearner: question(blocking): On second thought, can we move the entire retrieval process to _inference_invocation as a getter. Say something like:

``` 
  def _consume_streaming_timing(self) -> tuple[float | None, list[float]]:                                                                                                                                
      """Return TTFC and chunk gaps, then reset them.
                                                                                                                                                                                                          
      Recorder pattern: caller gets the values to emit, invocation drops them
      so they aren't double-counted if anything else inspects the invocation.                                                                                                                             
      """                                 
      gaps = self.chunk_gap_seconds                                                                                                                                                                       
      ttfc = self.ttfc_seconds                                                                                                                                                                            
      self.chunk_gap_seconds = []         
      self.ttfc_seconds = None                                                                                                                                                                            
      return ttfc, gaps
```

then modify this as 

```
  ttfc, gaps = invocation._consume_streaming_timing()                                                                                                                                                     
  if ttfc is not None:                        
      self._ttfc_histogram.record(ttfc, attributes=...)                                                                                                                                                   
  for gap in gaps:
      self._time_per_chunk_histogram.record(gap, attributes=...) 
```

wdyt?

eternalcuriouslearner (changes requested):
LGTM!! Couple of changes and a dumb performance question:

1. Can we please add a VCR backed test to verify the span and metrics?
2. Can we extend the tests we have for sync wrapper to async wrapper?
3. Any idea on how large a `chunk_gap_seconds` can grow for long running streams or multi turn conversations?

aabmass:

util/opentelemetry-util-genai/src/opentelemetry/util/genai/metrics.py
Nik-Reddy: Sure. The recorder now branches on `isinstance(invocation, InferenceInvocation)` and pulls timing via the new `_consume_streaming_timing()` getter rather than getattr duck typing. Pairs with the change for the line 80 thread.

util/opentelemetry-util-genai/src/opentelemetry/util/genai/stream.py
Nik-Reddy: Added _consume_streaming_timing() on InferenceInvocation:
 ```
def _consume_streaming_timing(self) -> tuple[float | None, list[float]]:
    ttfc = self.ttfc_seconds
    gaps = self.chunk_gap_seconds
    self.ttfc_seconds = None
    self.chunk_gap_seconds = []
    return ttfc, gaps
```
 The recorder calls it inside the isinstance(invocation, InferenceInvocation) branch, so values are emitted once and reset on the `invocation. _apply_finish` reads ttfc_seconds for the span attribute before the recorder runs, so the span value still lands. The InvocationMetricsRecorder import in `_inference_invocation.py` moves under TYPE_CHECKING to break the cycle the runtime import would have introduced. Test test_stop_inference_records_ttfc_and_chunk_gaps now also asserts both fields are cleared after stop()

File: util/opentelemetry-util-genai/src/opentelemetry/util/genai/stream.py (hunk 12 above)
Comment (Nik-Reddy) [reply util/opentelemetry-util-genai/src/opentelemetry/util/genai/stream.py]: Added -> None on both `SyncStreamWrapper.__init__` and `AsyncStreamWrapper.__init__`

File: instrumentation/opentelemetry-instrumentation-openai-v2/src/opentelemetry/instrumentation/openai_v2/chat_wrappers.py (hunk 8 above)
Comment (Nik-Reddy) [reply instrumentation/opentelemetry-instrumentation-openai-v2/src/opentelemetry/instrumentation/openai_v2/chat_wrappers.py]: Correct. The current implementation appends `t_after_read - t_before_read`, which is the blocking read time for that chunk, not the gap between consecutive chunk arrivals. The semconv definition for `gen_ai.client.operation.time_per_output_chunk` is "**Duration from the end of the previous output chunk to the end of the current output chunk**", which in code is` t_after_read[n] - t_after_read[n-1]`
 
 I will hold off changing the math here until @lmolkova / @MikeGoldsmith confirm the intent, since the metric name and the description both leave a little room. Once confirmed, the fix is local to` __next__ / __anext__` and looks like this:
 ```
if self._self_first_chunk_at is None:
    self._self_first_chunk_at = t_after_read
    if self._self_start_time_s is not None:
        self._self_ttfc_seconds = t_after_read - self._self_start_time_s
else:
    self._self_chunk_gaps.append(t_after_read - self._self_last_chunk_at)
self._self_last_chunk_at = t_after_read
 ```
 with a new self._self_last_chunk_at: float | None = None in __init__. Tests in test_stream.py would also need their gap expectations updated to `t_after_read[n] - t_after_read[n-1]`

File: util/opentelemetry-util-genai/src/opentelemetry/util/genai/metrics.py (hunk 13 above)
Comment (Nik-Reddy) [reply util/opentelemetry-util-genai/src/opentelemetry/util/genai/metrics.py]: Added it almost exactly as you mentioned:
 ```
def _consume_streaming_timing(self) -> tuple[float | None, list[float]]:
    """Return TTFC and chunk gaps, then reset them on the invocation.

    Called by InvocationMetricsRecorder so the timing values are emitted
    once and not held past finalization.
    """
    ttfc = self.ttfc_seconds
    gaps = self.chunk_gap_seconds
    self.ttfc_seconds = None
    self.chunk_gap_seconds = []
    return ttfc, gaps
 ```
 Recorder is now:
```
if isinstance(invocation, InferenceInvocation):
    ttfc, gaps = invocation._consume_streaming_timing()
    if ttfc is not None:
        self._ttfc_histogram.record(ttfc, attributes=attributes, context=invocation._span_context)
    for gap in gaps:
        self._time_per_chunk_histogram.record(gap, attributes=attributes, context=invocation._span_context)
 ```
 One thing worth flagging: `_apply_finish reads ttfc_seconds` for the span attribute before the recorder runs, so the span value still lands, the reset only affects anything that inspects the invocation after stop(). Extended `test_stop_inference_records_ttfc_and_chunk_gaps` to assert both fields are cleared. The InvocationMetricsRecorder import in ` _inference_invocation.py` had to move under TYPE_CHECKING to avoid the circular import the new metrics -> _inference_invocation direction introduced.

instrumentation/opentelemetry-instrumentation-openai-v2/src/opentelemetry/instrumentation/openai_v2/chat_wrappers.py
MikeGoldsmith: I think this and the comment above have got mixed up.

For TTFC, I think measuring from invocation creation is fine. The intent is operator-facing latency ("how long until the user sees the first chunk"), and the time between `start_inference()` and the wrapped SDK call is likely negligible.

With the constraint, instrumentations should not do meaningful work between those two points — anything heavier than building a few attribute dicts would silently inflate TTFC. It's probably worth a comment near `monotonic_start_s` to make that contract explicit.

util/opentelemetry-util-genai/src/opentelemetry/util/genai/stream.py
MikeGoldsmith: I think `time_per_output_chunk` should follow the semconv definition — `t_after_read[n] - t_after_read[n-1]` (end of previous chunk to end of current chunk), not the current `t_after_read - t_before_read` (blocking read duration).

I'd suggest two things, part of what I suggested before (remove list of measurements):

1. Fix the math to match semconv — `gap = t_after_read - self._self_last_chunk_at`, tracking `_self_last_chunk_at` instead of using `t_before_read`.

2. Drop the wrapper-side list — instead of appending into `self._self_chunk_gaps`, push each gap to the timing target as it's computed. Add a `_record_chunk_gap(gap)` method on the `_TimingTarget` protocol; the invocation buffers internally and `_consume_streaming_timing()` continues to drain at `stop()`.

After both, the wrapper holds O(1) state (`_self_last_chunk_at`, `_self_first_chunk_at`, `_self_ttfc_seconds`), the buffering lives in one place (the invocation), and the math matches the spec description.

Tests in `test_stream.py` will need their gap expectations updated to the inter-chunk math.

instrumentation/opentelemetry-instrumentation-openai-v2/src/opentelemetry/instrumentation/openai_v2/chat_wrappers.py
eternalcuriouslearner: I believe the gist of the discussion is we are in right direction and I agree with @MikeGoldsmith a comment explaining the intention might be helpful in understanding the implementation.

util/opentelemetry-util-genai/src/opentelemetry/util/genai/stream.py
Nik-Reddy: Done caa8e5f. Wrapper state is now O(1): 
- **_self_last_chunk_at, _self_first_chunk_at, _self_ttfc_seconds**. Each gap is computed as **t_after_read - self._self_last_chunk_at** (so the first chunk emits no gap, only TTFC) and pushed to the invocation through the `new _record_chunk_gap` method on the _TimingTarget protocol. 
- InferenceInvocation implements it as a list append, `_consume_streaming_timing()` still drains the buffer at stop. `test_stream.py` was rebased to inter-chunk math and now asserts exact gap values on the target instead of length-only checks. 
- The wrapper-side `_self_chunk_gaps` list is gone. Also wrapped the target call in a small `_safe_record_chunk_gap` helper that logs at debug on exception, to keep the "telemetry must not raise" invariant the prior list-append behaviour gave us.
 
Happy to drop the helper and inline the **try/except** at the call site if you prefer @MikeGoldsmith

instrumentation/opentelemetry-instrumentation-openai-v2/src/opentelemetry/instrumentation/openai_v2/chat_wrappers.py
Nik-Reddy: @MikeGoldsmith agreed, kept the TTFC anchor at invocation creation. Added the explicit contract on **monotonic_start_s in _invocation.py** as part of caa8e5f, the docstring now states that the anchor is invocation creation and instrumentations must not perform meaningful work between **start_inference()** and the wrapped SDK call, so TTFC stays close to the user-facing first-chunk latency. 

Let me know if you would rather see the same note duplicated in **chat_wrappers.py** near the wrap site.

Nik-Reddy: @eternalcuriouslearner The wrapper itself no longer holds the list, after caa8e5f it carries O(1) state (**_self_last_chunk_at, _self_first_chunk_at, _self_ttfc_seconds**). The buffer now lives on **InferenceInvocation.chunk_gap_seconds**, which has one float per output chunk for the current inference and is drained by **_consume_streaming_timing() at stop(),** where the values are recorded into the histogram and discarded. So:
  - Per-stream upper bound: N floats, where N is the number of chunks the SDK yields for that one call. Each float is 8 bytes, so even a 10k-chunk response is roughly 80 KB held until stop, then released.
  - Multi-turn: there is no carry-over between invocations. Each new InferenceInvocation starts with an empty list, and the previous invocation's list is dropped after **_consume_streaming_timing()** runs at stop.
 
 If we ever care about an SDK that streams indefinitely without terminating, we could move to a rolling histogram-record on every chunk and skip the buffer entirely, but that would push the histogram-record cost into the read loop and would not match the current "record at stop" pattern used by the rest of the metrics. Happy to switch to streaming recording if you prefer.
 
 For items 1 and 2 (VCR-backed span/metrics test and extending sync coverage to async): both are good asks. The current PR keeps the test surface to the wrapper unit tests plus the existing openai-v2 chat metrics tests. **I would prefer to do the VCR-backed end-to-end test and the async parity sweep as a small follow-up PR so this one stays focused on the semconv math change**. Open to doing them here if you would rather not split.

Nik-Reddy:
> FYI if your PR needs a changelog entry, please update it to use towncrier #16

@aabmass Migrated the entries to towncrier news fragments at `util/opentelemetry-util-genai/.changelog/13.added` and `instrumentation/opentelemetry-instrumentation-openai-v2/.changelog/13.added`, and removed the duplicate Unreleased lines from the two CHANGELOG.md files

util/opentelemetry-util-genai/src/opentelemetry/util/genai/stream.py
```diff
@@ -191,11 +246,21 @@ class AsyncStreamWrapper(
     are owned by this base class.
     """
 
-    def __init__(self, stream: _AsyncStream[ChunkT]):
+    def __init__(
+        self,
+        stream: _AsyncStream[ChunkT],
+        start_time_s: float | None = None,
```
lzchen, line 263: Should `start_time_s` be mandatory? TTTC doesn't make much sense without a start time.

util/opentelemetry-util-genai/src/opentelemetry/util/genai/stream.py
```diff
@@ -30,6 +31,19 @@ def __init__(self, wrapped: object) -> None: ...
...
+    ``ttfc_seconds`` is set once by the wrapper at finalization. Each
+    inter-chunk gap is pushed to ``_record_chunk_gap`` as soon as it is
+    measured so the wrapper does not need to buffer measurements.
+    """
+
+    ttfc_seconds: float | None
+
+    def _record_chunk_gap(self, gap: float) -> None: ...
+
+
```
lzchen, line 49: Shouldn't `_consume_streaming_timing` be defined here as well?

util/opentelemetry-util-genai/src/opentelemetry/util/genai/stream.py
```diff
@@ -30,6 +31,19 @@ def __init__(self, wrapped: object) -> None: ...
 _logger = logging.getLogger(__name__)
 
 
+class _TimingTarget(Protocol):
```
lzchen, line 34: Does this mean we need all invocations to implement this? Shouldn't we just include this as part of `GenAIInvocation`?

util/opentelemetry-util-genai/src/opentelemetry/util/genai/_inference_invocation.py
```diff
@@ -172,6 +182,29 @@ def _get_metric_attributes(self) -> dict[str, Any]:
         attrs.update(self.metric_attributes)
         return attrs
 
+    def _record_chunk_gap(self, gap: float) -> None:
```
lzchen, line 185: Will this logic be pretty similar across different invocations? Probably can abstract this?

util/opentelemetry-util-genai/src/opentelemetry/util/genai/instruments.py
```diff
@@ -55,3 +63,21 @@ def create_token_histogram(meter: Meter) -> Histogram:
         unit="{token}",
         explicit_bucket_boundaries_advisory=_GEN_AI_CLIENT_TOKEN_USAGE_BUCKETS,
     )
+
+
+def create_ttfc_histogram(meter: Meter) -> Histogram:
+    return meter.create_histogram(
+        name=_GEN_AI_CLIENT_OPERATION_TIME_TO_FIRST_CHUNK,
+        description="Time from request start to first output chunk",
```
lzchen: Nit: `Time to receive the first chunk, measured from when the client issues the generation request to when the first chunk is received in the response stream.`

util/opentelemetry-util-genai/src/opentelemetry/util/genai/instruments.py
```diff
@@ -55,3 +63,21 @@ def create_token_histogram(meter: Meter) -> Histogram:
...
+        unit="s",
+        explicit_bucket_boundaries_advisory=_GEN_AI_CLIENT_OPERATION_DURATION_BUCKETS,
+    )
+
+
+def create_time_per_chunk_histogram(meter: Meter) -> Histogram:
+    return meter.create_histogram(
+        name=_GEN_AI_CLIENT_OPERATION_TIME_PER_OUTPUT_CHUNK,
+        description="Duration from the end of the previous output chunk to the end of the current output chunk",
```
lzchen: `Time per output chunk, recorded for each chunk received after the first one, measured as the time elapsed from the end of the previous chunk to the end of the current chunk.`

util/opentelemetry-util-genai/src/opentelemetry/util/genai/_inference_invocation.py
```diff
@@ -172,6 +182,29 @@ def _get_metric_attributes(self) -> dict[str, Any]:
...
+        """Buffer a time-per-output-chunk gap (in seconds).
+
+        Called by the stream wrapper as each chunk after the first arrives.
+        Buffered gaps are drained by ``_consume_streaming_timing`` when the
+        invocation stops.
+        """
+        self.chunk_gap_seconds.append(gap)
```
lzchen, line 192: Seems like there's no limit to this unbounded list, which might be a concern if the stream is really long. Can we record to the histogram directly every time instead?

util/opentelemetry-util-genai/src/opentelemetry/util/genai/stream.py
Nik-Reddy: Turned it into a constructor check: if `timing_target` is set, `start_time_s` has to be set too, otherwise the constructor raises `ValueError`. That way a misconfiguration fails loudly instead of silently emitting no TTFC, and the wrapper still works as a plain chunk-processing wrapper when neither arg is passed. Added tests for the new failure mode on both Sync and Async wrappers.

util/opentelemetry-util-genai/src/opentelemetry/util/genai/stream.py
Nik-Reddy: `_TimingTarget` is the write side that the stream wrapper uses to push timing into an invocation: ttfc_seconds gets set once at first chunk and chunk gaps are pushed through `_record_chunk_gap`. `_consume_streaming_timing` is the read side, called by InvocationMetricsRecorder at finalize to drain the buffered values. The wrapper never calls it, so adding it to the Protocol would advertise a method the only caller of the Protocol does not need.
 
 I can add it if you want both sides of the contract documented in one place, but on its own it would be unused by the wrapper.

util/opentelemetry-util-genai/src/opentelemetry/util/genai/stream.py
Nik-Reddy: Two reasons it is a Protocol today:
  1. Keeps `stream.py` off the concrete GenAIInvocation. The wrapper only needs an object that can receive ttfc_seconds and accept _record_chunk_gap, nothing else from the invocation surface.
  2. Non-streaming invocation types (embeddings, tools, workflows) would have to carry unused ttfc_seconds / chunk-gap state if those fields lived on the base. With the Protocol, only invocations that actually stream declare it.
 
 If you want it on GenAIInvocation anyway, I can move it in a follow-up. Trade is two extra fields on every invocation (None / empty on non-streaming ones) for a single place to document the contract. Let me know which way you want it.

util/opentelemetry-util-genai/src/opentelemetry/util/genai/_inference_invocation.py
Nik-Reddy: Same thread as `_TimingTarget` above. If we lift the streaming timing state up to GenAIInvocation, both `_record_chunk_gap` and `_consume_streaming_timing` move with it and the Protocol goes away. 
 
 If we keep the Protocol, those methods stay on InferenceInvocation since that is the only invocation that streams today. Will follow whichever direction you pick on the thread above and batch it with the next push.

util/opentelemetry-util-genai/src/opentelemetry/util/genai/_inference_invocation.py
Nik-Reddy: One snag with inline recording: `gen_ai.response.model`  is only set after the first response chunk arrives, so gaps between chunks 1 and 2 would emit without it while later gaps have it (same
 stream split across two attribute keys). 

Options I see:
 - 1. Inline + drop `response.model` from this metric.
-  2. Inline + snapshot attributes at first-chunk time.
-  3. Sliding-window cap on the buffer, keep finalize record.
-  4. Keep as is, document the memory shape.

util/opentelemetry-util-genai/src/opentelemetry/util/genai/instruments.py
Nik-Reddy: Applied verbatim in 78917c6, thanks.

util/opentelemetry-util-genai/src/opentelemetry/util/genai/instruments.py
Nik-Reddy: Applied as-is in 78917c6

Nik-Reddy:
@MikeGoldsmith @eternalcuriouslearner rebased this PR onto main (eaca532) to clear the conflict from PR #60, then squashed the history into a single commit (6f465ca) so the diff is easier to scan. CI is green across all 82 checks on the new head.
 
 The single conflict was the OpenAI changelog file. PR #60 renamed `opentelemetry-instrumentation-openai-v2/` to `opentelemetry-instrumentation-genai-openai/`, and my changelog entry was sitting at the old path. Moved it to `instrumentation/opentelemetry-instrumentation-genai-openai/.changelog/13.added` and dropped the now-redundant `v2` wording from the entry text. No behavior changes from the rebase, and all prior thread resolutions are preserved.
 
 One side effect of the squash is that previous inline comments are showing as outdated on the new commit. The line context is unchanged, so the resolutions should still hold against `6f465ca`.
 
 If everything looks good, could you take another pass when you have a moment?
 
 @lzchen thanks for the deep review. Your seven inline comments are all addressed in the squashed commit, and my replies to each are still on the original threads (now marked outdated by the force push).

util/opentelemetry-util-genai/src/opentelemetry/util/genai/_inference_invocation.py
Nik-Reddy: Leaning toward keeping the Protocol for this PR, batching the decision with the next push.
 
 The trade I see:
  - Protocol: `stream.py` stays decoupled from the concrete invocation types, and `EmbeddingInvocation`, `ToolInvocation`, `AgentInvocation`, `WorkflowInvocation` do not carry `ttfc_seconds` and the chunk-gap buffer as dead state. A future streaming invocation type implements the Protocol surface (`ttfc_seconds: float | None` attr and `_record_chunk_gap` method) and gets type-checking from pyright.
 - On `GenAIInvocation`: one doc site for the streaming-timing contract, but every invocation pays for state it does not use, and there is no signal that distinguishes streaming from non-streaming invocations at the type level.
 
 Threads at `stream.py:46` and `_inference_invocation.py:185` are the same call, so I will follow your direction here for all three.
 
 Happy to flip to the base-class version if you prefer the single-site documentation.

util/opentelemetry-util-genai/src/opentelemetry/util/genai/_inference_invocation.py
Nik-Reddy: @lzchen I'm leaning toward option 4 (keep as is, document the memory shape) for this PR. The buffer lifetime is bounded by the invocation, and the per-invocation list length is the SDK's chunk count for that one call. For openai-v2 chat completion streams in scope here, that stays in the thousands of floats per invocation and is released at stop. Can add a short comment near `chunk_gap_seconds` documenting the shape.
 
Option 2 (inline record with snapshot attributes at first-chunk) is the cleaner path if we later target agent-style indefinite streams. Happy to switch to option 2 here if you would rather not defer it.

util/opentelemetry-util-genai/src/opentelemetry/util/genai/stream.py
```diff
@@ -30,6 +31,19 @@ def __init__(self, wrapped: object) -> None: ...
...
+    """Any object that can receive streaming timing measurements.
+
+    ``ttfc_seconds`` is set once by the wrapper at finalization. Each
+    inter-chunk gap is pushed to ``_record_chunk_gap`` as soon as it is
+    measured so the wrapper does not need to buffer measurements.
+    """
+
+    ttfc_seconds: float | None
+
+    def _record_chunk_gap(self, gap: float) -> None: ...
```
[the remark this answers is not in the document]
Nik-Reddy: The leading underscore on both _TimingTarget and _record_chunk_gap is deliberate: this protocol is the internal wrapper-to-invocation seam inside this package, not a public extension point that third parties should implement directly. The single implementer today lives one module over (InferenceInvocation). Tightened the docstring in dbcafb1 to call that out explicitly so the word "exposed" does not mislead readers.

util/opentelemetry-util-genai/src/opentelemetry/util/genai/stream.py
```diff
@@ -116,22 +144,53 @@ def __next__(self) -> ChunkT:
...
+            self._self_timing_target._record_chunk_gap(gap)
+        except Exception:  # pylint: disable=broad-exception-caught
+            _logger.debug(
+                "GenAI stream instrumentation error recording chunk gap",
+                exc_info=True,
+            )
+
+    def _sync_timing_to_target(self) -> None:
+        if self._self_timing_target is not None:
+            self._self_timing_target.ttfc_seconds = self._self_ttfc_seconds
```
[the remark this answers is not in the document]
Nik-Reddy: The duplication is real and the direction is correct. Holding off in this PR because there is only one real caller of the timing path today; the cleanest extraction shape (a small private mixin or a free function) is easier to pick once PR 2 wires chat_stream through and we have a second caller driving the shape of the shared helper. The sync and async paths are short and structurally identical, so drift risk is contained in the meantime. Will revisit and extract in a follow-up once the seam has more than one consumer.

util/opentelemetry-util-genai/tests/test_stream.py
```diff
@@ -540,3 +541,218 @@ async def exercise():
...
+def test_sync_stream_wrapper_no_ttfc_without_start_time():
+    """Without start_time_s, TTFC stays None."""
+    stream = _FakeSyncStream(chunks=["a", "b"])
+    wrapper = _TestSyncStreamWrapper(stream)
+
+    list(wrapper)
+
+    assert wrapper._self_ttfc_seconds is None
+    # _self_last_chunk_at is set on every chunk after the first arrival
+    assert wrapper._self_last_chunk_at is not None
```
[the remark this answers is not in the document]
Nik-Reddy: Picked up alongside the guard change in dbcafb1. With timing disabled (no start_time_s and no timing_target), _self_last_chunk_at now stays None for the same reason _self_ttfc_seconds does, so the test comment is rewritten to describe the new behavior and the assertion is flipped to assert _self_last_chunk_at is None. Both the comment and the assertion now match the wrapper.

util/opentelemetry-util-genai/src/opentelemetry/util/genai/stream.py
Nik-Reddy: Addressed by the move to O(1) state on the wrapper (_self_first_chunk_at, _self_last_chunk_at, _self_ttfc_seconds) and pushing each inter-chunk gap to the timing target as it is measured, rather than buffering on the wrapper or measuring blocking-read time. Latest commit is dbcafb1. Marking this outdated thread as resolved.

instrumentation/opentelemetry-instrumentation-openai-v2/src/opentelemetry/instrumentation/openai_v2/chat_wrappers.py
Nik-Reddy: TTFC anchor lives at the SDK call site now: chat_wrappers.py captures the timer at the call boundary and passes it as start_time_s into SyncStreamWrapper / AsyncStreamWrapper, which compute ttfc_seconds against that anchor on first-chunk arrival. Latest commit is dbcafb1. Marking this outdated thread as resolved.

util/opentelemetry-util-genai/src/opentelemetry/util/genai/stream.py
```diff
@@ -116,22 +144,53 @@ def __next__(self) -> ChunkT:
...
+            self._self_first_chunk_at = t_after_read
+            if self._self_start_time_s is not None:
+                self._self_ttfc_seconds = (
+                    t_after_read - self._self_start_time_s
+                )
+        else:
+            previous = self._self_last_chunk_at
+            if previous is not None:
+                self._safe_record_chunk_gap(t_after_read - previous)
+        self._self_last_chunk_at = t_after_read
```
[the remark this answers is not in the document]
Nik-Reddy: Latest push (https://github.com/open-telemetry/opentelemetry-python-genai/commit/dbcafb17ce5e8316ef30875c6dad0e0b843cf577) gates the timer plus the TTFC and chunk-gap bookkeeping behind a single guard so they only run when start_time_s or timing_target is set. That keeps the zero-overhead path for users who only use the wrapper for finalization hooks without TTFC or gap collection. Mirrored the same guard in AsyncStreamWrapper.anext so the two paths stay in sync.

Nik-Reddy:
@MikeGoldsmith @lzchen  friendly ping for re-review when you have a cycle.


SyncStreamWrapper.next and AsyncStreamWrapper.anext now gate the per-chunk timer and the TTFC and chunk-gap bookkeeping behind a single guard, so the wrapper is zero-overhead when neither start_time_s nor timing_target is set.
_TimingTarget docstring tightened to call out that the protocol is the internal wrapper-to-invocation seam, not a public extension point. test_stream.py updated to assert _self_last_chunk_at stays None when timing is disabled (matches the new guard). Sync/Async helper duplication called out by Copilot deferred to a follow-up alongside PR 2 (chat_stream wiring), with rationale on the inline thread.

For your 5/14 review specifically:
instruments.py / time_per_output_chunk semconv wording: addressed; the description now matches the semconv link verbatim. metrics.py wrapper-side list cleanup: addressed per your 5/19 follow-up; wrapper holds O(1) state, the invocation buffers via _record_chunk_gap, and metrics.py drains via _consume_streaming_timing. chat_wrappers.py private monotonic_start attribute: addressed; the attribute is public (monotonic_start_s) at both write and read sites.

Nik-Reddy:
@MikeGoldsmith @lzchen gentle ping when you have a chance.
 
 I rebased this on current main, resolved the `test_handler_metrics.py` conflict, and fixed the Google GenAI cassette issue that showed up after CI refreshed. The branch is still a single commit.
 
 Could you please take another look when you have time?

instrumentation/opentelemetry-instrumentation-google-genai/tests/generate_content/cassettes/test_upload_hook_non_streaming[SPAN_AND_EVENT-vertexaiapi-sync-gemini-2.5-flash-enable_completion_hook].yaml
Comment (lmolkova) [inline instrumentation/opentelemetry-instrumentation-google-genai/tests/generate_content/cassettes/test_upload_hook_non_streaming[SPAN_AND_EVENT-vertexaiapi-sync-gemini-2.5-flash-enable_completion_hook].yaml:1]: are these files related to the PR?

lmolkova (changes requested):
I have a few suggestions demonstrated in https://github.com/Nik-Reddy/opentelemetry-python-genai/commit/b3c198b9277c9f7316b73fc57542ae3fff6fad72:

1. Report chunks as they arrive, don't accumulate timings
2. Pass InferenceInvocation instead of inventing _TimingTarget - we can support addition invocations for Agents / workflows if ever they would need to support streaming
3. Add conformance test validating streaming

PR #16, opened 2026-05-14
ci: introduce towncrier for fragment-based changelog management

MikeGoldsmith:
## Summary

Adopt [towncrier](https://towncrier.readthedocs.io/) for fragment-based changelog management. PRs with user-visible impact add a file under `<package>/.changelog/<PR>.<type>` rather than editing `CHANGELOG.md`, which removes the merge-conflict footgun on backports and concurrent PRs.

Every publishable package here releases independently, so each gets its own `[tool.towncrier]` config and `.changelog/` directory — mirroring the independent-package pattern in `opentelemetry-python-contrib`.

What's in this PR:
- `[tool.towncrier]` block in each package's `pyproject.toml` (7 instrumentations + `util-genai`)
- `<package>/.changelog/.gitignore` per package
- Do-not-edit comment and `<!-- changelog start -->` marker above each existing `## Unreleased` heading
- Shared template at `scripts/changelog_template.j2`
- `.github/workflows/changelog.yml`: blocks direct `CHANGELOG.md` edits, requires a fragment, previews rendered changelogs per package; skipped via the `Skip Changelog` label
- `tox -e changelog-preview` for local previewing
- `towncrier` in root dev dependencies
- `CONTRIBUTING.md` updated with fragment authoring instructions, and a condensed version added to `AGENTS.md`

Part of #15.

## Notes

- This PR carries the `Skip Changelog` label since the change is pure tooling.
- The `Skip Changelog` and `Stale` labels were created on this repo as a prerequisite. `prepare-release` is intentionally not created yet — to be added when release tooling lands (planned name: `Prepare Release`).
- Verified locally that the per-package draft renders correctly with a throwaway fragment and that `tox -e changelog-preview` runs cleanly across all 8 packages.

PR #17, opened 2026-05-14
 Refactor `GenAIInvocation` to be a `ContextManager` to simplify code. Minor change to `ToolInvocation`.

DylanRussell:
A couple minor changes needed to migrate Google GenAi over to the utils library..

Make `GenAiInvocation` class (and parent classes) a context manager by defining enter/exit methods.. This way it can be optionally used as a ContextManager like python's `file` object, and we just need 1 method (`invocation`) instead of 2 (`start_invocaton`/`invocation`)

Open questions:

What's the best way to deprecate these methods ? We could do a release of GenAi utils, then I could update all instrumentations to stop using these methods, then release those instrumentations, and then finally remove these methods. Or we could have a long deprecation period ? But that feels unnecessary..

Should we stop using `typing.Any` types on `ToolInvocation` and all these other classes ? IMO we should force instrumentation to serialize these to things that can go into the span/metric/log (like the OTEL  Attributes type), `Any` is too generic to serialize.. Also IMO the sem convs themselves should be more opinionated about how to serialize things..


util/opentelemetry-util-genai/src/opentelemetry/util/genai/_invocation.py
```diff
@@ -165,13 +165,17 @@ def fail(self, error: Error | BaseException) -> None:
...
+
+    def __exit__(
+        self,
+        type_: type[BaseException] | None,
+        value: BaseException | None,
+        traceback: TracebackType | None,
+    ) -> None:
+        if value:
+            self.fail(value)
             raise
```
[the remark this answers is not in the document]
DylanRussell: Done

util/opentelemetry-util-genai/src/opentelemetry/util/genai/_agent_invocation.py
```diff
@@ -59,7 +58,7 @@ def __init__(
         server_port: int | None = None,
         agent_name: str | None = None,
     ) -> None:
-        """Use handler.start_invoke_local_agent() or handler.start_invoke_remote_agent() instead of calling this directly."""
+        """Use handler.invoke_local_agent() or handler.invoke_remote_agent() instead of calling this directly."""
```
eternalcuriouslearner, line 61: question(non-blocking): Should we update the agents.md with this new invocation style?

util/opentelemetry-util-genai/src/opentelemetry/util/genai/_invocation.py
```diff
@@ -165,15 +165,19 @@ def fail(self, error: Error | BaseException) -> None:
...
-    @contextmanager
-    def _managed(self) -> Generator[Self, None, None]:
-        """Context manager that calls stop() on success or fail() on exception."""
-        try:
-            yield self
-        except Exception as exc:
-            self.fail(exc)
-            raise
-        self.stop()
+    def __enter__(self):
```
eternalcuriouslearner: suggestion(non-blocking): Can we use `contextlib.AbstractContextManager` here and modify this as:

```
 from contextlib import AbstractContextManager                                                                                                                                                           
                                                                                                                                                                                                          
  class GenAIInvocation(ABC, AbstractContextManager["GenAIInvocation"]):                                                                                                                                  
      ...                                                                                                                                                                                                 
      def __exit__(                                                                                                                                                                                       
          self,                                                                                                                                                                                         
          exc_type: type[BaseException] | None,
          exc_value: BaseException | None,
          traceback: TracebackType | None,                                                                                                                                                                
      ) -> bool | None:
          if exc_value is not None and isinstance(exc_value, Exception):                                                                                                                                  
              self.fail(exc_value)                                                                                                                                                                      
          else:                                                                                                                                                                                           
              self.stop()
          return None  # don't suppress     
```

Let me know If you feel this as overkill. I am fine with your current approach btw.

eternalcuriouslearner (approved):
Looks pretty neat!! Left few non-blocking suggestions. For the two questions in the pr descriptions:

1. I am leaning towards making incremental change with three libraries at a time. Lmk if you want to divide and conquer this. We can plan and accomplish this.
2. I am strongly leaning towards enforcing types. Types provide better contract guarantees imo.

aabmass:

util/opentelemetry-util-genai/src/opentelemetry/util/genai/_agent_invocation.py
DylanRussell: Ack, yes, done

util/opentelemetry-util-genai/src/opentelemetry/util/genai/_invocation.py
DylanRussell: SG, changed it to be like that..

DylanRussell:
Ack updated to town crier format..

@eternalcuriouslearner 

Sounds good, lets get this PR pushed through, do a release, then we can work together to update all the instrumentations..

Regarding types that also sounds good.. I've updated tool arguments / tool result to be an Attribute value in this PR, but there are lots of other instances of this issue, opened up https://github.com/open-telemetry/opentelemetry-python-genai/issues/41 to track this work...

lzchen:
This change makes a lot of sense to me but there are promises in this repo to not making breaking changes (at least carried over from the contrib repo). Are we releasing utils as a new package as well or continuing the current package name? Sorry if I've missed previous discussions already.

DylanRussell:
Yes I'm pretty sure the plan is to continue using the same package name..

Regarding breaking changes: do we think it's likely that anyone other than ourselves is actually using this ? Was this library actually meant for external use ? I thought it was just a helper library for our own instrumentations.. It would be good to clarify that.. It is still pretty new, and it is a pain (and error prone) to have to carry forward 2 duplicate methods for all these types..

lzchen:
@DylanRussell 

Definitely agree with the sentiment. I think now is a better time than any to get rid of any functionality/classes we deem as old or deprecated. The main purpose imo for keeping them before was if anyone was relying on the old implementations in instrumentations or there were instrumentation prs in progress, but now since we are releasing new instrumentation packages under new names, I think we can get rid of them.

One nit I would say is the pr title "minor changes" is a bit misleading as this type of api change should involve more of a discussion and preferably a maintainer as well.

util/opentelemetry-util-genai/src/opentelemetry/util/genai/_invocation.py
```diff
@@ -165,15 +167,19 @@ def fail(self, error: Error | BaseException) -> None:
...
```
lzchen: Nit: typing `GenAiInvocation` or `Self`?

lzchen (approved):

util/opentelemetry-util-genai/src/opentelemetry/util/genai/_invocation.py
DylanRussell: Yeah good idea. done

DylanRussell:
Yeah the PR name is pretty bad fair enough.. updated it to be more clear..

We will be sure to remove all usages of these methods before getting rid of them.

lmolkova (approved):

PR #23, opened 2026-05-15
Add small helper script/pyproject.toml file helpfully for manually inspect changes to Google GenAi instrumentation library

DylanRussell:
Add a small python script and example `pyproject.toml` file that enables you to manually compare logs / spans generated by the Google GenAi instrumentation with another version of the instrumentation (one from your local branch for example).

I found this very helpful when refactoring the code.. I was able to diff the logs / spans before / after a change I made to the instrumentation..

I'm not sure where the best place to put this is ? Open to suggestions on that..


aabmass:

instrumentation/opentelemetry-instrumentation-google-genai/tests/diff_instrumentation_changes.py
```diff
@@ -0,0 +1,108 @@
...
+from opentelemetry.sdk.trace.export.in_memory_span_exporter import (
+    InMemorySpanExporter,
+)
+
+os.environ["OTEL_SEMCONV_STABILITY_OPT_IN"] = "gen_ai_latest_experimental"
+os.environ["OTEL_INSTRUMENTATION_GENAI_CAPTURE_MESSAGE_CONTENT"] = (
+    "SPAN_AND_EVENT"
+)
+
+PROJECT_ID = "my-project-id"
```
MikeGoldsmith: Should we make these use env vars or input params instead of hardcoding?

instrumentation/opentelemetry-instrumentation-google-genai/dev/call_instrumentation_lib_and_save_telemetry.py
```diff
@@ -0,0 +1,108 @@
+# Copyright The OpenTelemetry Authors
```
MikeGoldsmith, line 1: We should move this helper script out of `tests/` so it doesn't get bundled when published to PyPI. `tests/` is in the package's sdist include list.

I suggest a top-level `dev/` or `scripts/` directory at the repo root (or a `dev/` sibling to `src/` within the package).

MikeGoldsmith (dismissed):
Looks good, thanks @DylanRussell. Just a little clean-up for location and a suggestion for passing in values.

instrumentation/opentelemetry-instrumentation-google-genai/dev/call_instrumentation_lib_and_save_telemetry.py
DylanRussell: Ah good idea.. created a `dev/` next to `src/`

instrumentation/opentelemetry-instrumentation-google-genai/tests/diff_instrumentation_changes.py
DylanRussell: yeah good idea.. adding a main.env and readme files

instrumentation/opentelemetry-instrumentation-google-genai/.changelog/23.added
aabmass, line 1: Can probably just skip the changelog?

instrumentation/opentelemetry-instrumentation-google-genai/dev/README.md
```diff
@@ -0,0 +1,10 @@
+# Copyright The OpenTelemetry Authors
```
aabmass: Was this supposed to be a comment? Should be html style

aabmass (approved):
I think it would be useful to update the tests (or write new ones) to use the inline-snapshot library so you could also diff the test outputs instead of just a single canned prompt, and make it easier to review and write tests.

But we can do this script for now

DylanRussell:
> I think it would be useful to update the tests (or write new ones) to use the inline-snapshot library so you could also diff the test outputs instead of just a single canned prompt, and make it easier to review and write tests.

This is tricky because it requires 2 different virtual environments to be set up with the 2 different dependency sets.. I'm not sure how to do that..

instrumentation/opentelemetry-instrumentation-google-genai/.changelog/23.added
DylanRussell: you asked me to add this but maybe that was before you look at the PR.. anyway i've removed it

instrumentation/opentelemetry-instrumentation-google-genai/dev/README.md
DylanRussell: I see other readme's don't have the copy right.. so i've just removed this

lmolkova (approved):
LGTM, it seems it could be generalized to any other library, and then it would end up in `util/opentelemetry-test-util-genai` folder. We can always generalize later

DylanRussell:
Can this be merged now ?

PR #25, opened 2026-05-17
Add LangChain invoke workflow and invoke agent span support

wrisa:
Brought from old repo: https://github.com/open-telemetry/opentelemetry-python-contrib/pull/4449/

Note: Current PR generates `invoke_agent` spans when genuine react agent eg: create_react_agent. Opened [issue](https://github.com/open-telemetry/semantic-conventions-genai/issues/264) to discuss which span we want when there are plain langgraph nodes(without any actual agent)


aabmass:

instrumentation/opentelemetry-instrumentation-langchain/src/opentelemetry/instrumentation/langchain/operation_mapping.py
```diff
@@ -0,0 +1,217 @@
...
+    if serialized:
+        name = serialized.get("name", "")
+        graph_id = (
+            serialized.get("graph", {}).get("id", "")
+            if isinstance(serialized.get("graph"), dict)
+            else ""
+        )
+        return LANGGRAPH_IDENTIFIER in name or LANGGRAPH_IDENTIFIER in graph_id
+
+    return True
```
lzchen, line 139: Continuing from the [previous discussion](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/4449#discussion_r3221579464). This makes sense to me. I was just wondering the nuance of "if something is root-level and we have zero information about it, assume it's a workflow rather than silently dropping it." I think we are taking an approach and assuming that we want more data than assuming to ignore it (which I think is correct). Is my understanding correct? If so, please add similar comments here like what you proposed :)

instrumentation/opentelemetry-instrumentation-langchain/src/opentelemetry/instrumentation/langchain/operation_mapping.py
```diff
@@ -0,0 +1,217 @@
...
+# ---------------------------------------------------------------------------
+
+
+class OperationName:
+    """Canonical GenAI semantic convention operation names."""
+
+    INVOKE_AGENT: str = GenAI.GenAiOperationNameValues.INVOKE_AGENT.value
+    # invoke_workflow is not yet in the semconv enum; use the expected
+    # string value so the mapping is forward-compatible.
+    INVOKE_WORKFLOW: str = "invoke_workflow"
```
lzchen: Nit: Could you create a tracking issue after this is merged?

instrumentation/opentelemetry-instrumentation-langchain/pyproject.toml
```diff
@@ -37,8 +38,8 @@ instruments = [
 langchain = "opentelemetry.instrumentation.langchain:LangChainInstrumentor"
 
 [project.urls]
-Homepage = "https://github.com/open-telemetry/opentelemetry-python-genai/tree/main/instrumentation/opentelemetry-instrumentation-langchain"
-Repository = "https://github.com/open-telemetry/opentelemetry-python-genai"
+Homepage = "https://github.com/open-telemetry/opentelemetry-python-contrib/tree/main/instrumentation-genai/opentelemetry-instrumentation-langchain"
```
lzchen: I think this would need to point to the new genai repo?

instrumentation/opentelemetry-instrumentation-langchain/tests/test_invocation_manager.py
```diff
@@ -104,7 +104,73 @@ def test_delete_invocation_state(invocation_manager, mock_invocation):
     assert run_id not in invocation_manager._invocations
 
 
-def test_delete_invocation_state_with_children(invocation_manager):
+def test_delete_invocation_state_deferred_while_children_live(
```
lzchen, line 107: Should we have vcr cassettes to test out real langchain interactions similar to the chat model calls?

instrumentation/opentelemetry-instrumentation-langchain/src/opentelemetry/instrumentation/langchain/utils.py
```diff
@@ -0,0 +1,94 @@
...
+    exclude_keys = {"messages", "intermediate_steps"}
+    input_data: dict[str, Any] = {
+        k: v
+        for k, v in data_dict.items()
+        if k not in exclude_keys and v is not None
+    }
+    if input_data:
+        serialized = serialize(input_data)
+        if serialized:
+            return [InputMessage(role="user", parts=[Text(serialized)])]
```
lmolkova, line 44: could you please create an issue in semantic-conventions-genai repo to support it properly? we should consider adding structured input part

instrumentation/opentelemetry-instrumentation-langchain/src/opentelemetry/instrumentation/langchain/operation_mapping.py
```diff
@@ -0,0 +1,217 @@
...
+
+    return True
+
+
+# ---------------------------------------------------------------------------
+# Public API
+# ---------------------------------------------------------------------------
+
+
+def should_ignore_chain(
```
lmolkova:
```suggestion
def _should_ignore_chain(
```

other helpers here are not public

instrumentation/opentelemetry-instrumentation-langchain/examples/agent/single_node_agent.py
lmolkova, line 1: Please add conformance test scenarios for agent and workflow similar to https://github.com/open-telemetry/opentelemetry-python-genai/blob/7cafd934e0cb23f26dd58a41460011d0f84ccbc2/instrumentation/opentelemetry-instrumentation-openai-v2/tests/conformance/tool_calling.py - these tests can actually replaces examples reasonably well, so feel free to remove examples.

instrumentation/opentelemetry-instrumentation-langchain/src/opentelemetry/instrumentation/langchain/callback_handler.py
```diff
@@ -34,6 +46,139 @@ def __init__(self, telemetry_handler: TelemetryHandler) -> None:
...
+        operation = classify_chain_run(
+            serialized, metadata, kwargs, parent_run_id
+        )
+
+        if operation == OperationName.INVOKE_WORKFLOW:
+            workflow_name = kwargs.get("name") or serialized.get("name")
+            workflow_name_override = (
+                metadata.get("workflow_name") if metadata else None
+            )
+            workflow = self._telemetry_handler.start_workflow(
```
lmolkova: check out #17 - this and other start_* are deprecated now and you can use `workflow` directly

instrumentation/opentelemetry-instrumentation-langchain/src/opentelemetry/instrumentation/langchain/callback_handler.py
```diff
@@ -34,6 +46,139 @@ def __init__(self, telemetry_handler: TelemetryHandler) -> None:
...
+                agent_invocation_name_lower = (
+                    agent_invocation_name.lower()
+                    if agent_invocation_name
+                    else None
+                )
+                if suggested_agent_name_lower != agent_invocation_name_lower:
+                    agent = self._telemetry_handler.start_invoke_local_agent(
+                        provider=metadata.get("ls_provider", "unknown")
+                        if metadata
+                        else "unknown",
```
lmolkova, line 97: please create an issue in semconv-genai repo to decide what to do with gen_ai.provider.name for agentic frameworks

instrumentation/opentelemetry-instrumentation-langchain/src/opentelemetry/instrumentation/langchain/callback_handler.py
```diff
@@ -34,6 +46,139 @@ def __init__(self, telemetry_handler: TelemetryHandler) -> None:
...
+            )
+            if suggested_agent_name:
+                suggested_agent_name_lower = suggested_agent_name.lower()
+                agent_invocation_name_lower = (
+                    agent_invocation_name.lower()
+                    if agent_invocation_name
+                    else None
+                )
+                if suggested_agent_name_lower != agent_invocation_name_lower:
+                    agent = self._telemetry_handler.start_invoke_local_agent(
```
lmolkova: we should pass agent name to `invoke_local_agent`, without it span name does not include it 

Image

lmolkova (commented):
Thanks, it looks good!  I've read #28 and agree with this modeling. Just a few minor concerns.

instrumentation/opentelemetry-instrumentation-langchain/src/opentelemetry/instrumentation/langchain/operation_mapping.py
wrisa: opentelemetry-semantic-conventions==0.63b0 has it, updated.

instrumentation/opentelemetry-instrumentation-langchain/tests/test_invocation_manager.py
wrisa: chat model vcr tests - integration tests that call real LLM APIs → cassettes replay the HTTP traffic 
invocation manager - pure unit tests of internal lifecycle logic; no I/O → no cassettes needed

instrumentation/opentelemetry-instrumentation-langchain/src/opentelemetry/instrumentation/langchain/utils.py
wrisa: [issue](https://github.com/open-telemetry/semantic-conventions-genai/issues/175) added

instrumentation/opentelemetry-instrumentation-langchain/src/opentelemetry/instrumentation/langchain/callback_handler.py
wrisa: created [issue](https://github.com/open-telemetry/semantic-conventions-genai/issues/178)

policies/genai_span_validation.rego
```diff
@@ -158,11 +158,9 @@ _execute_tool_expected := {
 
 # Invoke agent.
 # Required: gen_ai.operation.name, gen_ai.provider.name.
-# Most instrumentations should have agent.id; flag it as always-emit.
 _invoke_agent_expected := {
 	"gen_ai.operation.name",
 	"gen_ai.provider.name",
```
lmolkova, line 163: should we have gen_ai.t.name here instead?

lmolkova (approved):

tox.ini
```diff
@@ -122,6 +122,7 @@ deps =
   ; Langchain has no oldest/latest matrix yet (TODO above); pin the client
```
DylanRussell: Even if it doesn't have oldest/latest (which IMO is more of a burden than adding any value), shouldn't langchain's unit tests be added here ?

lzchen (approved):

policies/genai_span_validation.rego
wrisa: @lmolkova it is optional in [semcov](https://github.com/open-telemetry/semantic-conventions-genai/blob/main/docs/gen-ai/gen-ai-agent-spans.md#invoke-agent-client-span)

tox.ini
DylanRussell: Thanks for adding

DylanRussell (approved):

PR #37, opened 2026-05-18
Add tools support in langchain

wrisa:
[examples/tools/main.py](https://github.com/open-telemetry/opentelemetry-python-genai/pull/37/changes#diff-980e9416c3bf6757b78724418599181af95a7586e17ce8c788c3db7182fcfc81)
Tools span:
Screenshot 2026-05-18 at 9 52 10 AM

Inference spans with tool details:
Screenshot 2026-05-18 at 9 52 23 AM
Screenshot 2026-05-18 at 9 52 35 AM


util/opentelemetry-util-genai/src/opentelemetry/util/genai/_inference_invocation.py
```diff
@@ -252,6 +252,7 @@ class LLMInvocation:
     seed: int | None = None
     server_address: str | None = None
     server_port: int | None = None
+    tool_definitions: list[ToolDefinition] | None = None
```
lmolkova: is this necessary? let's update langchain to use new API instead of legacy LLMInvocation

instrumentation/opentelemetry-instrumentation-genai-langchain/tests/test_tools.py
lmolkova, line 1: could you please add a conformance test for tool calling in https://github.com/open-telemetry/opentelemetry-python-genai/tree/main/instrumentation/opentelemetry-instrumentation-langchain/tests/conformance ? 

It's testing tool calling along with multi-turn conversations and can be almost copied over from https://github.com/open-telemetry/opentelemetry-python-genai/blob/main/instrumentation/opentelemetry-instrumentation-openai-v2/tests/conformance/tool_calling.py

lmolkova (commented):
Looks good, let's just update langchain to use new `InferenceInvocation` instead of deprecated `LLMInvocation`

instrumentation/opentelemetry-instrumentation-genai-langchain/tests/conformance/tool_calling.py
```diff
@@ -0,0 +1,136 @@
...
+    }
+
+
+def _execute_weather_tool(arguments: str) -> str:
+    location = json.loads(arguments)["location"]
+    return WEATHER_BY_LOCATION[location]
+
+
+class ToolCallingScenario(Scenario):
+    expected_spans = ("chat",)
```
lmolkova: where's the tool call in expected spans?

instrumentation/opentelemetry-instrumentation-genai-langchain/src/opentelemetry/instrumentation/genai/langchain/callback_handler.py
```diff
@@ -402,6 +427,72 @@ def on_llm_error(
...
+
+        raw_arguments: Any = inputs if inputs is not None else input_str
+        arguments: str | None
+        if isinstance(raw_arguments, dict):
+            arguments = json.dumps(raw_arguments)
+        elif isinstance(raw_arguments, str):
+            arguments = raw_arguments
+        else:
+            arguments = None
+        tool_invocation = self._telemetry_handler.start_tool(
```
lmolkova: there is a better helper now, we should stop using start_ helpers - they will be removed at some point soon
```suggestion
        tool_invocation = self._telemetry_handler.tool(
```

instrumentation/opentelemetry-instrumentation-genai-langchain/src/opentelemetry/instrumentation/genai/langchain/callback_handler.py
```diff
@@ -402,6 +427,72 @@ def on_llm_error(
...
+        name = "unknown"
+        description = None
+        if serialized is not None:
+            name = serialized.get("name") or "unknown"
+            description = serialized.get("description")
+
+        raw_arguments: Any = inputs if inputs is not None else input_str
+        arguments: str | None
+        if isinstance(raw_arguments, dict):
+            arguments = json.dumps(raw_arguments)
```
lmolkova: is this necessary? can we pass Any ?

instrumentation/opentelemetry-instrumentation-genai-langchain/src/opentelemetry/instrumentation/genai/langchain/callback_handler.py
```diff
@@ -402,6 +427,72 @@ def on_llm_error(
...
+        output: Any,
+        *,
+        run_id: UUID,
+        parent_run_id: Optional[UUID] = None,
+        **_kwargs: Any,
+    ) -> None:
+        tool_invocation = self._invocation_manager.get_invocation(run_id)
+        if not isinstance(tool_invocation, ToolInvocation):
+            return
+        tool_invocation.tool_call_id = getattr(output, "tool_call_id", None)
```
lmolkova, line 470: is getattr necessary ?

instrumentation/opentelemetry-instrumentation-genai-langchain/src/opentelemetry/instrumentation/genai/langchain/callback_handler.py
```diff
@@ -333,19 +344,33 @@ def on_llm_end(
...
+                        tool_calls: list[ToolCallRequest] = []
+                        for tool_call in chat_generation.message.tool_calls:
+                            tool_call_request = ToolCallRequest(
+                                name=tool_call["name"],
+                                id=tool_call["id"],
+                                arguments=tool_call["args"],
+                            )
+                            tool_calls.append(tool_call_request)
+                        output_message = OutputMessage(
+                            role="assistant",
```
lmolkova: does it have to be hardcoded ?

instrumentation/opentelemetry-instrumentation-genai-langchain/src/opentelemetry/instrumentation/genai/langchain/callback_handler.py
wrisa: Yes, it's necessary. `invocation.arguments` is typed `AttributeValue | None`, which doesn't accept dict. The block converts the structured inputs: dict to a JSON string, falling back to the
  plain input_str: str. Passing Any directly would be a type violation and would break span attribute setting at runtime when inputs is a dict.

instrumentation/opentelemetry-instrumentation-genai-langchain/src/opentelemetry/instrumentation/genai/langchain/callback_handler.py
wrisa: It is defensive pattern here given the output:Any - typed parameter

instrumentation/opentelemetry-instrumentation-genai-langchain/src/opentelemetry/instrumentation/genai/langchain/callback_handler.py
```diff
@@ -333,19 +342,33 @@ def on_llm_end(
                             )
                         )
 
-                    # Get message content
-                    parts = [
-                        Text(
-                            content=chat_generation.message.content,
-                            type="text",
+                    if finish_reason == "tool_calls":
```
eternalcuriouslearner: question(non-blocking): Are we covering only OpenAI in this pr? Should we add tool_use also here?

instrumentation/opentelemetry-instrumentation-genai-langchain/src/opentelemetry/instrumentation/genai/langchain/callback_handler.py
lmolkova: serializing to string is not right either.`AttributeValue` is not the right type for tool.arguments, it should be `opentelemetry.util.types.AnyValue` that accepts `Mapping[str, "AnyValue"]` and could be either cast or transformed from`inputs`. Note - input_str should be deserialized if `inputs` are not provided 

https://github.com/open-telemetry/semantic-conventions-genai/blob/6967795c5ceeb77357ca4f3dfc9a512496fb9a11/docs/gen-ai/gen-ai-spans.md?plain=1#L858-L860

> It's expected to be an object - in case a serialized string is available
to the instrumentation, the instrumentation SHOULD do the best effort to
deserialize it to an object

instrumentation/opentelemetry-instrumentation-genai-langchain/src/opentelemetry/instrumentation/genai/langchain/utils.py
```diff
@@ -7,12 +7,45 @@
...
 
 
+def _get_property_value(obj: Any, property_name: str) -> Any:
+    if isinstance(obj, dict):
+        return cast(dict[str, Any], obj).get(property_name)
+
+    return getattr(obj, property_name, None)
+
+
+def prepare_tool_definitions(tools: list[Any]) -> list[ToolDefinition] | None:
```
eternalcuriouslearner, line 25: nit(non-blocker): can we use type here for the method parameter?

util/opentelemetry-util-genai/src/opentelemetry/util/genai/_tool_invocation.py
```diff
@@ -14,7 +16,18 @@
 from opentelemetry.util.genai.utils import (
     should_capture_content_on_spans,
 )
-from opentelemetry.util.types import AttributeValue
+from opentelemetry.util.types import AnyValue, AttributeValue
+
+
+def _any_value_to_attribute_value(value: AnyValue) -> AttributeValue:
```
eternalcuriouslearner: question(non-blocking): can we refine this piece of code as something like 

```suggestion

def _any_value_to_attribute_value(value: Any) -> AttributeValue | None:
    """Serialize an AnyValue to an AttributeValue for OTel span attributes."""
    if value is None:
        return None
    if isinstance(value, (bool, str, bytes, int, float)):
        return value  # type: ignore[return-value]
    try:
        return json.dumps(value, default=str)
    except (TypeError, ValueError):
        return str(value)

```

eternalcuriouslearner: Additionally do we need some tests around this? Little paranoid as I am unsure if this is going to raise or not.

eternalcuriouslearner (approved):

instrumentation/opentelemetry-instrumentation-genai-langchain/src/opentelemetry/instrumentation/genai/langchain/utils.py
wrisa: The body already handles correctly via `_get_property_value`

lmolkova (approved):

util/opentelemetry-util-genai/src/opentelemetry/util/genai/_tool_invocation.py
```diff
@@ -14,7 +16,19 @@
...
+
+
+def _any_value_to_attribute_value(value: AnyValue) -> AttributeValue | None:
+    """Serialize an AnyValue to an AttributeValue for OTel span attributes."""
+    if value is None:
+        return None
+    if isinstance(value, (bool, str, bytes, int, float)):
+        return value
+    try:
+        return json.dumps(value, default=str)
```
lmolkova:
```suggestion
        return gen_ai_json_dumps(value)
```

util/opentelemetry-util-genai/src/opentelemetry/util/genai/_tool_invocation.py
```diff
@@ -3,6 +3,8 @@
 
 from __future__ import annotations
 
+import json
+
```
lmolkova:
```suggestion
```

util/opentelemetry-util-genai/src/opentelemetry/util/genai/_tool_invocation.py
```diff
@@ -14,7 +16,19 @@
 from opentelemetry.util.genai.utils import (
     should_capture_content_on_spans,
```
lmolkova, line 16:
```suggestion
    gen_ai_json_dumps,
    should_capture_content_on_spans,
```

instrumentation/opentelemetry-instrumentation-genai-langchain/src/opentelemetry/instrumentation/genai/langchain/callback_handler.py
```diff
@@ -399,6 +422,72 @@ def on_llm_error(
...
+        *,
+        run_id: UUID,
+        parent_run_id: Optional[UUID] = None,
+        **_kwargs: Any,
+    ) -> None:
+        tool_invocation = self._invocation_manager.get_invocation(run_id)
+        if not isinstance(tool_invocation, ToolInvocation):
+            return
+        tool_invocation.tool_call_id = getattr(output, "tool_call_id", None)
+        tool_invocation.tool_result = getattr(output, "content", None)
```
lmolkova, line 471: it should also be AnyValue and needs to be serialized to json (inside tool invocation code).
Can you please add/update tests to cover complex value in tools arguments and result? Thanks!

PR #47, opened 2026-05-19
Remove remaining usage/references to deprecated functions in `opentelemetry-util-genai`.

DylanRussell:
# Description

Remove from instrumentation libraries the last usages of and references to deprecated functions in `opentelemetry-util-genai`.  Also do some minor code cleanup in `opentelemetry-instrumentation-google-genai`.

It's critical that (after merging this) we first do a release of `opentelemetry-util-genai`, then update the `pyproject.toml` of the following instrumentations to require (`>=`) the new version of `opentelemetry-util-genai`: `opentelemetry-instrumentation-langchain`, `opentelemetry-instrumentation-google-genai`, `opentelemetry-instrumentation-openai-v2`, `opentelemetry-instrumentation-anthropic`

Question for  the release pipeline: is there a nice way to handle this intra-repo package dependency ? The process right now feels kind of clunky: we have to do a utils release, then remember to bump all the instrumentation `.toml` files to use it and then do a release of the instrumentation packages...

## Type of change

- [x ] Bug fix (non-breaking change which fixes an issue)

# How has this been tested?

Unit tests

# Checklist

- [ x] Followed the style guidelines of this project
- [ x] Changelog updated if the change requires an entry
- [ x] Unit tests added
- [ x] Documentation updated


instrumentation/opentelemetry-instrumentation-google-genai/src/opentelemetry/instrumentation/google_genai/tool_call_wrapper.py
```diff
@@ -75,16 +75,17 @@ def _wrap_tool_function(
...
         @functools.wraps(tool_function)
         async def wrapped_function(*args, **kwargs):
+            # Always json.dumps. First we convert args / result to something that we can serialize, then we serialize.
+            # The return value of _to_otel_value could be a dict, which currently cannot be a span attribute..
+            # In the future that could change (see https://github.com/open-telemetry/opentelemetry-specification/pull/4485), and we could possibly stop using json.dumps here.
             with telemetry_handler.tool(
-                tool_function.__name__, tool_description=tool_function.__doc__
+                tool_function.__name__,
+                tool_description=tool_function.__doc__,
+                arguments=json.dumps(
```
lmolkova: this should not be a start argument - it's expensive to obtain and should not be captured if content is disabled

lmolkova (changes requested):
I'd prefer us not to remove public  things from genai-utils until we deprecate old versions of libs in python-contrib. We should give people some grace period to update to new versions of everything. 

Other than removals of public methods from genai utils this looks great!

instrumentation/opentelemetry-instrumentation-google-genai/src/opentelemetry/instrumentation/google_genai/tool_call_wrapper.py
DylanRussell: The utils library will not add them to the span unless content captured is enabled, so we are covered there.

The problem with adding arguments after the function has been executed is that is if the execute tool call fails, arguments would then be missing

DylanRussell:
Ok i've reverted  the function deletions

instrumentation/opentelemetry-instrumentation-google-genai/src/opentelemetry/instrumentation/google_genai/tool_call_wrapper.py
lmolkova: I mean from perf-standpoint - they should not be provided at start time, we should not serialize them just to drop them later if content is disabled. 

we can do 
```
            with telemetry_handler.tool(
                tool_function.__name__,
                tool_description=tool_function.__doc__
            ) as tool_invocation:
              if handler.content_enabled:
                tool_invocation.arguments=json.dumps( 
                    _get_function_args(tool_function, args, kwargs)
                ),
```

arguments should not even be settable in the `tool` factory (it's not sampling-relevant attribute according to conventions), but it's not something introduced in your PR

DylanRussell: Ah ok didn't realize that.. That was something I added myself just the other day so I could do this.. I've now reverted that change in this PR..

lmolkova (approved):

util/opentelemetry-util-genai/.changelog/47.removed
```diff
@@ -0,0 +1 @@
+Remove remaining usages of and references to deprecated functions
```
lmolkova, line 1: nit, it seems it's not applicable to util package - the change is on the instrumentation packages

DylanRussell: I just updated a changelog in 1 of the packages I made changes in to save time, if you think it's important to update all changelogs please unresolved this and I'll do it..

util/opentelemetry-util-genai/src/opentelemetry/util/genai/_tool_invocation.py
lzchen: Update the docstring?

util/opentelemetry-util-genai/src/opentelemetry/util/genai/handler.py
```diff
@@ -208,7 +206,6 @@ def start_tool(
         self,
         name: str,
         *,
-        arguments: AttributeValue | None = None,
```
lzchen, line 211: What's this change for?

lzchen:
Was the plan to remove the deprecated functions from all instrumentations in this pr or just google genai? Like [here](https://github.com/open-telemetry/opentelemetry-python-genai/blob/main/instrumentation/opentelemetry-instrumentation-genai-openai/src/opentelemetry/instrumentation/genai/openai/utils.py#L355)

util/opentelemetry-util-genai/src/opentelemetry/util/genai/_tool_invocation.py
DylanRussell: Ah missed that one.. fixed thanks

util/opentelemetry-util-genai/src/opentelemetry/util/genai/handler.py
DylanRussell: Arguments can be expensive to serialize and aren't supposed to be used as part of the span sampling decision according to Ludmilla, so we don't want them passed at the start.. I had added arguments as a parameter to start_tool in a previous PR and that hasn't been released yet.. So this change shouldn't break anyone

DylanRussell:
All instrumentations !

lzchen:
> All instrumentations !

Was [this](https://github.com/open-telemetry/opentelemetry-python-genai/blob/main/instrumentation/opentelemetry-instrumentation-genai-openai/src/opentelemetry/instrumentation/genai/openai/utils.py#L355) missed?

DylanRussell:
I don't think so ?? I removed that reference in this PR

lzchen (approved):

PR #49, opened 2026-05-20
Improve OpenAI Agents conformance and metrics

alfozan:
# Description

Updates the OpenAI Agents instrumentation to better handle current Agents SDK tracing data and the GenAI semantic-convention issues reported in #86.

High-level changes:
- Handles current response and generation span payload shapes, including string response input, response IDs/models, token usage, tool definitions, and system instructions.
- Uses the current Agents SDK span-data module layout directly instead of carrying a legacy import fallback.
- Avoids emitting `gen_ai.operation.name = unknown` for non-GenAI Agents SDK spans such as task, turn, MCP list-tools, speech group, and custom spans.
- Removes `gen_ai.system` from OpenAI Agents spans and leaves successful spans with unset status.
- Emits required response/content fields when the SDK omits them, including `gen_ai.response.finish_reasons` and output-message `finish_reason` fallback values.
- Removes undocumented handoff GenAI operation/attributes from handoff spans.
- Emits `gen_ai.tool.call.id` for tool spans, falling back to the SDK span ID when no call ID is available.
- Stops labeling the root workflow span as `invoke_agent`, so arbitrary workflow trace names are not validated as invoke-agent span names.
- Enables metrics support with the shared GenAI duration/token histogram helpers and the configured `meter_provider`.
- Preserves custom/sandbox span attributes from `CustomSpanData.data` (for example `sandbox.*` and process exit attributes) without assigning them a GenAI operation.
- Refreshes the examples and supported `openai-agents` test range for the current Agents SDK.

This addresses the OpenAI Agents instrumentation failures called out in #86 without marking that issue closed here, since this PR still does not add a live weaver scenario for the package.

## Type of change

# How has this been tested?

Latest checks on the final pushed diff:
- [x] `uvx --with tox-uv tox -e py310-test-instrumentation-genai-openai_agents-oldest,py310-test-instrumentation-genai-openai_agents-latest,py311-test-instrumentation-genai-openai_agents-oldest,py311-test-instrumentation-genai-openai_agents-latest,py312-test-instrumentation-genai-openai_agents-oldest,py312-test-instrumentation-genai-openai_agents-latest,py313-test-instrumentation-genai-openai_agents-oldest,py313-test-instrumentation-genai-openai_agents-latest,py314-test-instrumentation-genai-openai_agents-oldest,py314-test-instrumentation-genai-openai_agents-latest,lint-instrumentation-genai-openai_agents -- -q`
- [x] `uvx --with tox-uv tox -e precommit`
- [x] `git diff --check`

Earlier validation for this PR:
- [x] `uvx --with tox-uv tox -e lint-license-header-check`
- [x] `uvx ruff format --check instrumentation/opentelemetry-instrumentation-genai-openai-agents`
- [x] `uvx --from towncrier==25.8.0 towncrier build --draft --version Unreleased`
- [x] Real SDK smoke checks with `openai-agents==0.17.0` and `openai-agents==0.17.2`
- [x] Example dependency install checks for the manual and zero-code examples using local editable packages.
- [x] Example smoke checks for content capture, manual instrumentation, and zero-code instrumentation.
- [x] Healthcare OTEL e2e smoke with local Collector/Jaeger using `prior_auth_confusion_ct`; verified service name, GenAI attrs, no unknown operation names, and sandbox attrs in Jaeger trace `dea30d9909ec5238e7347130f25dc4c2`.

# Checklist

See [CONTRIBUTING.md](https://github.com/open-telemetry/opentelemetry-python-genai/blob/main/CONTRIBUTING.md) for the style guide, changelog guidance, and more.

- [x] Unit tests added
- [x] Documentation updated

### DIFF
--- a/instrumentation/opentelemetry-instrumentation-genai-openai-agents/src/opentelemetry/instrumentation/genai/openai_agents/span_processor.py
+++ b/instrumentation/opentelemetry-instrumentation-genai-openai-agents/src/opentelemetry/instrumentation/genai/openai_agents/span_processor.py
@@ -19,43 +19,31 @@ References:
 
 from __future__ import annotations
 
-import importlib
 import logging
+from collections.abc import Mapping as MappingABC
+from collections.abc import Sequence as SequenceABC
 from dataclasses import dataclass
 from datetime import datetime, timezone
 from enum import Enum
 from typing import TYPE_CHECKING, Any, Dict, Iterator, Optional, Sequence
 from urllib.parse import urlparse
 
-from opentelemetry.util.genai.utils import gen_ai_json_dumps
-
-try:
-    from agents.tracing import Span, Trace, TracingProcessor
-    from agents.tracing.span_data import (
-        AgentSpanData,
-        FunctionSpanData,
-        GenerationSpanData,
-        GuardrailSpanData,
-        HandoffSpanData,
-        ResponseSpanData,
-        SpeechSpanData,
-        TranscriptionSpanData,
-    )
-except ModuleNotFoundError:  # pragma: no cover - test stubs
-    tracing_module = importlib.import_module("agents.tracing")
-    Span = getattr(tracing_module, "Span")
-    Trace = getattr(tracing_module, "Trace")
-    TracingProcessor = getattr(tracing_module, "TracingProcessor")
-    AgentSpanData = getattr(tracing_module, "AgentSpanData", Any)  # type: ignore[assignment]
-    FunctionSpanData = getattr(tracing_module, "FunctionSpanData", Any)  # type: ignore[assignment]
-    GenerationSpanData = getattr(tracing_module, "GenerationSpanData", Any)  # type: ignore[assignment]
-    GuardrailSpanData = getattr(tracing_module, "GuardrailSpanData", Any)  # type: ignore[assignment]
-    HandoffSpanData = getattr(tracing_module, "HandoffSpanData", Any)  # type: ignore[assignment]
-    ResponseSpanData = getattr(tracing_module, "ResponseSpanData", Any)  # type: ignore[assignment]
-    SpeechSpanData = getattr(tracing_module, "SpeechSpanData", Any)  # type: ignore[assignment]
-    TranscriptionSpanData = getattr(
-        tracing_module, "TranscriptionSpanData", Any
-    )  # type: ignore[assignment]
+from agents.tracing import Span, Trace, TracingProcessor
+from agents.tracing.span_data import (
+    AgentSpanData,
+    CustomSpanData,
+    FunctionSpanData,
+    GenerationSpanData,
+    GuardrailSpanData,
+    HandoffSpanData,
+    MCPListToolsSpanData,
+    ResponseSpanData,
+    SpeechGroupSpanData,
+    SpeechSpanData,
+    TaskSpanData,
+    TranscriptionSpanData,
+    TurnSpanData,
+)
 
 from opentelemetry.context import attach, detach
 from opentelemetry.metrics import Histogram, get_meter
@@ -73,6 +61,11 @@ from opentelemetry.trace import (
     Tracer,
     set_span_in_context,
 )
+from opentelemetry.util.genai.instruments import (
+    create_duration_histogram,
+    create_token_histogram,
+)
+from opentelemetry.util.genai.utils import gen_ai_json_dumps
 from opentelemetry.util.types import AttributeValue
 
 # Import all semantic convention constants
@@ -122,7 +115,6 @@ class GenAIOperationName:
     TRANSCRIPTION = "transcription"
     SPEECH = "speech_generation"
     GUARDRAIL = "guardrail_check"
-    HANDOFF = "agent_handoff"
     RESPONSE = "response"  # internal aggregator in current processor
 
     CLASS_FALLBACK = {
@@ -240,12 +232,19 @@ GEN_AI_TOOL_DEFINITIONS = "gen_ai.tool.definitions"
 GEN_AI_ORCHESTRATOR_AGENT_DEFINITIONS = "gen_ai.orchestrator.agent.definitions"
 GEN_AI_GUARDRAIL_NAME = "gen_ai.guardrail.name"
 GEN_AI_GUARDRAIL_TRIGGERED = "gen_ai.guardrail.triggered"
-GEN_AI_HANDOFF_FROM_AGENT = "gen_ai.handoff.from_agent"
-GEN_AI_HANDOFF_TO_AGENT = "gen_ai.handoff.to_agent"
 GEN_AI_EMBEDDINGS_DIMENSION_COUNT = "gen_ai.embeddings.dimension.count"
 GEN_AI_TOKEN_TYPE = _attr("GEN_AI_TOKEN_TYPE", "gen_ai.token.type")
 
 # ---- Normalization utilities (embedded from utils.py) ----
+
+_CUSTOM_ATTRIBUTE_RESERVED_PREFIXES = (
+    "gen_ai.",
+    "otel.",
+    "server.",
+    "telemetry.",
+)
+_CUSTOM_ATTRIBUTE_RESERVED_KEYS = {"span.kind"}
+_CUSTOM_ATTRIBUTE_VALUE_LENGTH_LIMIT = 8192
 
 
 def normalize_provider(provider: Optional[str]) -> Optional[str]:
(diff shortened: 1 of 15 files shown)


instrumentation/opentelemetry-instrumentation-openai-agents-v2/src/opentelemetry/instrumentation/openai_agents/span_processor.py
```diff
@@ -647,15 +657,15 @@ def _emit_content_events(
         return
 
     def _collect_system_instructions(
-        self, messages: Sequence[Any] | None
+        self, messages: Sequence[Any] | str | None
```
lzchen: Are `messages` still stored as a list in newer versions of the openai-agents sdk? If we are bumping the version to >= 0.17.0, we probably don't need to be backwards compatible in this case.

instrumentation/opentelemetry-instrumentation-genai-openai-agents/src/opentelemetry/instrumentation/genai/openai_agents/span_processor.py
```diff
@@ -1536,19 +1607,23 @@ def _get_operation_name(self, span_data: Any) -> str:
...
         if _is_instance_of(span_data, TranscriptionSpanData):
             return GenAIOperationName.TRANSCRIPTION
-        if _is_instance_of(span_data, SpeechSpanData):
+        if _is_instance_of(span_data, (SpeechGroupSpanData, SpeechSpanData)):
             return GenAIOperationName.SPEECH
         if _is_instance_of(span_data, GuardrailSpanData):
             return GenAIOperationName.GUARDRAIL
         if _is_instance_of(span_data, HandoffSpanData):
             return GenAIOperationName.HANDOFF
-        return "unknown"
```
lzchen, line 1551: Why this change?

instrumentation/opentelemetry-instrumentation-openai-agents-v2/src/opentelemetry/instrumentation/openai_agents/span_processor.py
alfozan: a raw string input can happen at the lower-level ResponseSpanData.input boundary, but it cannot contain role-scoped system instructions. So _collect_system_instructions doesn’t need to advertise str support.
will remove `str` type

alfozan:
Hi @lzchen, PR is updated. Could you please take another look when you have a chance?

lmolkova:
I think it would be best to fix existing issues in OpenAI Agents instrumentation before adding more features to it - https://github.com/open-telemetry/opentelemetry-python-genai/issues/86

alfozan:
> I think it would be best to fix existing issues in OpenAI Agents instrumentation before adding more features to it - #86

Hi @lmolkova, agreed. I updated this PR to address the conformance failures from #86.

alfozan:
Hi @lzchen, @lmolkova. Could you please take a look when you have a chance?

alfozan:
Hi @nagkumar91, @hectorhdzg, @rads-1996, could one of you take a look at this PR when you have a chance?

instrumentation/opentelemetry-instrumentation-genai-openai-agents/src/opentelemetry/instrumentation/genai/openai_agents/span_processor.py
```diff
@@ -240,13 +232,22 @@ def _attr(name: str, fallback: str) -> str:
...
-GEN_AI_HANDOFF_FROM_AGENT = "gen_ai.handoff.from_agent"
-GEN_AI_HANDOFF_TO_AGENT = "gen_ai.handoff.to_agent"
 GEN_AI_EMBEDDINGS_DIMENSION_COUNT = "gen_ai.embeddings.dimension.count"
 GEN_AI_TOKEN_TYPE = _attr("GEN_AI_TOKEN_TYPE", "gen_ai.token.type")
 
+_DEFAULT_FINISH_REASON = "unknown"
+
 # ---- Normalization utilities (embedded from utils.py) ----
 
+_CUSTOM_ATTRIBUTE_RESERVED_PREFIXES = (
```
rads-1996, line 240: Would it make more sense to move all of these constants to a constant.py file instead of adding it in this one?

alfozan: I kept them here because they are only used in `span_processor.py`, and this package does not currently have a constants module. I’d prefer to avoid the extra refactor in this PR, but we can do it in a follow-up PR once this is merged.

instrumentation/opentelemetry-instrumentation-genai-openai-agents/src/opentelemetry/instrumentation/genai/openai_agents/span_processor.py
```diff
@@ -240,13 +232,22 @@ def _attr(name: str, fallback: str) -> str:
 GEN_AI_ORCHESTRATOR_AGENT_DEFINITIONS = "gen_ai.orchestrator.agent.definitions"
 GEN_AI_GUARDRAIL_NAME = "gen_ai.guardrail.name"
 GEN_AI_GUARDRAIL_TRIGGERED = "gen_ai.guardrail.triggered"
-GEN_AI_HANDOFF_FROM_AGENT = "gen_ai.handoff.from_agent"
-GEN_AI_HANDOFF_TO_AGENT = "gen_ai.handoff.to_agent"
 GEN_AI_EMBEDDINGS_DIMENSION_COUNT = "gen_ai.embeddings.dimension.count"
 GEN_AI_TOKEN_TYPE = _attr("GEN_AI_TOKEN_TYPE", "gen_ai.token.type")
 
+_DEFAULT_FINISH_REASON = "unknown"
```
lzchen: Do we need a default if [gen_ai.response.finish_reasons](https://github.com/open-telemetry/semantic-conventions-genai/blob/72d5815a7823f323be3d2298fa7d4e4e1d1ca9fb/docs/registry/attributes/gen-ai.md) is recommended and not mandatory?

instrumentation/opentelemetry-instrumentation-genai-openai-agents/src/opentelemetry/instrumentation/genai/openai_agents/span_processor.py
```diff
@@ -1284,19 +1400,72 @@ def _get_span_kind(self, span_data: Any) -> SpanKind:
...
+
+        return "span"
+
     def on_trace_start(self, trace: Trace) -> None:
         """Create root span when trace starts."""
         if self._tracer:
             attributes = {
                 GEN_AI_PROVIDER_NAME: self.system_name,
-                GEN_AI_SYSTEM_KEY: self.system_name,
-                GEN_AI_OPERATION_NAME: GenAIOperationName.INVOKE_AGENT,
```
lzchen, line 1299: This is kind of problematic as `[gen_ai.operation.name](https://github.com/open-telemetry/semantic-conventions-genai/blob/main/docs/registry/attributes/gen-ai.md)` is a required field. I understand that the root workflow span shouldn't be marked as `invoke_agent` however. @lmolkova  any thoughts?

instrumentation/opentelemetry-instrumentation-genai-openai-agents/src/opentelemetry/instrumentation/genai/openai_agents/span_processor.py
```diff
@@ -1319,8 +1488,6 @@ def on_trace_start(self, trace: Trace) -> None:
     def on_trace_end(self, trace: Trace) -> None:
         """End root span when trace ends."""
         if root_span := self._root_spans.pop(trace.trace_id, None):
-            if root_span.is_recording():
-                root_span.set_status(Status(StatusCode.OK))
```
lzchen, line 1323: A lot of this behavior is undefined. Should we mark root spans as `OK` status if the end properly? @lmolkova @aabmass wdyt?

instrumentation/opentelemetry-instrumentation-genai-openai-agents/src/opentelemetry/instrumentation/genai/openai_agents/span_processor.py
alfozan: Updated, this now only emits `gen_ai.response.finish_reasons` when the SDK provides a finish reason.

lmolkova (changes requested):
Thanks for the contribution! This should be covered with this PR - https://github.com/open-telemetry/opentelemetry-python-genai/pull/90, would appreciate your review!

alfozan:
Closing this in favor of #90, which was just merged and closed #86.
Thanks everyone for the reviews and guidance.

PR #59, opened 2026-05-21
Skip changelog check for dependency PRs

lmolkova:
not sure why, but copilot could not create PR from the chat session I started


trask (approved):

trask:
> not sure why, but copilot could not create PR from the chat session I started

I've enabled copilot coding agent in both of the new GenAI repos now, hopefully that will fix.

PR #64, opened 2026-05-22
Remove workflow file generation

trask:
Solves the issue with Renovate trying to update the generated workflow files.


lmolkova (approved):
nice!! @aabmass @lzchen @MikeGoldsmith is this something we can also leverage in python contrib?

PR #71, opened 2026-05-22
Less Renovate rebasing

trask:


lmolkova (approved):

trask:
I'm realizing this may only be a problem because of "Require branches to be up to date before merging", which will go away soon with

- https://github.com/open-telemetry/opentelemetry-python-genai/pull/65
- https://github.com/open-telemetry/admin/pull/677

PR #92, opened 2026-05-26
refactor: responses and messages genai streams integration

eternalcuriouslearner:
# Description

Migrate Responses and Messages streams to common gen-ai streams

Fixes # (issue)

## Type of change

# How has this been tested?

configuration.

- existing tests are running successfully.

# Checklist

- [ ] Unit tests added
- [ ] Documentation updated


instrumentation/opentelemetry-instrumentation-genai-openai/src/opentelemetry/instrumentation/genai/openai/response_wrappers.py
```diff
@@ -91,63 +95,54 @@ def __getattr__(self, name: str):
...
 
-    def close(self) -> None:
-        try:
-            self.stream.close()
-        finally:
-            self._stop(None)
+    def _process_chunk(
+        self, chunk: "ResponseStreamEvent[TextFormatT]"
+    ) -> None:
+        self.process_event(chunk)
```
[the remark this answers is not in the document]
eternalcuriouslearner: This is not relevant.

instrumentation/opentelemetry-instrumentation-genai-openai/src/opentelemetry/instrumentation/genai/openai/response_wrappers.py
```diff
@@ -354,6 +342,15 @@ def parse(self) -> "AsyncResponseStreamWrapper[TextFormatT]":
...
         return self
 
+    @property
+    def stream(self) -> "AsyncResponseStream[TextFormatT]":
+        return self._self_stream
+
+    @stream.setter
+    def stream(self, stream: "AsyncResponseStream[TextFormatT]") -> None:
+        self._self_stream = stream
+        self._self_aiter = aiter(stream)
```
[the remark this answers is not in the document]
eternalcuriouslearner: Current minimum supported version is 3.10. We are good. Hence disregarding this.

instrumentation/opentelemetry-instrumentation-genai-anthropic/src/opentelemetry/instrumentation/genai/anthropic/wrappers.py
```diff
@@ -199,80 +155,94 @@ def _process_chunk(
...
+        return self._self_stream
+
+    @stream.setter
+    def stream(
+        self,
+        stream: AsyncStream[RawMessageStreamEvent]
+        | AsyncMessageStream[ResponseFormatT],
+    ) -> None:
+        self._self_stream = stream
+        self._self_aiter = aiter(stream)
```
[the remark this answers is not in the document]
eternalcuriouslearner: Same as above.

instrumentation/opentelemetry-instrumentation-genai-openai/src/opentelemetry/instrumentation/genai/openai/response_wrappers.py
```diff
@@ -91,63 +95,52 @@ def __getattr__(self, name: str):
...
+        self._self_response_finalized = False
 
-    def __enter__(self) -> "ResponseStreamWrapper":
-        return self
+    def _stop(
+        self, result: "ParsedResponse[TextFormatT] | Response | None"
+    ) -> None:
+        if self._self_response_finalized:
+            return
+        with self._safe_instrumentation("response attribute extraction"):
```
lmolkova: I think we discussed it before, we should never throw from instrumentation code (unless it's exception from instrumented code) and therefore we should never catch exceptions from instrumentation code

instrumentation/opentelemetry-instrumentation-genai-anthropic/src/opentelemetry/instrumentation/genai/anthropic/wrappers.py
```diff
@@ -105,174 +109,161 @@ def message(self) -> Message:
...
-            self.invocation, self._message, self._capture_content
+            self._self_invocation,
+            self._self_message,
+            self._self_capture_content,
         )
-        self.invocation.stop()
-        self._finalized = True
+        self._self_invocation.stop()
+        self._self_message_finalized = True
+        self._self_finalized = True
```
lmolkova: are these two flags ever different? do we need both?

eternalcuriouslearner: These flags represent two different layers of finalization.

  `_self_finalized` is owned by `SyncStreamWrapper` / `AsyncStreamWrapper`. It tracks the shared stream wrapper lifecycle: whether the wrapper has already handled stream end or stream failure. The base
  sets it before invoking the provider hook:

```python
  def _finalize_success(self) -> None:
      if self._self_finalized:
          return
      self._self_finalized = True
      self._on_stream_end()
```

The provider-specific flag tracks telemetry finalization: whether this provider wrapper has already called invocation.stop() or invocation.fail().

For example, OpenAI Responses can finalize telemetry before the underlying stream is exhausted when it receives a terminal event:

```python
 def process_event(self, event):
      if event.type == "response.completed":
          self._stop(event.response)
          return

  def _stop(self, result):
      if self._self_response_telemetry_finalized:
          return

      _set_response_attributes(...)
      self._self_invocation.stop()
      self._self_response_telemetry_finalized = True
```

  Later, when the underlying stream naturally ends, the shared wrapper still calls _on_stream_end():

```python
  def _on_stream_end(self):
      self._stop(None)
```

At that point the provider telemetry flag prevents a duplicate invocation.stop().

We cannot use _self_finalized for this provider guard because the shared base sets _self_finalized = True before calling _on_stream_end() / _on_stream_error(). If _stop() checked only _self_finalized, normal stream finalization would return before recording telemetry.

So the split is:

- _self_finalized: shared stream lifecycle has finalized.
- _self_response_telemetry_finalized / _self_message_telemetry_finalized: provider telemetry has finalized.

I updated the individual stream flag names from _self_response_finalized / _self_message_finalized to _self_response_telemetry_finalized / _self_message_telemetry_finalized to make this distinction explicit and avoid confusion. The provider flag avoids duplicate telemetry finalization, while the shared flag avoids duplicate stream lifecycle finalization. Lmk if you have any questions. I am open to change this to simplify the flow.

instrumentation/opentelemetry-instrumentation-genai-anthropic/src/opentelemetry/instrumentation/genai/anthropic/wrappers.py
```diff
@@ -105,174 +109,161 @@ def message(self) -> Message:
...
 
     def _fail(self, exc: BaseException) -> None:
-        if self._finalized:
+        if self._self_message_finalized:
             return
-        self.invocation.fail(exc)
-        self._finalized = True
+        self._self_invocation.fail(exc)
+        self._self_message_finalized = True
+        self._self_finalized = True
```
lzchen: Instead of directly modifying the base class attribute, it be useful to have a `_set_finalized` api in the base class in the future. Having instrumentations have knowledge of the base class attributes is awkward. Probably for a different pr.

instrumentation/opentelemetry-instrumentation-genai-openai/src/opentelemetry/instrumentation/genai/openai/response_wrappers.py
```diff
@@ -91,63 +95,52 @@ def __getattr__(self, name: str):
...
-        try:
-            event = next(self.stream)
-        except StopIteration:
-            self._stop(None)
-            raise
-        except Exception as error:
-            self._fail(str(error), type(error))
-            raise
-        with self._safe_instrumentation("event processing"):
-            self.process_event(event)
```
lzchen, line 149: This will probably produce a more generic error message instead of the one generated by self._safe_instrumentation now right?

eternalcuriouslearner: Yes @lzchen self._safe_instrumentation is probably an overkill so I removed it. I discussed it with @lmolkova on the anthropic pr and removed it from messages instrumentation. So I am removing it from here.

instrumentation/opentelemetry-instrumentation-genai-anthropic/src/opentelemetry/instrumentation/genai/anthropic/wrappers.py
eternalcuriouslearner: @lzchen   I updated the provider-specific flag names to make the distinction clearer:

  - `_self_response_finalized` -> `_self_response_telemetry_finalized`
  - `_self_message_finalized` -> `_self_message_telemetry_finalized`

I also removed the direct writes to the shared base `_self_finalized` flag from the provider wrappers, so Responses/Messages no longer modify the base class lifecycle attribute directly. The shared `SyncStreamWrapper` / `AsyncStreamWrapper` remain the only owners of `_self_finalized`.

Do you still think we should add a `_set_finalized()` in the shared API?

instrumentation/opentelemetry-instrumentation-genai-anthropic/src/opentelemetry/instrumentation/genai/anthropic/wrappers.py
```diff
@@ -105,174 +109,161 @@ def message(self) -> Message:
...
             current_snapshot=cast(
-                "ParsedMessage[ResponseFormatT] | None", self._message
+                "ParsedMessage[ResponseFormatT] | None", self._self_message
             ),
         )
 
+    def _current_message_snapshot(
+        self,
+    ) -> ParsedMessage[ResponseFormatT] | None:
+        return None
```
lmolkova: is this method necessary? Is it correct to always return None here?

also it seems sync tests don't actually validate proper accumulation https://github.com/eternalcuriouslearner/opentelemetry-python-genai/blob/262790f81db55a6a660b8414ddc5d9ae1cc85e5a/instrumentation/opentelemetry-instrumentation-genai-anthropic/tests/test_sync_messages.py#L469

I don't see any async streaming tests - how do we know it works ? 🤔

eternalcuriouslearner: Originally `_current_message_snapshot()` was added as a small hook because `_MessagesStreamMixin` did not own a typed `stream` property. The concrete sync/async wrappers had `stream`, so the mixin called `_current_message_snapshot()` and each wrapper could provide the stream-specific lookup.

I think the current implementation returning `None` was little confusing because the mixin is only used with concrete wrappers that do have a stream. I updated this to remove the default method and have `_process_chunk()` read `current_message_snapshot` from the concrete wrapped stream via a small protocol/cast. Lmk what you think about it.

I also strengthened the tests:
- sync `Messages.stream` content tests now assert the exact accumulated output text (`Hello!`) instead of only checking that parts exist.
- sync and async wrapper tests now verify that `current_message_snapshot` is captured into `_self_message`.

Async `Messages.stream()` instrumentation is not wired end-to-end today, so the async coverage is wrapper-level rather than an instrumentor/VCR test.

lmolkova (approved):

lzchen (approved):

PR #117, opened 2026-06-08
[opentelemetry-util-genai] Remove experimental sem conv flag, usage, and references. Cleanup tests

DylanRussell:
# Description

This PR removes the experimental sem conv flag, usage, and references from `opentelemetry-util-genai`. Also does cleanup on some of the associated tests..

I also removed the last vestiges of the flag from the Google GenAi instrumentation which were blocked on it being removed from the utils package first..

The other instrumentation packages need to migrate off some of the deprecated code before this can be submitted..

Fixes # (issue)

## Type of change

- [ x] Bug fix (non-breaking change which fixes an issue)

# How has this been tested?

Unit tests

# Checklist

- [ x] Changelog updated if the change requires an entry
- [ x] Unit tests added
- [ x] Documentation updated


util/opentelemetry-util-genai/src/opentelemetry/util/genai/utils.py
```diff
@@ -22,24 +17,11 @@
 logger = logging.getLogger(__name__)
 
 
-def is_experimental_mode() -> bool:
```
lmolkova, line 25: there are a lot of tests and maybe instrumentations in this repo that depend on it. let's keep it for now and return True always. It might also be used by old versions in the contrib repo that we want to keep runtime compatibility with until we deprecate them.

util/opentelemetry-util-genai/tests/test_events_options.py
```diff
@@ -1,240 +0,0 @@
...
-            _OpenTelemetrySemanticConventionStability._initialized = False
-            _OpenTelemetrySemanticConventionStability._initialize()
-            return test_case(*args, **kwargs)
-
-        return wrapper
-
-    return decorator
-
-
-class TestShouldEmitEvent(unittest.TestCase):
```
lmolkova, line 43: are we moving these tests somewhere? I assume they still validate something when flag is on.

util/opentelemetry-util-genai/src/opentelemetry/util/genai/utils.py
```diff
@@ -65,62 +47,34 @@ def should_emit_event() -> bool:
...
-            return True
-        # NO_CONTENT and SPAN_ONLY don't require events, so default to False
-        return False
-    except ValueError:
-        # If get_content_capturing_mode raises ValueError (not in experimental mode),
-        # default to False
-        return False
-
-
-def should_capture_content_on_spans_in_experimental_mode() -> bool:
```
lmolkova, line 103: is this used by something in contrib? If yes we should not remove it from public API until we deprecate those

util/opentelemetry-util-genai/src/opentelemetry/util/genai/handler.py
```diff
@@ -107,30 +97,6 @@ def __init__(
...
-                "true",
-                "span_only",
-                "event_only",
-                "span_and_event",
-            )
-        self._capture_content = content_enabled or not isinstance(
-            self._completion_hook, _NoOpCompletionHook
-        )
-
-    def should_capture_content(self) -> bool:
```
lmolkova, line 127: this is used by instrumentation libs and is still useful and necessary for them regardless of the experimental flag.

Calling get_content_capturing_mode would result in extra logic for callers and need to read env var potentially on a hot path.

util/opentelemetry-util-genai/tests/test_handler_completion_hook.py
```diff
@@ -171,101 +152,6 @@ def stamp_ref(*, log_record, **kwargs):
             "s3://bucket/inputs.json",
         )
 
-    @patch.dict(
-        os.environ, {OTEL_INSTRUMENTATION_GENAI_CAPTURE_MESSAGE_CONTENT: ""}
-    )
-    def test_should_capture_content_false_by_default(self):
```
lmolkova, line 177: what happens now on malformed env var? do we throw? do we not capture content? do we have tests for it if we're removing those?

lmolkova (commented):
Thanks! Just some questions and suggestions to keep public API

util/opentelemetry-util-genai/src/opentelemetry/util/genai/utils.py
DylanRussell: getting rid of all the dependencies in this repo -- we are on track to complete that task.. I wasn't planning to submit this until that is done.

maintaining compatibility with the old repo -- that sounds complicated.. we probably want to set up cross-repo tests (like we have between otel-python and otel-python-contrib) if we want to enforce that, but I feel like we should just deprecate the stuff in the old repo and not worry about that..

I've added it back with a deprecation warning..

util/opentelemetry-util-genai/tests/test_events_options.py
DylanRussell: Yes I refactored them and moved them into `test_utils.py`

util/opentelemetry-util-genai/src/opentelemetry/util/genai/utils.py
DylanRussell: I don't see any usages in contrib for this method -- ok to remove it ?

util/opentelemetry-util-genai/src/opentelemetry/util/genai/handler.py
DylanRussell: Ok I see how this is being used now.. I got very confused by the name and the logic here.. 

So if this is `True` instrumentations should set the "message attributes" on the `invocation` objects, and let this util library decide when/where to add those attributes..

util/opentelemetry-util-genai/tests/test_handler_completion_hook.py
DylanRussell: Ok I've added back in these tests now that I understand what this function is for.. I refactored them a little

util/opentelemetry-util-genai/src/opentelemetry/util/genai/handler.py
DylanRussell: I've returned this method to the code..

util/opentelemetry-util-genai/src/opentelemetry/util/genai/utils.py
```diff
@@ -54,6 +36,14 @@ def get_content_capturing_mode() -> ContentCapturingMode:
         return ContentCapturingMode.NO_CONTENT
 
 
+def is_experimental_mode() -> bool:
+    """
+    Kept for backwards compatibility. The utils in this library only support the experimental mode sem convs now.
```
lzchen, line 41: If we are removing all experimental flags, why keep this? I thought now was the time to make breaking changes?

instrumentation/opentelemetry-instrumentation-google-genai/tests/common/base.py
```diff
@@ -3,31 +3,19 @@
...
-        with patch.dict(
-            "os.environ",
-            {
-                "OTEL_INSTRUMENTATION_GENAI_CAPTURE_MESSAGE_CONTENT": "SPAN_AND_EVENT",
-                "OTEL_SEMCONV_STABILITY_OPT_IN": "gen_ai_latest_experimental",
-            },
-        ):
-            _OpenTelemetrySemanticConventionStability._initialized = False
-            _OpenTelemetrySemanticConventionStability._initialize()
+        os.environ["OTEL_INSTRUMENTATION_GENAI_CAPTURE_MESSAGE_CONTENT"] = (
```
lzchen: Nit: Why not keep the patch instead of changing real env vars?

util/opentelemetry-util-genai/src/opentelemetry/util/genai/utils.py
lmolkova: because it's a public API used by libs in the contrib. 

The plan:
- release this repo
- release the last version on libs in the contrib (they should still work with these utils) and deprecated them along with the release
- after that we're free to party on and break runtime compatibility with them.

You can argue that everything is experimental and  not expected to be compatible, but really this version of util here is likely the only thing every lib in contrib repo is compatible with.

lzchen (approved):

lmolkova (approved):

util/opentelemetry-util-genai/src/opentelemetry/util/genai/utils.py
DylanRussell: Why should the latest version in contrib work with these utils ? Can't we require a version of utils that is only from the old repo ?

instrumentation/opentelemetry-instrumentation-google-genai/tests/common/base.py
DylanRussell: Ack good point.. switched back to using patch.dict

util/opentelemetry-util-genai/src/opentelemetry/util/genai/utils.py
DylanRussell: Resolving since I think this PR is ready to merge either way

PR #119, opened 2026-06-09
[opentelemetry-instrumentation-google-genai] Add response id attribute, cleanup `dictutil` code

DylanRussell:
# Description

Add response id attribute to inference span and completion event for Google GenAi...

See https://opentelemetry.io/docs/specs/semconv/gen-ai/gen-ai-spans/

I also drastically cleaned up the `dict_util` code which converts the `GenerateContentConfig` object into OTEL Attributes dictionary.. I deleted some unused functionality and cleaned up the code. 

## Type of change

- [ x] New feature (non-breaking change which adds functionality)

# How has this been tested?

Unit tests

# Checklist

- [ x] Unit tests added
- [ x] Documentation updated


lmolkova (approved):

eternalcuriouslearner (approved):

DylanRussell:
@eternalcuriouslearner @lmolkova mind taking another look ? Ended up doing a bunch of code cleanup after noticing some messy code here..

util/opentelemetry-util-genai/src/opentelemetry/util/genai/_inference_invocation.py
```diff
@@ -93,6 +93,9 @@ def __init__(
         self.cache_creation_input_tokens: int | None = None
         self.cache_read_input_tokens: int | None = None
         self.tool_definitions: list[ToolDefinition] | None = None
+        self.top_k: float | None = None
+        self.request_choice_count: int | None = None
```
eternalcuriouslearner, line 96: note(non-blocking): Not sure, but I need to check if these attributes should be added from both openai and anthropic instrumentations.

eternalcuriouslearner: Only question, can we remove this from this pr and add it separately?

instrumentation/opentelemetry-instrumentation-google-genai/src/opentelemetry/instrumentation/google_genai/generate_content.py
```diff
@@ -279,43 +278,44 @@ async def _to_tool_definition_async(
...
+            f"{GCP_GENAI_OPERATION_CONFIG}.frequency_penalty",
+            f"{GCP_GENAI_OPERATION_CONFIG}.presence_penalty",
+            f"{GCP_GENAI_OPERATION_CONFIG}.seed",
         },
     )
-    response_mime_type = config.get("response_mime_type")
-    if response_mime_type:
+    attributes = {k: v for k, v in attributes.items() if allow_list.allowed(k)}
+    if response_mime_type := config.get("response_mime_type"):
         if response_mime_type == "text/plain":
```
eternalcuriouslearner: question(non-blocking): can we do something like this here 

```python
    attributes = {k: v for k, v in attributes.items() if allow_list.allowed(k)}
    if response_mime_type := config.get("response_mime_type"):
        if response_mime_type == "text/plain":
            invocation.output_type = "text"
        elif response_mime_type == "application/json":
            invocation.output_type = "json"
        else:
            invocation.output_type = response_mime_type
    invocation.attributes.update(attributes)
```

eternalcuriouslearner (approved):

util/opentelemetry-util-genai/src/opentelemetry/util/genai/_inference_invocation.py
DylanRussell: We could but why bother doing it in a separate PR ? I can open an issue to track adding them to openai/anthropic if that helps ?

instrumentation/opentelemetry-instrumentation-google-genai/src/opentelemetry/instrumentation/google_genai/generate_content.py
DylanRussell: Yeah for sure.. I meant to do that and got distracted updating the invocation type.. done now

util/opentelemetry-util-genai/src/opentelemetry/util/genai/_inference_invocation.py
DylanRussell: Opened https://github.com/open-telemetry/opentelemetry-python-genai/issues/138

DylanRussell:
The failing tests In the CI will be fixed by https://github.com/open-telemetry/opentelemetry-python-genai/pull/117 and or https://github.com/open-telemetry/opentelemetry-python-genai/pull/133

DylanRussell:
I think this PR should be good to be merged now

instrumentation/opentelemetry-instrumentation-google-genai/src/opentelemetry/instrumentation/google_genai/dict_util.py
```diff
@@ -126,168 +55,63 @@ def _flatten_compound_value_using_json(
...
-    result[_concat_key(key_prefix, "length")] = len(lst)
-    for index, value in enumerate(lst):
-        full_key = f"{key_prefix}[{index}]"
-        flattened = _flatten_value(
-            full_key,
-            value,
-            exclude_keys=exclude_keys,
-            rename_keys=rename_keys,
-            flatten_functions=flatten_functions,
+    for key, value in [(k, v) for k, v in d.items() if k not in exclude_keys]:
```
aabmass: nit probably more readable to either do the checks in the loop

```py
for key, value in ...:
  if k not in exclude_keys: continue

  ...
```

or alternatively just use set key difference

```py
for key in d.keys() - exclude_keys:
  value = d[key]
  ...
```

instrumentation/opentelemetry-instrumentation-google-genai/src/opentelemetry/instrumentation/google_genai/generate_content.py
```diff
@@ -279,57 +278,53 @@ async def _to_tool_definition_async(
...
+    invocation.frequency_penalty = config.get("frequency_penalty")
+    invocation.presence_penalty = config.get("presence_penalty")
+    invocation.seed = config.get("seed")
+    if response_mime_type := config.get("response_mime_type"):
+        if response_mime_type == "text/plain":
+            invocation.output_type = "text"
+        elif response_mime_type == "application/json":
+            invocation.output_type = "json"
+        else:
+            invocation.output_type = response_mime_type
```
aabmass: Any way we could make this type safe? IIRC, `GenerateContentConfigOrDict` has a pydantic model variant and you can use `model_validate(config)`

WDYT?

aabmass (approved):
Thanks!

instrumentation/opentelemetry-instrumentation-google-genai/src/opentelemetry/instrumentation/google_genai/dict_util.py
DylanRussell: Ack, done

instrumentation/opentelemetry-instrumentation-google-genai/src/opentelemetry/instrumentation/google_genai/generate_content.py
DylanRussell: Ok your comment made me realize we were parsing the GenerateContentConfig inconsistently all over the place, instead of just doing it in one place, one way and then passing it around..

Ultimately decided to try to use `model_validate` when it's a dict, and return a base empty object when that fails, and pass that around so everything just expects a GenerateContentConfig object..

When I call the real `generate_content` I pass in the original unmodified config (that might have failed to parse, since I don't want to override how the real library deals with a invalid config being passed in..) unless we got a valid config with tools which we wrapped to get the execute tool span..

Issue #141, opened 2026-06-15
Tracking: migrate OpenInference Python instrumentations into this repo


lmolkova:
This issue tracks porting the OpenInference Python instrumentation packages into
`opentelemetry-python-genai`, so that GenAI instrumentation is built on the shared
`opentelemetry-util-genai` utilities and emits telemetry that conforms to the
[GenAI semantic conventions](https://github.com/open-telemetry/semantic-conventions-genai).

> [!IMPORTANT]
> - Use the instrumentations from https://github.com/open-telemetry/donation-openinference/ as the source
> (once seeded). DO NOT use https://github.com/Arize-ai/openinference/.
> 
> - Before starting work on a specific library, please ask to be assigned in the table —
for example, by leaving a comment on this issue.
>
> - Prioritize packages with more downloads and the ones your organization is interested in.
>
> - Make sure to link this issue from the migration PRs.

Candidate packages (ordered by recent PyPI downloads):

| OI package | Downloads last month | Assignee | Status | Notes |
| :---- | ----: | :---- | :---- | :---- |
| openinference-instrumentation-langchain | 2,303,629 | @wrisa @keith-decker @lzchen @rads-1996 | https://github.com/open-telemetry/opentelemetry-python-genai/issues/543 |  |
| openinference-instrumentation-openai | 1,988,421 | @eternalcuriouslearner | https://github.com/open-telemetry/opentelemetry-python-genai/issues/541 |  |
| openinference-instrumentation-pydantic-ai | 777,820 |  | Won't port | Has [native instrumentation](https://ai.pydantic.dev/logfire/), see [conformance](https://github.com/open-telemetry/semantic-conventions-conformance/blob/main/scenarios/gen-ai/python/pydantic-ai/native/data.json) |
| openinference-instrumentation-anthropic | 715,867 | @eternalcuriouslearner | https://github.com/open-telemetry/opentelemetry-python-genai/issues/542 |  |
| openinference-instrumentation-claude-agent-sdk | 625,120 | @gh-raju | #297 (closed, not merged) |  |
| openinference-instrumentation-openai-agents | 576,006 | @eternalcuriouslearner | https://github.com/open-telemetry/opentelemetry-python-genai/issues/549 |  |
[...]

PR #142, opened 2026-06-16
Update agent instructions

lmolkova:
1. Cover new streaming helpers in AGENT.md and copilot review instructions
2. Add basic guidance on adding new package to workspace
3. More info on tox commands
4. recommend TODO comment in AI-generated cassettes
5. More guidance for writing tests

This is in preparation to https://github.com/open-telemetry/opentelemetry-python-genai/issues/141


eternalcuriouslearner (approved):

AGENTS.md
```diff
@@ -112,15 +188,29 @@ Apply to packages under `instrumentation/`.
...
+  re-implementing provider/exporter/VCR plumbing. Import scrub helpers
+  (`scrub_response_headers` / `scrub_response_headers_overwrite`) from
+  `opentelemetry.test_util_genai.vcr` where a `vcr_config` needs them.
+- Drive instrumentation in tests through the shared `instrument` context manager from
+  `opentelemetry.test_util_genai.instrumentor` — `instrument(SomeInstrumentor(),
+  tracer_provider=…, logger_provider=…, meter_provider=…, semconv=…, content_capture=…)`. It sets
+  the content-capture (`OTEL_INSTRUMENTATION_GENAI_CAPTURE_MESSAGE_CONTENT`) env var *before* instrumenting and
+  restores them after, so a package's `instrument_*` fixtures don't manage that env themselves
+  (`TelemetryHandler` snapshots content-capture at construction, so the env must be set before it
+  is built).
```
eternalcuriouslearner, line 210: suggestion(non-blocking): Should we add a note here to include test scenarios verifying that both stream exceptions and user-defined exceptions are bubbled up to the context manager when process_chunk is implemented?

AgentGymLeader (commented):
Read through it — really nice change. The streaming wrapper part is clear: the three hooks and the "don't call `stop()` when the SDK hands back the stream" rule are nicely laid out. I also really like the "always ship a scenario even if a gap blocks it, and never drop one that fails today" policy.

Just one tiny thing — in the root `AGENTS.md` the streaming section's last paragraph runs straight into `### Exception handling` with no blank line, while the same block in `util/opentelemetry-util-genai/AGENTS.md` has the blank line before `### Anti-patterns`, so adding one in the root would match. Otherwise LGTM 🙂

trask (approved):

AGENTS.md
lmolkova: great point, added

PR #148, opened 2026-06-18
[chore] Remove exception handling around instrumentation code

lmolkova:
Remove `_safe_finalize_success`, `_safe_finalize_failure` methods and other places where we attempt to handle exceptions  raised by instrumentation code. 

Instrumentation code is not expected to raise. We must write it in the way that it doesn't.


eternalcuriouslearner (approved):
LGTM!!

PR #150, opened 2026-06-18
Don't call set attribute twice for the same key and make start attributes unsettable

lmolkova:


instrumentation/opentelemetry-instrumentation-genai-openai/tests/test_embedding_invocation_unit.py
```diff
@@ -77,8 +75,6 @@ def test_model_preserved_when_provided(handler, span_exporter):
     )
     invocation.stop()
 
-    assert invocation.request_model == "text-embedding-3-small"
```
lmolkova, line 80: attributes validated below

util/opentelemetry-util-genai/src/opentelemetry/util/genai/_agent_invocation.py
```diff
@@ -99,37 +99,35 @@ def __init__(
         self.system_instruction: list[MessagePart] = []
         self.tool_definitions: list[ToolDefinition] | None = None
 
-        self._start(self._get_base_attributes())
+        self._start(self._get_start_attributes())
 
-    def _get_base_attributes(self) -> dict[str, Any]:
+    @property
+    def agent_name(self) -> str | None:
```
lmolkova, line 104: this one is used by langchain, convert to read-only property

util/opentelemetry-util-genai/src/opentelemetry/util/genai/_agent_invocation.py
```diff
@@ -64,12 +64,12 @@ def __init__(
             else _operation_name,
             span_kind=span_kind,
         )
-        self.provider = provider
-        self.request_model = request_model
-        self.server_address = server_address
-        self.server_port = server_port
+        self._provider: str = provider
```
DylanRussell: so `_` means don't directly set this, instead pass it in on initialization.. Is that enforced somehow or is it just a convention ? Does it cause a lint error at least ?

lmolkova: it a conventions, pylint can detect it, but is not configured to do it in this and contrib repos. I can look into enabling it, but it's a common convention in python to mark internal API, it's what we do for all private methods on these classes. I don't think it needs to be bullet-proof

DylanRussell: Ack, I think it's fine although in some places in utils we allow instrumentations to access `_` (like the streaming wrappers although maybe that's a special case).. anyway I think it's better then nothing

DylanRussell (approved):

util/opentelemetry-util-genai/src/opentelemetry/util/genai/_inference_invocation.py
```diff
@@ -118,20 +118,20 @@ def _get_finish_reasons(self) -> list[str] | None:
...
+            (server_attributes.SERVER_ADDRESS, self._server_address),
+            (server_attributes.SERVER_PORT, self._server_port),
         )
         return {
             GenAI.GEN_AI_OPERATION_NAME: self._operation_name,
             **{k: v for k, v in optional_attrs if v is not None},
         }
 
     def _get_attributes(self) -> dict[str, Any]:
-        attrs = self._get_base_attributes()
```
lzchen, line 132: I believe `_get_attributes()` is called in `_apply_finish()` when the inference ends and is also set on the span when finished. Is this setting the starting attributes twice then?

aabmass (approved):

PR #153, opened 2026-06-19
fix(genai-utils): use AttributeValue instead of Any for attributes/metric_attributes

### LINKED ISSUE #41: [genai-utils] Remove all usages of `typing.Any`
None of the utils should accept `Any`, as that is an impossible type for the utils to serialize to the OTel proto.. Instead it should accept the types listed in https://github.com/open-telemetry/opentelemetry-python/blob/7fab34d4f49c1cd67df07fa53471c0ae6c269008/opentelemetry-api/src/opentelemetry/util/types.py#L20 -- this is already done in some places in the utils but not everywhere.. We should update the code..

Span attribute values currently must be of the `AttributeValue` type (at some point this will change to `AnyValue`

Not certain what metrics / logs uses but should be easy to figure that out..

bhumikadangayach:
Addresses the span/metric attribute portion of #41.

Replaces `typing.Any` with `opentelemetry.util.types.AttributeValue` for `attributes` and `metric_attributes` fields and their related getter methods, since these values are ultimately serialized to OTel span/metric attributes and `AttributeValue` is the correct constraint.

Files changed:
- `_invocation.py` (base class)
- `_tool_invocation.py`
- `_embedding_invocation.py`
- `_agent_invocation.py`
- `_workflow_invocation.py`
- `_inference_invocation.py`

**Out of scope** (left for follow-up, per discussion in #41):
- Dataclass fields in `types.py` holding structured/nested payloads (`ToolCallRequest.arguments`, `ServerToolCall.server_tool_call`, `FunctionToolDefinition.parameters`, etc.) — these aren't flat attribute values and need separate discussion on the right type given semantic-conventions/messages models.
- Metrics/logs usages not yet audited (per @DylanRussell's note in the issue).

All existing tests pass (240/240, no regressions).

Fixes #41 (partial — see scope note above)


eternalcuriouslearner:
Hello there!! Thanks for your contribution. Can you please take a look at the copilot suggestions around indentation and formatting?

lmolkova (approved):

DylanRussell:
I think even the data class fields should be typed to Attribute value because they end up in attributes.. this lgtm though

bhumikadangayach:
@eternalcuriouslearner Thanks for checking! I addressed the Copilot suggestions (import grouping, formatting, the get_content_attributes return type) in the follow-up commit c133de9. Happy to take another pass if you spot anything else.

eternalcuriouslearner (approved):
LGTM, Thanks again for your contribution!! @DylanRussell can we address data class fields in another pr or do you want to address that in this pr?

DylanRussell:
Separate is fine.. smaller prs usually better

eternalcuriouslearner:
Thanks for confirming @DylanRussell.

eternalcuriouslearner (dismissed):
@bhumikadangayach could you do the following:

1. can you please add a changelog in towncrirer format.
2. Fix lint, precommit and typecheck.

bhumikadangayach:
@eternalcuriouslearner Done on both:

1. Added a towncrier changelog fragment (153.changed) following the existing convention in .changelog/.
2. pre-commit run --all-files passes clean (Ruff, Ruff-format, uv-lock, and rstcheck), and pyright reports 0 errors on the package. Also cleaned up two now-unnecessary cast() calls in _retrieval_invocation.py that were waiting on the base class typing fix.

Pushed in 9fee6f7. Let me know if anything else needs a look!

eternalcuriouslearner (approved):

PR #167, opened 2026-06-24
[`opentelemetry-instrumentation-google-genai`] Update `generate_content` streaming variants to use `AsyncStreamWrapper` and `SyncStreamWrapper`  from utils

DylanRussell:
# Description

Update the `generate_content` streaming variants to use/return the util streaming wrappers.. This makes the code a bit cleaner and also ensures we let the user invoke the stream when they want to instead of us doing it immediately..

Fixes # (issue)

## Type of change

- [x ] Bug fix (non-breaking change which fixes an issue)

# How has this been tested?

Unit tests

# Checklist

- [ x] Unit tests added
- [x ] Documentation updated


instrumentation/opentelemetry-instrumentation-google-genai/src/opentelemetry/instrumentation/google_genai/generate_content.py
```diff
@@ -508,57 +588,44 @@ def _execute(
...
-                        )
-                candidates = []
-                try:
-                    for resp in wrapped(
+                return GenerateContentStreamWrapper(
+                    wrapped(
                         model=model,
                         contents=contents,
                         config=wrapped_config if has_wrapped_tools else config,
                         *_args,
```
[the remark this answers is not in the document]
DylanRussell: That is OK IMO.. we shouldn't instrument if creation fails

instrumentation/opentelemetry-instrumentation-google-genai/src/opentelemetry/instrumentation/google_genai/generate_content.py
```diff
@@ -484,6 +488,82 @@ def _execute(
...
+        invocation: InferenceInvocation,
+        telemetry_handler: TelemetryHandler,
+    ) -> None:
+        super().__init__(stream)
+        self._self_invocation = invocation
+        self._self_telemetry_handler = telemetry_handler
+        self._self_finish_reasons: list[str] = []
+        self._self_candidates = []
+
+    def _process_chunk(self, chunk: GenerateContentResponse) -> None:
```
lmolkova, line 504: not for this PR, should we mark these as public (no _) in the base class? not sure what's common approach in python for protected methods

lmolkova (approved):

instrumentation/opentelemetry-instrumentation-google-genai/src/opentelemetry/instrumentation/google_genai/generate_content.py
DylanRussell: I think `_` is the correct convention for protected

eternalcuriouslearner (approved):

PR #184, opened 2026-06-30
Instrument OpenAI Responses.retrieve and AsyncResponses.retrieve

JacksonWeber:
Adds instrumentation for the OpenAI Responses `retrieve` API (sync and async), addressing the OpenAI item in #141.

`Responses.retrieve` / `AsyncResponses.retrieve` fetch a stored response by id without going through `create()`, so the existing create wrappers don't cover them. Both are now wrapped (reusing the create wrappers, since the result is the same `Response` shape) and unwrapped on uninstrument.

Note: `retrieve` only takes a `response_id`, so the span has no `gen_ai.request.model` — all attributes come from the fetched response.

Tests: sync/async basic + content-capture cases with cassettes.

The retrieve cassettes were synthesized without provider access, so per CONTRIBUTING.md they carry a `# TODO: this is generated by AI, re-record` marker. Re-recording is tracked in #341.

Refs #141


eternalcuriouslearner (approved):

instrumentation/opentelemetry-instrumentation-genai-openai/tests/test_async_responses.py
```diff
@@ -184,6 +184,94 @@ async def test_async_responses_create_basic(
...
+        response = await async_openai_client.responses.retrieve(
+            RETRIEVE_RESPONSE_ID
+        )
+
+    (span,) = span_exporter.get_finished_spans()
+    assert_messages_attribute(
+        span.attributes[GenAIAttributes.GEN_AI_OUTPUT_MESSAGES],
+        format_simple_expected_output_message(response.output_text),
+    )
+    # retrieve has no request-side input messages to capture.
```
lmolkova: which operation name it's reported as? This should not be an inference span - no inference has happened, no tokens burned. It seems it looks like `chat`, but it does not seem right. 

We need to add conventions for this first.

JacksonWeber: Created a follow-up PR that adds the necessary work to genai semantic conventions: https://github.com/open-telemetry/semantic-conventions-genai/pull/353

eternalcuriouslearner: Should we close this until that pr is merged?

instrumentation/opentelemetry-instrumentation-genai-openai/tests/test_async_responses.py
```diff
@@ -197,6 +197,94 @@ async def test_async_responses_create_basic(
...
+    assert span.attributes[GenAIAttributes.GEN_AI_RESPONSE_ID] == response.id
+    assert (
+        span.attributes[GenAIAttributes.GEN_AI_USAGE_INPUT_TOKENS]
+        == response.usage.input_tokens
+    )
+    assert (
+        span.attributes[GenAIAttributes.GEN_AI_USAGE_OUTPUT_TOKENS]
+        == response.usage.output_tokens
+    )
+    assert span.attributes[GenAIAttributes.GEN_AI_RESPONSE_FINISH_REASONS] == (
```
lmolkova: should we check status too?

instrumentation/opentelemetry-instrumentation-genai-openai/tests/test_async_responses.py
```diff
@@ -197,6 +197,94 @@ async def test_async_responses_create_basic(
...
+        "test_async_responses_retrieve_basic[content_mode0].yaml"
+    ):
+        response = await async_openai_client.responses.retrieve(
+            RETRIEVE_RESPONSE_ID
+        )
+
+    (span,) = span_exporter.get_finished_spans()
+    # retrieve carries no request model/input; telemetry comes from the
+    # fetched response only.
+    assert GenAIAttributes.GEN_AI_REQUEST_MODEL not in span.attributes
```
lmolkova: let's check operation name and span name

eternalcuriouslearner (approved):

instrumentation/opentelemetry-instrumentation-genai-openai/src/opentelemetry/instrumentation/genai/openai/__init__.py
```diff
@@ -188,6 +188,22 @@ def _instrument(self, **kwargs):
...
             )
 
+            # retrieve() fetches a stored response by id without going through
+            # create(), so wrap it separately. The result is a Response, so the
+            # create wrappers extract its telemetry correctly.
+            if hasattr(responses_module.Responses, "retrieve"):
+                wrap_function_wrapper(
+                    "openai.resources.responses.responses",
+                    "Responses.retrieve",
+                    responses_create(handler),
```
lmolkova: this would produce inference `chat` span instead of fetch_response

instrumentation/opentelemetry-instrumentation-genai-openai/tests/test_async_responses.py
JacksonWeber: Helper now asserts `gen_ai.response.status` on each fetch test.

instrumentation/opentelemetry-instrumentation-genai-openai/tests/test_async_responses.py
JacksonWeber: Added a shared helper, `assert_fetch_response_attributes()` to check these.

instrumentation/opentelemetry-instrumentation-genai-openai/src/opentelemetry/instrumentation/genai/openai/__init__.py
JacksonWeber: Added a dedicated `responses_retrieve` and `async_responses_retrieve` wrapper and used those instead. This uses the new `handler.fetch_response()` and now produces `gen_ai.olperation.name=fetch_response`.

instrumentation/opentelemetry-instrumentation-genai-openai/tests/test_async_responses.py
JacksonWeber: Above merged.

instrumentation/opentelemetry-instrumentation-genai-openai/src/opentelemetry/instrumentation/genai/openai/patch_responses.py
```diff
@@ -45,6 +55,39 @@
...
+
+    ``with_streaming_response`` hands back a raw response whose body has *not*
+    been read yet; parsing it would consume the body before the caller asks for
+    it, so those calls are left alone.
+    """
+    if _is_streamed_raw_response(kwargs) or not isinstance(
+        result, ParsableResponse
+    ):
+        return result
+    return result.parse()
```
lmolkova: we should have it already for responses - https://github.com/open-telemetry/opentelemetry-python-genai/blob/9d925935ad1a9056cfcb9501de7326da78184c28/instrumentation/opentelemetry-instrumentation-genai-openai/src/opentelemetry/instrumentation/genai/openai/response_extractors.py#L458, let's reuse it 

it seems _is_streamed_raw_response would be useful there too btw.

instrumentation/opentelemetry-instrumentation-genai-openai/src/opentelemetry/instrumentation/genai/openai/patch_responses.py
```diff
@@ -191,6 +234,176 @@ async def traced_method(
...
+                return wrap_stream_result(
+                    AsyncFetchResponseStreamWrapper,
+                    result,
+                    invocation,
+                    capture_content,
+                )
+
+            parsed = _parse_raw_response(result, kwargs)
+            if inspect.isawaitable(parsed):
+                parsed = await parsed
```
lmolkova: this is probably also necessary in https://github.com/open-telemetry/opentelemetry-python-genai/blob/9d925935ad1a9056cfcb9501de7326da78184c28/instrumentation/opentelemetry-instrumentation-genai-openai/src/opentelemetry/instrumentation/genai/openai/response_extractors.py#L466

instrumentation/opentelemetry-instrumentation-genai-openai/src/opentelemetry/instrumentation/genai/openai/response_extractors.py
```diff
@@ -265,6 +271,91 @@ def _finish_reason_from_status(status: str | None) -> str | None:
...
 
+# `incomplete_details.reason` values that map onto a cross-provider finish
+# reason; an unrecognized reason is reported as-is.
+_INCOMPLETE_REASON_TO_FINISH_REASON = {
+    "max_output_tokens": "length",
+    "content_filter": "content_filter",
+}
+
+
+def get_fetched_finish_reasons(response: "Response | None") -> list[str]:
```
lmolkova: can we align fetch and responses create finish reasons - there is a related helper above - _finish_reason_from_status and responses.create also returns status. we should use the same helpers for both

instrumentation/opentelemetry-instrumentation-genai-openai/src/opentelemetry/instrumentation/genai/openai/patch_responses.py
```diff
@@ -191,6 +234,176 @@ async def traced_method(
...
+    handler: TelemetryHandler,
+    instance: "Responses | AsyncResponses",
+    response_id: str,
+    kwargs: dict[str, Any],
+) -> FetchResponseInvocation:
+    invocation = handler.fetch_response(
+        **get_fetch_response_creation_kwargs(response_id, instance)
+    )
+    invocation.stream_cursor = _get_stream_cursor(kwargs)
+    invocation.attributes[OpenAIAttributes.OPENAI_API_TYPE] = (
```
lmolkova, line 225: nice! we should set it on responses.create too

util/opentelemetry-util-genai/src/opentelemetry/util/genai/_fetch_response_invocation.py
```diff
@@ -0,0 +1,181 @@
...
+            GenAI.GEN_AI_OPERATION_NAME: self._operation_name,
+            GenAI.GEN_AI_PROVIDER_NAME: self._provider,
+            **{k: v for k, v in optional_attrs if v is not None},
+        }
+        attrs.update(self.metric_attributes)
+        return attrs
+
+    def _get_attributes(self) -> dict[str, AttributeValue]:
+        optional_attrs: tuple[tuple[str, AttributeValue | None], ...] = (
+            (_GEN_AI_REQUEST_STREAM_CURSOR, self.stream_cursor),
```
lmolkova, line 158: we should also set gen_ai.request.stream when there was streaming

instrumentation/opentelemetry-instrumentation-genai-openai/src/opentelemetry/instrumentation/genai/openai/__init__.py
```diff
@@ -188,6 +190,22 @@ def _instrument(self, **kwargs):
                 async_responses_stream(handler),
             )
 
+            # retrieve() fetches a stored response by id. No inference happens
+            # and no tokens are consumed, so it is traced as a fetch_response
+            # operation rather than through the create wrappers.
+            if hasattr(responses_module.Responses, "retrieve"):
```
lmolkova: is `hasattr(responses_module.Responses, "retrieve"):` necessary?

instrumentation/opentelemetry-instrumentation-genai-openai/src/opentelemetry/instrumentation/genai/openai/patch_responses.py
JacksonWeber: Moved response detection/parsing to `response_extractors.py` and now reuse it for inference and fetch-response extraction.

instrumentation/opentelemetry-instrumentation-genai-openai/src/opentelemetry/instrumentation/genai/openai/patch_responses.py
JacksonWeber: Added in `response_extractors.py`

instrumentation/opentelemetry-instrumentation-genai-openai/src/opentelemetry/instrumentation/genai/openai/response_extractors.py
JacksonWeber: Create and fetch now use the same logic.

instrumentation/opentelemetry-instrumentation-genai-openai/src/opentelemetry/instrumentation/genai/openai/patch_responses.py
JacksonWeber: I've set this value on all Responses now

util/opentelemetry-util-genai/src/opentelemetry/util/genai/_fetch_response_invocation.py
JacksonWeber: The `TelemetryHandler.fetch_resopnse()` now accepts if the request is streamed and records it at span creation.

instrumentation/opentelemetry-instrumentation-genai-openai/src/opentelemetry/instrumentation/genai/openai/__init__.py
JacksonWeber: I've confirmed the `hasattr` check here is unnecessary since `_get_responses_module()` already gates the block.

lmolkova (approved):

PR #202, opened 2026-07-08
Drop the semconv / OTEL_SEMCONV_STABILITY_OPT_IN plumbing from the shared test helper

### LINKED ISSUE #200: Remove experimental flag in tests
unrelated to this PR, but we should get rid of the need for the SemConv param here.. "gen_ai_latest_experimental" should be the default

_Originally posted by @DylanRussell in https://github.com/open-telemetry/opentelemetry-python-genai/pull/131#discussion_r3537441641_

AgentGymLeader:
Closes #200.

`OTEL_SEMCONV_STABILITY_OPT_IN` is no longer read anywhere — the code behind it was removed from `opentelemetry-util-genai` and `opentelemetry-instrumentation-google-genai`, and the experimental conventions are now the default and only behavior. So the shared test helper's `semconv` parameter (which only forwarded to that env var) is now a no-op.

This drops the `semconv` parameter and its `OTEL_SEMCONV_STABILITY_OPT_IN` mapping from `opentelemetry.test_util_genai.instrumentor.instrument()`, and removes the now-dead `semconv=` arguments at every call site, including the `anthropic` package's `stable`-mode fixture.

The `openai` `content_mode` fixture keeps its tuple shape on purpose: its parametrize id is what the VCR cassette filenames key off, so reshaping it would orphan the recorded cassettes. Its now-unused first element (and the leftover `OTEL_SEMCONV_STABILITY_OPT_IN` setenvs in the langchain/openai tests) are dead too and could be cleaned up in a follow-up.

All test and lint CI jobs for the touched packages (anthropic, openai, langchain, openai-agents) pass. The only failing check is the changelog gate — this is a test-only change to a `Private :: Do Not Upload` package, so it needs either a fragment or the `Skip Changelog` label; happy to add a fragment if you'd prefer.

### DIFF
--- a/util/opentelemetry-test-util-genai/src/opentelemetry/test_util_genai/instrumentor.py
+++ b/util/opentelemetry-test-util-genai/src/opentelemetry/test_util_genai/instrumentor.py
@@ -33,7 +33,6 @@ from contextlib import contextmanager
 from typing import Any
 
 from opentelemetry.instrumentation._semconv import (
-    OTEL_SEMCONV_STABILITY_OPT_IN,
     _OpenTelemetrySemanticConventionStability,
 )
 from opentelemetry.instrumentation.instrumentor import BaseInstrumentor
@@ -50,7 +49,6 @@ def instrument(
     tracer_provider: Any,
     logger_provider: Any,
     meter_provider: Any,
-    semconv: str | None = None,
     content_capture: str | None = None,
     emit_event: bool = False,
     extra_env: Mapping[str, str] | None = None,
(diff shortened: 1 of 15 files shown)


util/opentelemetry-test-util-genai/src/opentelemetry/test_util_genai/instrumentor.py
```diff
@@ -68,17 +68,18 @@ def instrument_with_content(
...
                 logger_provider=logger_provider,
                 meter_provider=meter_provider,
-                semconv="gen_ai_latest_experimental",
                 content_capture="SPAN_ONLY",
             ) as instrumentor:
                 yield instrumentor
 
-    ``semconv`` is forwarded to ``OTEL_SEMCONV_STABILITY_OPT_IN``;
-    ``content_capture`` to ``OTEL_INSTRUMENTATION_GENAI_CAPTURE_MESSAGE_CONTENT``;
+    ``semconv`` is forwarded to ``OTEL_SEMCONV_STABILITY_OPT_IN`` and
```
lmolkova: can we just drop it? We don't need to set OTEL_SEMCONV_STABILITY_OPT_IN anymore

AgentGymLeader: You're right — with 117.removed / 110.removed the code reading the opt-in is already gone, so rather than changing the default I'll drop the `semconv` plumbing (the `OTEL_SEMCONV_STABILITY_OPT_IN` mapping) from the helper entirely.

The two remaining explicit `semconv` call sites — anthropic's `stable` fixture and openai's content-mode parametrization — should be no-ops now that nothing reads the flag, so I'd clean those up in the same pass.

AgentGymLeader: Confirmed — `content_capture` alone is enough, and nothing reads `OTEL_SEMCONV_STABILITY_OPT_IN` anymore (its code was removed from util-genai and google-genai; no instrumentation src reads it). So I dropped the `semconv` / `OTEL_SEMCONV_STABILITY_OPT_IN` plumbing from the helper and the now-dead `semconv=` call sites, including anthropic's `stable` fixture, in the latest commit.

One deliberate non-change: I left the openai `content_mode` fixture's tuple shape intact — its parametrize id is what the VCR cassette filenames key off, so reshaping it would orphan ~40 cassettes. Its now-unused flag (and the leftover `OTEL_SEMCONV_STABILITY_OPT_IN` setenvs in the langchain/openai tests) are dead too; happy to clean those up in a follow-up if you'd like.

Resolving this thread.

DylanRussell (approved):
I think there are a few more instances of this value that you can clean up in this PR -- check out https://github.com/search?q=repo%3Aopen-telemetry%2Fopentelemetry-python-genai%20gen_ai_latest_experimental&type=code --

https://github.com/open-telemetry/opentelemetry-python-genai/blob/056dd80888d38a8e361b98cb941bef6995d272f1/instrumentation/opentelemetry-instrumentation-genai-openai/README.rst#L126 

and 

https://github.com/open-telemetry/opentelemetry-python-genai/blob/056dd80888d38a8e361b98cb941bef6995d272f1/instrumentation/opentelemetry-instrumentation-genai-openai/examples/manual/.env#L21


I didn't look through all of them..

lmolkova (approved):

PR #232, opened 2026-07-09
[`opentelemetry-instrumentation-google-genai`] Add weaver tests. Add missing server address attribute

DylanRussell:
# Description

Adds weaver tests for `opentelemetry-instrumentation-google-genai`. Also add the missing server address attribute to the telemetry we generate.

## Type of change

- [ x] New feature (non-breaking change which adds functionality)

# How has this been tested?

These are tests being added

# Checklist

- [x] Unit tests added
- [x] Documentation updated


eternalcuriouslearner (approved):

instrumentation/opentelemetry-instrumentation-google-genai/tests/conformance/tool_calling.py
```diff
@@ -0,0 +1,65 @@
...
+
+
+class ToolCallingScenario(Scenario):
+    expected_spans = ("generate_content", "execute_tool")
+    expected_metrics = (
+        "gen_ai.client.operation.duration",
+        "gen_ai.client.token.usage",
+    )
+    expected_violations = (
+        ExpectedViolation(
```
lmolkova, line 32: not sure what to do about it (we probably don't want this as a violation reported by weaver, but also in many cases it's good to have it flagged), will keep thinking how we can improve it

lmolkova (approved):
Nice! it seems it caught something :)

instrumentation/opentelemetry-instrumentation-google-genai/tests/conformance/tool_calling.py
DylanRussell: Ack, I think for this PR we can leave it as is ?

DylanRussell:
Nice so I fixed one missing attribute it correctly reported missing. CI is passing now, but is failing for openai stuff, I suspect that is broken at head because I didn't touch that..

PR #233, opened 2026-07-09
[`opentelemetry-instrumentation-google-genai`] Add response ID attribute to the interactions API telemetry (inference span and event)

DylanRussell:
# Description

Set response id attribute to interactions.id

## Type of change

# How has this been tested?

Unit tests

# Checklist

- [x] Unit tests added
- [x] Documentation updated


instrumentation/opentelemetry-instrumentation-google-genai/src/opentelemetry/instrumentation/google_genai/interactions.py
```diff
@@ -301,6 +301,12 @@ def instrumented_interactions_create(
             server_address=server_address,
         )
 
+        previous_interaction_id = kwargs.get(
+            "previous_interaction_id"
+        ) or _get_field(kwargs.get("config"), "previous_interaction_id")
+        if previous_interaction_id:
+            invocation.conversation_id = str(previous_interaction_id)
```
lmolkova: is this a conversation id though?

Assuming there are multiple turns within one conversation, the conversation.id stays the same, but I imagine interaction id changes with each turn?

DylanRussell: hmm good point, I thought it stayed the same but it changes on each turn. 

So I could put `Interactions.id` as the response ID sem conv attribute (and which is what the user uses in a subsequent request as the `previous_interactions_id`, but I don't think there's a good place to put `previous_interactions_id` which comes through via the request kwargs

lmolkova: > I could put Interactions.id as the response ID 

👍 

>  I don't think there's a good place to put previous_interactions_id which comes through via the request kwargs

Would you mind sending PR to semconv to propose it (or file an issue)? OpenAI responses has the same problem and it makes sense to introduce a common attribute

lmolkova (approved):

instrumentation/opentelemetry-instrumentation-google-genai/src/opentelemetry/instrumentation/google_genai/interactions.py
```diff
@@ -82,6 +82,8 @@ def _apply_interaction_response_attributes(
     telemetry_handler: TelemetryHandler,
 ) -> None:
     invocation.response_model_name = response.model
+    if getattr(response, "id", None):
+        invocation.response_id = response.id
```
eternalcuriouslearner, line 135: question(non-blocking): do we need a check in vcr or conformance testing for this attribute?

eternalcuriouslearner (approved):

instrumentation/opentelemetry-instrumentation-google-genai/src/opentelemetry/instrumentation/google_genai/interactions.py
DylanRussell: Updated PR to use response id, and sent out https://github.com/open-telemetry/semantic-conventions-genai/pull/372 for `previous_reponse_id` sem conv

instrumentation/opentelemetry-instrumentation-google-genai/src/opentelemetry/instrumentation/google_genai/interactions.py
DylanRussell: Added it in

eternalcuriouslearner: Awesome thank you sir.

PR #234, opened 2026-07-09
ci: restructure release workflows to unblock PyPI trusted publishing

MikeGoldsmith:
Releases fail at `pypa/gh-action-pypi-publish` because it runs from a reusable workflow, which PyPI trusted publishing doesn't support ([warehouse#11096](https://github.com/pypi/warehouse/issues/11096)).

This PR moves the publish step into a job defined directly in each top-level release workflow. The rest of the release logic stays DRY via two reusable workflows (`_release-build.yml`, `_release-finalize.yml`). Prepare workflows are renamed and consolidated at the same time.

**Follow-up (not in this PR)**: register two new trusted publishers per PyPI project (`release-package.yml` and `release-all.yml`, environment `pypi`). See the updated `RELEASING.md` table.


.github/workflows/release-all.yml
```diff
@@ -67,21 +67,52 @@ jobs:
...
+      - uses: actions/download-artifact@v4
+        with:
+          name: dist-${{ matrix.package }}
+          path: dist/
+
+      - name: Publish to PyPI
+        uses: pypa/gh-action-pypi-publish@release/v1
+        with:
+          packages-dir: dist/
+          skip-existing: true
```
aabmass, line 102: I believe you can run this action once to upload many build wheel/sdists in the `packages-dir`. I guess you could theoretically gather all the artifacts together and run once instead of in a matrix.

Maybe this was has some benefits though.

.github/workflows/prepare-package-patch.yml
```diff
@@ -0,0 +1,36 @@
...
+        type: choice
+        options:
+        - opentelemetry-instrumentation-genai-anthropic
+        - opentelemetry-instrumentation-google-genai
+        - opentelemetry-instrumentation-genai-langchain
+        - opentelemetry-instrumentation-genai-openai-agents
+        - opentelemetry-instrumentation-genai-openai
+        - opentelemetry-util-genai
+        description: 'Package to prepare for patch release'
+        required: true
```
aabmass, line 15: Is there a way to do a patch release for the whole repo?

aabmass (approved):

.github/workflows/prepare-package-patch.yml
MikeGoldsmith: No, I didn't add a patch-all workflow. I assumed patches were more likely to be critical bug rather than bump everything. 

We can look to add more variants, I think a release major might be useful in the future too.

.github/workflows/release-all.yml
MikeGoldsmith: Nice, we can look at doing that optimisation later 👍🏻

MikeGoldsmith: Open this issue to track:
- #235

.github/workflows/prepare-package-patch.yml
MikeGoldsmith: Open tracking issue:
- https://github.com/open-telemetry/opentelemetry-python-genai/issues/236

.github/workflows/_release-build.yml
```diff
@@ -0,0 +1,119 @@
...
+        required: false
+        type: string
+        default: ""
+      on_main:
+        required: false
+        type: boolean
+        default: false
+
+permissions:
+  contents: read
```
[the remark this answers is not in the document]
MikeGoldsmith: actions/upload-artifact@v4 doesn't use GITHUB_TOKEN scopes for same-workflow artifacts. It uses the runner's ACTIONS_RUNTIME_TOKEN, which the runner provisions independently of the permissions block. No actions: write needed.

.github/workflows/release-package.yml
```diff
@@ -0,0 +1,52 @@
...
+  contents: read
+
+run-name: "Release: ${{ inputs.package }}"
+
+jobs:
+  build:
+    uses: ./.github/workflows/_release-build.yml
+    with:
+      package: ${{ inputs.package }}
+
```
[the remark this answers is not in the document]
MikeGoldsmith: Job-level permissions on a uses: call to a reusable workflow are ignored, the reusable workflow declares its own permissions. That's why I dropped the equivalent block on the matrix build in release-all.yml in a follow-up commit.

.github/workflows/release-package.yml
```diff
@@ -0,0 +1,52 @@
...
+  publish:
+    needs: build
+    runs-on: ubuntu-latest
+    environment: pypi
+    permissions:
+      id-token: write
```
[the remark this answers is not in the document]
MikeGoldsmith: actions: read is only required for downloading artifacts from a different workflow run. Same-run downloads (build and publish are in the same run here) work with just id-token: write.

.github/workflows/release-package.yml
```diff
@@ -0,0 +1,52 @@
...
+      - uses: actions/download-artifact@v4
+        with:
+          name: dist-${{ inputs.package }}
+          path: dist/
+
+      - name: Publish to PyPI
+        uses: pypa/gh-action-pypi-publish@release/v1
+        with:
+          packages-dir: dist/
+          skip-existing: true
```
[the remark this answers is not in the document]
MikeGoldsmith: @release/v1 is what PyPA officially recommends for this action. It's a moving tag they update to push security fixes to all consumers. The workflow this PR replaces used the same tag.

.github/workflows/release-all.yml
```diff
@@ -67,21 +67,52 @@ jobs:
...
+    strategy:
+      fail-fast: false
+      matrix:
+        package: ${{ fromJson(needs.enumerate.outputs.packages) }}
+    uses: ./.github/workflows/_release-build.yml
+    with:
+      package: ${{ matrix.package }}
+      ref: ${{ needs.enumerate.outputs.ref }}
+      on_main: true
+
```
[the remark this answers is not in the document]
MikeGoldsmith: actions/upload-artifact@v4 doesn't use GITHUB_TOKEN scopes for same-workflow artifacts. It uses the runner's ACTIONS_RUNTIME_TOKEN, which the runner provisions independently of the permissions block. No actions: write needed.

.github/workflows/release-all.yml
```diff
@@ -67,21 +67,52 @@ jobs:
...
+        package: ${{ fromJson(needs.enumerate.outputs.packages) }}
+    environment: pypi
     permissions:
-      contents: write
       id-token: write
+    steps:
+      - uses: actions/download-artifact@v4
+        with:
+          name: dist-${{ matrix.package }}
+          path: dist/
```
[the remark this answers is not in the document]
MikeGoldsmith: actions: read is only required for downloading artifacts from a different workflow run. Same-run downloads (build and publish are in the same run here) work with just id-token: write.

.github/workflows/release-all.yml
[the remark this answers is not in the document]
MikeGoldsmith: @release/v1 is what PyPA officially recommends for this action. It's a moving tag they update to push security fixes to all consumers. The workflow this PR replaces used the same tag.

RELEASING.md
```diff
@@ -112,42 +119,37 @@ When a new package is ready to ship:
...
-| Workflow name | `release-all.yml` |
-| Environment name | `pypi` |
-
-Register a publisher entry for each release workflow (`release-all.yml` and
-`package-release.yml`).
-
-4. Optionally reserve the package name to prevent name-squatting shortly after
-   the introductory PR lands on `main` by navigating to
-   <https://pypi.org/manage/organization/opentelemetry/projects/>, scrolling to the
-   bottom (**Add project to organization**), and using the form.
```
aabmass, line 135: Did you mean to change this? I sent the update yesterday in #231

MikeGoldsmith: Oops - no, I didn't meant to remove that. I've put it back in e9c34cf 👍🏻

lmolkova (approved):

PR #261, opened 2026-07-13
Fix attribute error for raw tuples in langchain

rads-1996:
# Description

`to_input_messages` assumed every entry in an input messages list was a LangChain BaseMessage. LangGraph/create_agent inputs commonly use raw ("role", text) tuples (e.g. {"messages": [("human", user_input)]}). Accessing .content on such a tuple raised AttributeError("'tuple' object has no attribute 'content'"), which aborted the callback before the invocation could be registered.

image

This exception fired inside the root agent's on_chain_start, before the agent invocation was registered in the invocation manager. As a result, find_nearest_agent could never locate the root agent, so each LangGraph super-step node (model, tools, ...) created its own invoke_agent span, producing duplicate invoke_agent spans instead of one agent span with nested tool/LLM children.

Screenshot 2026-07-13 094210

**After Change -** 

image

## Type of change

# How has this been tested?

configuration.

- [x]  uv run tox -e py312-test-instrumentation-genai-langchain -- -q
- [x]  uv run tox -e py312-test-instrumentation-genai-langchain-conformance -- -q
- [x]  uv run --python 3.12 tox -e lint-instrumentation-genai-langchain

# Checklist

- [x] Unit tests added
- [ ] Documentation updated

### DIFF
--- a/instrumentation/opentelemetry-instrumentation-genai-langchain/.changelog/261.changed
+++ b/instrumentation/opentelemetry-instrumentation-genai-langchain/.changelog/261.changed
@@ -1,0 +1,1 @@
+Support raw ("role", content) tuple messages in agent inputs, preventing duplicate `invoke_agent` spans.
--- a/instrumentation/opentelemetry-instrumentation-genai-langchain/.changelog/261.changed
+++ b/instrumentation/opentelemetry-instrumentation-genai-langchain/.changelog/261.changed
@@ -1,1 +1,1 @@
-Support raw ("role", content) tuple messages in agent inputs, preventing duplicate `invoke_agent` spans.
+Support raw LangChain message inputs (`("role", content)` tuples, dicts, and strings) in input messages, so the prompt is recorded and duplicate `invoke_agent` spans are avoided.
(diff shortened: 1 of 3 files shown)


instrumentation/opentelemetry-instrumentation-genai-langchain/src/opentelemetry/instrumentation/genai/langchain/utils.py
```diff
@@ -157,6 +157,8 @@ def to_input_messages(
     """Convert LangChain messages into spec-conformant ``InputMessage`` s."""
     result: list[InputMessage] = []
     for message in messages:
+        if not isinstance(message, BaseMessage):
```
lmolkova: could you please create a follow up issue to record something when we get non-standard messages here? It'd be good to have some information, especially since langchain allow it.

instrumentation/opentelemetry-instrumentation-genai-langchain/.changelog/261.changed
```diff
@@ -0,0 +1 @@
+Support raw ("role", content) tuple messages in agent inputs, preventing duplicate `invoke_agent` spans.
```
lmolkova: is this accurate? it seems we don't support them, we just ignore them for the time being

rads-1996: Yes, I will have to rephrase it. I was thinking instead of skipping, we should convert any non-standard messages into BaseMessage format so we don't loose the prompt and important attributes. There is a method in langchain_core named [`convert_to_messages`](https://github.com/langchain-ai/langchain/blob/00c0f69dc0dfd619f023f1ee9dc83574d8a173ed/libs/core/langchain_core/messages/utils.py#L706) that normalizes formats that langchain accepts to BaseMessage.

lmolkova (approved):
TY!

eternalcuriouslearner (approved):

PR #269, opened 2026-07-15
Add streaming metrics and attribute

lmolkova:
Replace #13 

- Report time-to-first-chunk and time-between chunks in genai utils for inference
- Onboard anthropic and openai
- Add gen_ai.request.stream attribute


instrumentation/opentelemetry-instrumentation-genai-anthropic/tests/requirements.oldest.txt
```diff
@@ -21,5 +21,5 @@
 # OpenTelemetry SDK and test utilities come transitively from opentelemetry-test-util-genai, which
 # every oldest env installs. Pin here only test-only deps that nothing else already provides.
 #
-# There is currently nothing to pin: anthropic has no test-only dependency that isn't already
-# provided by a declared bound or by the shared test fixtures.
+# Drop this once opentelemetry-util-genai 1.1b0 is published.
+-e util/opentelemetry-util-genai
```
lmolkova, line 25: todo: let's automate it in post-release workflow

util/opentelemetry-util-genai/src/opentelemetry/util/genai/stream.py
```diff
@@ -116,7 +127,12 @@ def __next__(self) -> ChunkT:
         except Exception as error:
             self._finalize_failure(error)
             raise
+        invocation = self._self_invocation
+        chunk_at = timeit.default_timer() if invocation is not None else None
         self._process_chunk(chunk)
+        # Record after _process_chunk so response.model is on the metrics.
+        if invocation is not None and chunk_at is not None:
+            invocation._on_stream_chunk(chunk_at)
```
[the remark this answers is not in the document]
lmolkova: it's not measuring next() duration, it's capturing delta as chunks arrive from server. measures service perf, not client.

Nik-Reddy:
Thanks for picking this up @lmolkova! I'm glad to see the streaming timing work from my original PR #13 moving forward. Happy to review this since I worked on the initial implementation.

Also happy to co-author this if it helps, since a lot of the design decisions here came out of the iteration on #13. Let me know!

util/opentelemetry-util-genai/src/opentelemetry/util/genai/stream.py
Nik-Reddy, line 135: Good call grabbing `chunk_at` before `_process_chunk` so the arrival time is clean, then deferring `_on_stream_chunk` until after so `response.model` lands on the metric attributes. 

One thing I noticed though: if `_process_chunk` raises, the timing for that chunk just gets silently skipped. I ran into the same question in #13 and ended up leaving it as-is since a failed chunk probably shouldn't record timing anyway, but maybe worth a short comment here explaining that?

util/opentelemetry-util-genai/src/opentelemetry/util/genai/_invocation.py
```diff
@@ -110,6 +117,38 @@ def _get_metric_token_counts(self) -> dict[str, int]:  # pylint: disable=no-self
...
+            self._ttfc_seconds = delta
+            self._metrics_recorder.record_time_to_first_chunk(
+                delta,
+                attributes=attributes,
+                context=self._span_context,
+            )
+        else:
+            self._metrics_recorder.record_time_per_chunk(
+                delta,
+                attributes=attributes,
```
Nik-Reddy, line 150: If `_monotonic_start_s` is None here the method just returns silently and no TTFC gets recorded. That's probably the right behavior, but someone debugging missing streaming metrics would have a hard time tracing it back to this guard. Would a logger.debug line here make sense so it's easier to spot?

instrumentation/opentelemetry-instrumentation-genai-openai/src/opentelemetry/instrumentation/genai/openai/chat_wrappers.py
```diff
@@ -33,17 +33,18 @@ class _ChatStreamMixin:
     _self_capture_content: bool
     _self_choice_buffers: list[ChoiceBuffer]
     _self_response_id: Optional[str]
-    _self_response_model: Optional[str]
     _self_service_tier: Optional[str]
     _self_prompt_tokens: Optional[int]
     _self_completion_tokens: Optional[int]
 
     def _set_response_model(self, chunk: ChatCompletionChunk) -> None:
```
Nik-Reddy, line 43: Makes sense to set `response_model_name` eagerly here instead of waiting for `_cleanup`. Without this, all the per-chunk streaming metrics would be missing `gen_ai.response.model` until the stream finishes, which kind of defeats the purpose of having per-chunk timing in the first place. Clean refactor.

util/opentelemetry-util-genai/src/opentelemetry/util/genai/_inference_invocation.py
```diff
@@ -92,8 +92,19 @@ def __init__(
...
         self._start(self._get_base_attributes())
 
+    @property
+    def response_model_name(self) -> str | None:
+        return self._response_model_name
+
+    @response_model_name.setter
+    def response_model_name(self, value: str | None) -> None:
+        if value != self._response_model_name:
+            self._response_model_name = value
```
lmolkova, line 109: can you please not leave AI generated reviews? We have copilot for this.

instrumentation/opentelemetry-instrumentation-genai-openai/src/opentelemetry/instrumentation/genai/openai/chat_wrappers.py
lmolkova: this seems to be an AI-generated comment. I appreciate support and reviews from humans

util/opentelemetry-util-genai/src/opentelemetry/util/genai/stream.py
lmolkova: the metric is defined as time to chunk with no error.type on in. I don't think it needs a comment.

util/opentelemetry-util-genai/src/opentelemetry/util/genai/_invocation.py
lmolkova: It should never be none, updated type to be float and removed defensive early return

eternalcuriouslearner:
@lmolkova I am looking into this. Running few scenarios. If you get approvals, please feel free to merge it.

util/opentelemetry-util-genai/.changelog/269.added
```diff
@@ -0,0 +1 @@
+Add streaming timing metrics (``gen_ai.client.operation.time_to_first_chunk``, ``gen_ai.client.operation.time_per_output_chunk``), the ``gen_ai.response.time_to_first_chunk`` span attribute, and the ``gen_ai.request.stream`` span attribute, set by the shared stream wrappers.
```
aabmass: This and other changelogs I thought were markdown but have <code>``</code> double ticks. Is it supposed to be RST?

util/opentelemetry-util-genai/src/opentelemetry/util/genai/_inference_invocation.py
```diff
@@ -161,16 +173,29 @@ def _get_attributes(self) -> dict[str, AttributeValue]:
...
         return attrs
 
     def _get_metric_attributes(self) -> dict[str, AttributeValue]:
-        attrs = self._get_start_attributes()
-        if self.response_model_name is not None:
-            attrs[GenAI.GEN_AI_RESPONSE_MODEL] = self.response_model_name
-        attrs.update(self.metric_attributes)
-        return attrs
+        # Cached because this is called once per streaming chunk; invalidated
+        # by the ``response_model_name`` setter and ``_apply_error_attributes``.
```
aabmass: I couldn't really understand the sequence of these calls that update state, it feels a little fragile because of you must call the seemingly unrelated setters IIUC.

util/opentelemetry-util-genai/src/opentelemetry/util/genai/stream.py
aabmass, line 135: Any reason the `_on_stream_chunk()` can't call `default_timer()` internally instead of passing it? Looks like there would be negligible time difference but maybe it's just for testing purposes

util/opentelemetry-util-genai/src/opentelemetry/util/genai/instruments.py
```diff
@@ -55,3 +55,21 @@ def create_token_histogram(meter: Meter) -> Histogram:
...
+        description="Time to receive the first chunk, measured from when the client issues the generation request to when the first chunk is received in the response stream.",
+        unit="s",
+        explicit_bucket_boundaries_advisory=_GEN_AI_CLIENT_OPERATION_DURATION_BUCKETS,
+    )
+
+
+def create_time_per_output_chunk_histogram(meter: Meter) -> Histogram:
+    return meter.create_histogram(
+        name=gen_ai_metrics.GEN_AI_CLIENT_OPERATION_TIME_PER_OUTPUT_CHUNK,
+        description="Time per output chunk, recorded for each chunk received after the first one, measured as the time elapsed from the end of the previous chunk to the end of the current chunk.",
```
aabmass, line 72: I can sort of guess why, but can you remind me why we treat the first chunk specially?

instrumentation/opentelemetry-instrumentation-genai-anthropic/tests/test_metrics.py
```diff
@@ -0,0 +1,90 @@
...
+                pass
+
+    _assert_streaming_timing_metrics(metric_reader)
+
+
+@pytest.mark.asyncio()
+async def test_async_messages_streaming_timing_metrics(
+    metric_reader, async_anthropic_client, instrument_with_content, vcr
+):
+    with vcr.use_cassette("test_async_messages_create_streaming.yaml"):
```
aabmass, line 80: I don't see any cassettes recordings checked-in to this PR, is that expected?

aabmass (approved):

util/opentelemetry-util-genai/src/opentelemetry/util/genai/stream.py
lmolkova: time between chunks would be negligible on it's own (less than millisecond, easy) so it seems reasonable to care about higher precision here

util/opentelemetry-util-genai/src/opentelemetry/util/genai/_inference_invocation.py
lmolkova: anyone who changes this method would naturally care about what invalidates cache. Setting any settable metric attribute would (which is response model and error). I refactored it a bit to hopefully make it more clear. 

The reason I still want cache is that there could be hundreds on chunks per stream if not thousands and allocating the same dict of attribute on each of them is wasteful.

util/opentelemetry-util-genai/src/opentelemetry/util/genai/instruments.py
lmolkova: same as time-to-first-byte and time-to-last-byte - the former tells when server started replying and it's one type of SLI, the latter is how fast it finished and it's a different metric that has maybe more to do with network quality than server. These two friends is what vLLM and the inference serving industry care about independently

instrumentation/opentelemetry-instrumentation-genai-anthropic/tests/test_metrics.py
lmolkova: we're reusing existing ones - any successful streaming recording works (sync/async  doesn't matter since it's API-level concern, recordings are the same)

PR #271, opened 2026-07-15
[`opentelemetry-instrumentation-google-genai`] Add tool definitions attribute on interactions API, other minor changes

DylanRussell:
# Description

Added tool definitions attribute on the interactions API spans.

Function calling for interactions API is described at: https://ai.google.dev/gemini-api/docs/function-calling

Note that automatic function calling is not yet supported, so we can't get the execute tool spans like we do with `generate_content` instrumentation. The user is on the hook for calling their own functions..

I used `GenericToolDefinition` to capture built in tools and MCP server being provided as tools to the API.. 

I also updated the code to use the invoke remote agent span when an `agent` is specified instead of an `inference` span since a remote agent is being called.

I deleted obsolete TODOs and updated the README.

## Type of change

# How has this been tested?

Unit tests

# Checklist

- [x] Unit tests added
- [x] Documentation updated


instrumentation/opentelemetry-instrumentation-google-genai/src/opentelemetry/instrumentation/google_genai/interactions.py
```diff
@@ -381,15 +456,29 @@ async def instrumented_interactions_create(
...
+            )
+        else:
+            invocation = telemetry_handler.inference(
+                provider=provider,
+                request_model=kwargs.get("model"),
+                operation_name="interactions.create",
+                server_address=server_address,
+            )
+        invocation.tool_definitions = _maybe_get_tool_definitions(
+            kwargs.get("tools")
```
lmolkova: nit: looks very similar to line 387-410, maybe add a helper?

lmolkova (approved):

eternalcuriouslearner (approved):

instrumentation/opentelemetry-instrumentation-google-genai/src/opentelemetry/instrumentation/google_genai/interactions.py
DylanRussell: True nice. Done

PR #275, opened 2026-07-16
Set `gen_ai.workflow.name` span attribute on workflow invocations

lmolkova:
WorkflowInvocation now sets `gen_ai.workflow.name` at span creation when the workflow name is known, so it is available for sampling.

Also add a rego live-check manifest for invoke_workflow (expected attributes and span-name format) that enforces gen_ai.workflow.name is emitted.


eternalcuriouslearner (approved):

rads-1996 (approved):
LGTM

util/opentelemetry-util-genai/src/opentelemetry/util/genai/_workflow_invocation.py
```diff
@@ -58,10 +58,16 @@ def __init__(
...
+        """Return sampling-relevant attributes available at span creation time.
+
+        Includes ``gen_ai.workflow.name`` when a workflow name is set so it is
+        available to samplers at span creation.
+        """
         attrs: dict[str, AttributeValue] = {
             GenAI.GEN_AI_OPERATION_NAME: self._operation_name,
         }
+        if self._name:
+            attrs[GenAI.GEN_AI_WORKFLOW_NAME] = self._name
```
aabmass: Do we want to treat empty string same as `None` here?

aabmass (approved):

util/opentelemetry-util-genai/src/opentelemetry/util/genai/_workflow_invocation.py
lmolkova: no

Issue #46, opened 2026-05-19
openai-v2: StreamWrapper.headers fails for with_raw_response streaming even after #4184 fix


adityamehra:
## Description

The fix in open-telemetry/opentelemetry-python-contrib#4184 added `__getattr__` to `StreamWrapper` to proxy unknown attributes to `self.stream`. However, accessing `.headers` still fails when using `with_raw_response.create(stream=True)` with LiteLLM (or any client that accesses `raw_response.headers` directly on the stream wrapper).

This is a follow-up to open-telemetry/opentelemetry-python-contrib#4032 and open-telemetry/opentelemetry-python-contrib#4113.

> **Note**: Originally filed as open-telemetry/opentelemetry-python-contrib#4606 — moving here as the `openai-v2` instrumentation now lives in this repo.

## Root Cause

`__getattr__` proxies to `self.stream`, which is an `AsyncStream` (the result of calling `result.parse()`). However, `AsyncStream` does **not** have a `.headers` attribute — headers live on the original `LegacyAPIResponse` that was discarded before wrapping.

The call chain in `patch.py`:
```python
result = await wrapped(*args, **kwargs)        # LegacyAPIResponse (has .headers)
parsed_result = result.parse()                  # AsyncStream (no .headers)
return StreamWrapper(parsed_result, ...)        # self.stream = AsyncStream
```

When a caller then does:
```python
headers = dict(raw_response.headers)            # StreamWrapper.__getattr__("headers")
                                                # → getattr(AsyncStream, "headers")
                                                # → AttributeError  ← still crashes!
```

## Steps to Reproduce

```python
import asyncio
from unittest.mock import AsyncMock, MagicMock, patch

import httpx
from openai import AsyncAzureOpenAI
from opentelemetry.instrumentation.openai_v2 import OpenAIInstrumentor
from opentelemetry.sdk.trace import TracerProvider

SSE_CHUNKS = [
    b'data: {"id":"chatcmpl-123","object":"chat.completion.chunk","model":"gpt-4o","choices":[{"index":0,"delta":{"role":"assistant","content":"Hello"},"finish_reason":null}]}\n\n',
    b'data: {"id":"chatcmpl-123","object":"chat.completion.chunk","model":"gpt-4o","choices":[{"index":0,"delta":{"content":"!"},"finish_reason":"stop"}]}\n\n',
    b"data: [DONE]\n\n",
]

def make_mock_response():
    async def aiter_bytes(_=None):
[...]

PR #278, opened 2026-07-16
[openai] No side effects when streaming; stay resilient to edge cases with parse(custom_type)

lmolkova:
Replaces https://github.com/open-telemetry/opentelemetry-python-genai/pull/147

Today we call `parse` on legacy streams introducing side-effects (changing stream type for end users). This leads to issues like #46 (not having methods / properties that stream has without instrumentation being applied).

This PR uses different approach: 

- Wrap with_raw_response streaming transparently - parse deferred until/if user calls it
- Finalize the span from an httpx close/aclose fallback for callers that never parse. We won't record content / response attributes, but still end the span. 
- Tolerate caller-defined chunk/event types: skip chunks with unknown type


instrumentation/opentelemetry-instrumentation-genai-openai/src/opentelemetry/instrumentation/genai/openai/_raw_response.py
```diff
@@ -0,0 +1,147 @@
...
+    ) -> None:
+        super().__init__(raw_response)
+        self._self_wrap_stream = wrap_stream
+        self._self_finalize: Callable[[], None] | None = finalize
+        self._self_parsed: object | None = None
+        self._install_close_fallback(raw_response)
+
+    def _install_close_fallback(self, raw_response: object) -> None:
+        # httpx exposes a sync ``close`` and an async ``aclose``; wrap whichever
+        # the underlying response has so the appropriate one finalizes the span.
```
aabmass, line 95: Wondering does this play nice with `opentelemetry-instrumentation-httpx`?

instrumentation/opentelemetry-instrumentation-genai-openai/src/opentelemetry/instrumentation/genai/openai/_raw_response.py
```diff
@@ -0,0 +1,147 @@
...
+    ends by closing the underlying httpx response — so a fallback on its
+    ``close``/``aclose`` finalizes the span when ``parse()`` never built a
+    wrapper that would finalize it instead.
+    """
+
+    def __init__(
+        self,
+        raw_response: object,
+        wrap_stream: Callable[[object], object],
+        finalize: Callable[[], None],
```
aabmass: Is it possible to type these more tightly than `object`?

instrumentation/opentelemetry-instrumentation-genai-openai/src/opentelemetry/instrumentation/genai/openai/_raw_response.py
```diff
@@ -0,0 +1,147 @@
...
+    def _install_close_fallback(self, raw_response: object) -> None:
+        # httpx exposes a sync ``close`` and an async ``aclose``; wrap whichever
+        # the underlying response has so the appropriate one finalizes the span.
+        http_response = getattr(raw_response, "http_response", None)
+        close = getattr(http_response, "close", None)
+        if close is not None:
+            http_response.close = self._wrap_sync_close(close)
+        aclose = getattr(http_response, "aclose", None)
+        if aclose is not None:
+            http_response.aclose = self._wrap_async_close(aclose)
```
aabmass: Going along with above comment about typing the ctor inputs a little more, can we use `isinstance()` checks or some other type narrowing strategy?

instrumentation/opentelemetry-instrumentation-genai-openai/src/opentelemetry/instrumentation/genai/openai/_raw_response.py
```diff
@@ -0,0 +1,147 @@
...
+        # When ``parse()`` built a stream wrapper, that wrapper owns
+        # finalization; only finalize here for callers that never parsed.
+        if self._self_parsed is None:
+            self._finalize_once()
+
+    def _finalize_once(self) -> None:
+        # ``stop()`` is not idempotent, so finalize at most once.
+        if self._self_finalize is not None:
+            finalize, self._self_finalize = self._self_finalize, None
+            finalize()
```
aabmass, line 138: do we need it to be thread safe?

instrumentation/opentelemetry-instrumentation-genai-openai/src/opentelemetry/instrumentation/genai/openai/_raw_response.py
```diff
@@ -0,0 +1,147 @@
...
+
+    def parse(self, *args: object, **kwargs: object) -> object:
+        # We memoize the first parse regardless of the arguments it was called
+        # with, so calling parse() again with different args (e.g. a different
+        # ``to=`` type) returns the same wrapper rather than re-parsing. This is
+        # deliberate: one raw response backs one span, and re-parsing a stream
+        # would consume it twice anyway.
+        if self._self_parsed is not None:
+            return self._self_parsed
```
aabmass, line 147: also not thread safe but maybe that's ok

instrumentation/opentelemetry-instrumentation-genai-openai/src/opentelemetry/instrumentation/genai/openai/_raw_response.py
```diff
@@ -0,0 +1,147 @@
...
+    """
+
+    @classmethod
+    def wrap_result(
+        cls: type,
+        result: object,
+        invocation: GenAIInvocation,
+        capture_content: bool,
+    ) -> object:
+        if hasattr(result, "parse"):
```
aabmass: might be reasonable to just check for `isinstance(result, LegacyAPIResponse)` since you mentioned it may change in future versions and we can narrow down the code here.

alternatively if you want it to be open to changes or we end up supporting multiple versions, I'm partial to using Protocols with [`@runtime_checkable`](https://typing.python.org/en/latest/spec/protocol.html#runtime-checkable-decorator-and-narrowing-types-by-isinstance) to capture intent in the typing but maybe it's too annoying

instrumentation/opentelemetry-instrumentation-genai-openai/src/opentelemetry/instrumentation/genai/openai/_raw_response.py
```diff
@@ -0,0 +1,147 @@
...
+    The concrete wrapper must accept ``(stream, invocation, capture_content)``.
+    Call it from a streaming branch: a ``with_raw_response`` result (a
+    ``LegacyAPIResponse``, detected by its ``parse`` method) is wrapped so its
+    metadata resolves natively and ``parse()`` is deferred; a plain SDK stream
+    is wrapped directly.
+    """
+
+    @classmethod
+    def wrap_result(
+        cls: type,
```
aabmass: nit maybe a stylistic thing, but im not personally a fan of mixins like this because it silently depends on all constructors having a compatible signature

instrumentation/opentelemetry-instrumentation-genai-openai/src/opentelemetry/instrumentation/genai/openai/response_wrappers.py
```diff
@@ -178,13 +181,24 @@ def response(self):
         return _ResponseProxy(response, lambda: self._stop(None))
 
     def process_event(self, event: "ResponseStreamEvent[TextFormatT]") -> None:
-        event_type = event.type
+        # raw-response stream can be parsed into a caller-defined event type.
+        event_type = getattr(event, "type", None)
+        if not isinstance(event_type, str):
+            # Log once per stream to avoid spamming on long streams.
```
aabmass: Since it's debug level, maybe it's OK to just spam? Idk keep me honest but I'm not sure it's worth the effort

instrumentation/opentelemetry-instrumentation-genai-openai/src/opentelemetry/instrumentation/genai/openai/response_wrappers.py
```diff
@@ -178,13 +181,24 @@ def response(self):
...
+                )
+            return
+
         response: "ParsedResponse[TextFormatT] | Response | None" = getattr(
             event, "response", None
         )
 
         if response and not self._self_invocation.request_model:
-            model = response.model
+            model = getattr(response, "model", None)
```
aabmass, line 194: why do we need this change? Judging by the "cast" on L196, I think it's OK

instrumentation/opentelemetry-instrumentation-genai-openai/src/opentelemetry/instrumentation/genai/openai/_raw_response.py
```diff
@@ -0,0 +1,147 @@
...
+# SPDX-License-Identifier: Apache-2.0
+
+"""Transparent proxy for streaming ``with_raw_response`` results."""
+
+from __future__ import annotations
+
+import logging
+from typing import TYPE_CHECKING, Awaitable, Callable
+
+from openai import AsyncStream, Stream
```
aabmass, line 21: Does this cause an `ImportError` if the instrumenting is attempted but the OpenAI SDK is not installed or not a compatible version? I think that was the reason for the lazy/defensive imports we previously had, but it might be OK already depending if this module is imported lazily from the `Instrumentor`

.github/instructions/instrumentation.instructions.md
```diff
@@ -36,6 +36,11 @@ prefer opt-in or additive. Breaking changes need explicit justification in the P
 - Adding attributes to invocations produced by the util is fine.
 - Streaming responses must be instrumented by subclassing the util's `SyncStreamWrapper` /
   `AsyncStreamWrapper` (`opentelemetry.util.genai.stream`). Flag hand-rolled stream wrappers.
+- Instrumentation must not change what a call returns or when its work happens. Flag: side effects
```
aabmass: My 2c, there will always be some whack-a-mole, but we gotta keep our sanity in terms of limiting side effects or anticipating future changes. I hope this doesn't make the AI take it too far 😆 

As an extreme example (that I think is tolerable risk), some of the function wrappers added like `_wrap_async_close()` could still break users if 

- the callable is a class implementing `__call__()` (pretty common in python but not the case for HTTPX)
- user is inspecting type signatures, docstrings, etc

instrumentation/opentelemetry-instrumentation-genai-openai/src/opentelemetry/instrumentation/genai/openai/_raw_response.py
```diff
@@ -0,0 +1,147 @@
...
+                return original(*args, **kwargs)
+            finally:
+                self._finalize_close_fallback()
+
+        return _close
+
+    def _wrap_async_close(
+        self, original: Callable[..., Awaitable[object]]
+    ) -> Callable[..., Awaitable[object]]:
+        async def _aclose(*args: object, **kwargs: object) -> object:
```
aabmass, line 120: consider using https://docs.python.org/3/library/functools.html#functools.wraps

aabmass (approved):
Sorry for the nitpicky comments on typing. Approach LGTM.

instrumentation/opentelemetry-instrumentation-genai-openai/src/opentelemetry/instrumentation/genai/openai/_raw_response.py
lmolkova: LegacyAPIResponse is in internal module and there are a few more types this thing can be, added Protocols with @runtime_checkable

instrumentation/opentelemetry-instrumentation-genai-openai/src/opentelemetry/instrumentation/genai/openai/_raw_response.py
lmolkova: replaced with module function - looks much cleaner

instrumentation/opentelemetry-instrumentation-genai-openai/src/opentelemetry/instrumentation/genai/openai/response_wrappers.py
lmolkova: I believe it came up in some test  - it's needed because it breaks in corner cases

instrumentation/opentelemetry-instrumentation-genai-openai/src/opentelemetry/instrumentation/genai/openai/_raw_response.py
lmolkova: we're testing with oldest version of openai SDK, so it should be ok, also I don't believe we protect anywhere from original package not being installed

.github/instructions/instrumentation.instructions.md
lmolkova: yeah, soften it

instrumentation/opentelemetry-instrumentation-genai-openai/src/opentelemetry/instrumentation/genai/openai/_raw_response.py
lmolkova: nice, thanks!

instrumentation/opentelemetry-instrumentation-genai-openai/src/opentelemetry/instrumentation/genai/openai/_raw_response.py
lmolkova: response streams we're wrapping are not thread safe so it should be ok

instrumentation/opentelemetry-instrumentation-genai-openai/src/opentelemetry/instrumentation/genai/openai/_raw_response.py
lmolkova: made underlying utils idempotent to multiple calls, but thread safety should not be necessary - underlying objects aren't

instrumentation/opentelemetry-instrumentation-genai-openai/src/opentelemetry/instrumentation/genai/openai/_raw_response.py
lmolkova: it does, but added test. httpx instrumentation patches different methods - we don't await stream closure in http because it's super-fragile there

eternalcuriouslearner (approved):

PR #282, opened 2026-07-18
Clean up error recording

lmolkova:
Use canonical exception name when exception is thrown, use proper error codes for openai responses when error is returned


instrumentation/opentelemetry-instrumentation-genai-openai/src/opentelemetry/instrumentation/genai/openai/response_wrappers.py
```diff
@@ -188,23 +193,26 @@ def process_event(self, event: "ResponseStreamEvent[TextFormatT]") -> None:
             if model:
                 self._self_invocation.request_model = model
 
-        if event_type == "response.completed":
+        if event_type in {"response.completed", "response.incomplete"}:
```
rads-1996, line 203: Why has `response.incomplete` been moved here?

lmolkova: because incomplete response is not a failure, it's expressed in finish reason - similarly to chat calls - max_length finish reason is reported as success

rads-1996 (approved):

instrumentation/opentelemetry-instrumentation-genai-openai/src/opentelemetry/instrumentation/genai/openai/response_extractors.py
```diff
@@ -365,6 +368,21 @@ def extract_finish_reasons(response: "Response | None") -> list[str]:
...
+    """Return an ``Error`` when the response failed, else ``None``.
+
+    A failed response carries a ``ResponseError`` (``code`` + ``message``).
+    Incomplete responses (``incomplete_details``) are *not* errors — they
+    surface as a finish reason instead.
+    """
+    if Response is None or Error is None or not isinstance(response, Response):
+        return None
+    error = response.error
+    if error is None or not getattr(error, "code", None):
```
aabmass: looks like `error.code` always has a value, and it's a pydantic model so it should be validated already

```suggestion
    if error is None:
```

instrumentation/opentelemetry-instrumentation-genai-openai/src/opentelemetry/instrumentation/genai/openai/response_wrappers.py
```diff
@@ -188,23 +193,26 @@ def process_event(self, event: "ResponseStreamEvent[TextFormatT]") -> None:
...
             return
 
-        if event_type == "response.error":
-            error_type = getattr(event, "code", None) or "response.error"
-            message = getattr(event, "message", None) or error_type
-            self._fail(message, RuntimeError)
+        if event_type == "error":
+            error_type = getattr(event, "code", None) or "error"
+            message = getattr(event, "message", None)
+            self._fail(Error(type=error_type, message=message))
```
aabmass:
```suggestion
        if event.type == "error":
            error_type = event.code or "error"
            message = event.message or None
            self._fail(Error(type=error_type, message=message))
```

makes the typechecker very happy. pyright isn't able to narrow down the original object when the discriminator is extracted to a separate variable

aabmass (approved):
💥

lzchen (approved):

PR #283, opened 2026-07-18
Include reasoning tokens in google-genai output token count

lmolkova:
candidates_token_count from the google-genai response excludes thought tokens, so the reasoning tokens were missing from the gen_ai.client.token_usage metric (the span attribute happened to include them only because util summed thinking_tokens into output).

Make the instrumentation set output_tokens to the full count


eternalcuriouslearner (approved):
LGTM!! Curious if we can simulate a vcr backed test for this scenario(not a blocker). Also should we revisit this in openai and anthropic?

DylanRussell (approved):

rads-1996 (approved):
LGTM

lmolkova:
> Also should we revisit this in openai and anthropic?

[Anthropic](https://platform.claude.com/docs/en/api/beta/messages/create#beta_message.usage%20+%20(resource)%20beta.messages.output_tokens_details)  and [openai](https://developers.openai.com/api/reference/resources/beta#(resource)%20beta.responses%20%3E%20(model)%20beta_response_usage%20%3E%20(schema)%20%3E%20(property)%20output_tokens_details) include reasoning tokens into total count.

> simulate a vcr backed test for this scenario

👍

PR #284, opened 2026-07-18
Align GenericPart, Modality and FinishReason with semconv models

lmolkova:
Add `name` to InputMessage/OutputMessage, add Compaction, SystemInstructionPart, MemoryRecord, and RetrievalDocument, extend Modality with `document`, and update FinishReason to the schema enum (`tool_calls` -> `tool_call`, add `compaction`).

Message and tool-definition serialization now omits None-valued fields (new `asdict_without_none` helper) to match the semconv JSON schemas, on the span, event, and upload-hook paths.


util/opentelemetry-util-genai/src/opentelemetry/util/genai/types.py
```diff
@@ -202,26 +219,60 @@ class GenericToolDefinition:
...
     finish_reason: str | FinishReason
+    name: str | None = None
+
+
+@dataclass()
+class MemoryRecord:
+    """Represents a single memory record exposed to or produced by the model.
+
+    This model is specified as part of semconv in `GenAI messages Python models - MemoryRecord
+    <https://github.com/open-telemetry/semantic-conventions/blob/main/docs/gen-ai/non-normative/models.ipynb>`__.
```
DylanRussell: This link (and below) doesn't resolve, they point to the old repo

util/opentelemetry-util-genai/src/opentelemetry/util/genai/types.py
```diff
@@ -202,26 +219,60 @@ class GenericToolDefinition:
...
+    """
+
+    content: Any
+    id: str | None = None
+    metadata: dict[str, Any] | None = None
```
DylanRussell: I feel we should start just using AttributeValue instead of Any in these types, since they ultimately go into a span/log attribute right ?

For `Input/OutputMessage` we just end up completely serializing those (using json.dumps) types into a string.. so I think those Anys are fine to be left as is..

.github/instructions/instrumentation.instructions.md
```diff
@@ -68,6 +68,13 @@ prefer opt-in or additive. Breaking changes need explicit justification in the P
   come from `opentelemetry.semconv._incubating.attributes.gen_ai_attributes`.
 - For attributes with a well-known value set in semconv, use the generated enum from the same
   module (e.g. `GenAiOutputTypeValues` for `gen_ai.output.type`) instead of string literals.
+- Flag a dict, `TypedDict`, or other ad-hoc value built for an attribute that semconv models in
```
lmolkova:
```suggestion
- Flag `Any`, a dict, `TypedDict`, or other ad-hoc value built for an attribute that semconv models in
```

.github/instructions/util-genai.instructions.md
```diff
@@ -24,13 +24,24 @@ land the semconv change first.
...
+  - a type added or edited without citing the model it mirrors;
 
 ## 3. API backwards compatibility
 
 - Do not remove or rename public objects. Deprecate first via a docstring note pointing to the
   replacement (not `@deprecated` — unreliable).
 - Private modules and module-private objects start with `_`.
 - Default to internal (`_`-prefixed) unless instrumentations need it public.
+- Flag every `Any` in the public API (anything not under a `_`-prefixed module) — parameters,
+  return types, and dataclass fields alike. Require the narrowest type that fits: the semconv
```
eternalcuriouslearner, line 43: praise: great add.👍

eternalcuriouslearner (approved):

PR #285, opened 2026-07-18
Use weaver advice-data instead of hackery with json schemas in rego

lmolkova:
Add a `make generate-conformance-policies` target that emits the committed policies/_schemas.rego from the pinned semconv GenAI registry, so the schema policies are versioned instead of regenerated at test-run time. 

Add the auto-conformance-policies workflow to regenerate and push the policies on Renovate branches when versions.env pins are bumped.

This lays down ground to generate span validation rego with weaver


policies/_schemas.rego
lmolkova, line 1: this won't be necessary with next weaver release https://github.com/open-telemetry/weaver/pull/1539, will address in a follow up

eternalcuriouslearner (approved):

policies/_schemas.rego
lmolkova: we have new weaver, so this is removed

PR #289, opened 2026-07-20
feat(util-genai): add CompactionPart message part type

### LINKED ISSUE #270: Add a first-class CompactionPart type in util-genai
The Anthropic reference scenario in semantic-conventions (open-telemetry/semantic-conventions-genai#324) currently serialises server-side compaction message parts by hand because `opentelemetry-util-genai` has no `CompactionPart` type.

A compaction part carries an opaque **encrypted** continuity value (not the unencrypted summary the semconv `CompactionPart.content` documents), so a typed part plus handler support would let scenarios emit it directly instead of via manual `json.dumps`.

Raised by @JWinermaSplunk in review of open-telemetry/semantic-conventions-genai#324.

bhumikadangayach:
Adds CompactionPart to opentelemetry-util-genai, mirroring the semconv model exactly (type, id, content), per design discussion on #270.

A bare CompactionPart (only type set) is the expected common case — providers like Anthropic never return an unencrypted summary, only an opaque continuity value. Serialization intentionally emits explicit nulls for unset fields, consistent with existing precedent (ServerToolCall, ServerToolCallResponse) rather than introducing inconsistent omission behavior.

Fixes #270


AgentGymLeader (commented):
Reviewed from the #270 side, and ran it to be sure: checked out the branch, both new tests pass, the full util-genai suite is green (236 passed), and a live round-trip gives exactly {"id":null,"content":null,"type":"compaction"} for a bare part, in the union.

Design-wise this is exactly the shape we converged on: mirrors the semconv model as it stands (type/id/content, with id included), no continuity field ahead of #192, bare part valid with the Anthropic rationale right in the docstring, union membership with no serializer special-casing, and the wire-shape test pins explicit-null behavior to the existing ServerToolCall precedent instead of inventing a new one. The corrected models.py link is right too.

One cosmetic thing: the PR title looks like it picked up the branch name — "…message part typeFeat/compaction part". Worth trimming to "feat(util-genai): add CompactionPart message part type" before merge, since it'll likely become the squash commit title.

(Unrelated to this PR, for anyone running it standalone: the package's `test` extra alone doesn't collect these tests — test_toolcall.py imports opentelemetry.sdk, which comes from the tox testenv rather than the extra. Pre-existing, just noting it since I hit it while verifying.)

LGTM from the issue side. Once this lands I'll take the follow-up on migrating the anthropic reference scenario in semconv-genai onto the typed part, as discussed.

bhumikadangayach:
Thanks for going through this so thoroughly, @AgentGymLeader — really appreciate you actually running it rather than just reading the diff. Glad the shape landed where we converged.

Fixed the title. Noted the extra test gap for standalone test_toolcall. py runs—agreed, that's pre-existing and out of scope here, but good to have it on record.

Whenever #192 settles the continuity-value naming, happy to pick up adding that field as a follow-up. Looking forward to the #324 fixture migration on your side once this lands.

DylanRussell (approved):

DylanRussell:
LGTM. For gemini I see some discussion about compaction being reached after some number of tokens: https://ai.google.dev/gemini-api/docs/antigravity-agent#capabilities -- but I don't see that showing up in the request/response anywhere, so I don't think there's anything to do for the google-genai instrumentation

util/opentelemetry-util-genai/tests/test_toolcall.py
```diff
@@ -140,6 +142,35 @@ def test_server_tool_call_in_message():
...
+    assert part.id == "compact_001"
+    assert part.content == "Summary of earlier turns."
+    assert part.type == "compaction"
+
+    msg = InputMessage(role="assistant", parts=[part])
+    assert len(msg.parts) == 1
+    assert isinstance(msg.parts[0], CompactionPart)
+
+
+def test_compactionpart_bare_serializes_with_explicit_nulls():
```
lmolkova: is this a good behavior we should test for? I'd rather not serialize nulls

AgentGymLeader: Fair challenge — walking through it from your side, I think you're right. This test canonizes what asdict() happens to do, not a shape anyone chose, and the semconv reference corpus already votes for omission: the anthropic scenario fixtures record a bare compaction part as just {"type": "compaction"}.

One flag before dropping Nones at the shared layer: it can't be a blanket filter. ToolCallRequest.arguments / ToolCallResponse.response are Any-typed, where None can be a real value (a tool legitimately returning null vs no response recorded). The safe version drops only the Optional metadata fields (id, content) — which would also clean up the "id": null that ServerToolCall has emitted since 1.0b0, and beta seems like the right window for that. Happy either way on scoping: fold the small dict_factory into this PR and flip this test to assert {"type": "compaction"}, or land the type as-is minus this test and do the serializer as an immediate follow-up.

lmolkova: Could you please avoid excessive AI usage when talking to humans?

A good approach would be to encode it into semconv using pydantic / json schema properties where possible and special-casing tool call arguments / result. Once it's formalized in semconv, it should be ok to have per-property serialization choices.

For this PR, I'll be happy to merge with this test removed

bhumikadangayach: Done — removed the test in [commit hash]. Happy to pick up the dict_factory/semconv-property follow-up whenever that's ready.

lmolkova (approved):

PR #292, opened 2026-07-20
Minor cleanup around tool invocation tool call args / results

DylanRussell:
# Description

Updates langchain instrumentation so it doesn't bother encoding arguments because ultimately they are json dumped into an attribute, also updated the code to ignore arguments when the content capture flag is not set..

I removed the gen ai json dumps of tool call args / results  from the util library and instead updated it to take an AnyValue type, I think it's better for instrumentations to decide how they want results / arguments serialized and to do so themselves.

Most likely in the next otel python release AnyValue will soon be used everywhere and will include extended attributes so I used that for tool call args / results and attribute values, this PR will fail CI until that release happens..

## Type of change

# How has this been tested?

Unit tests

# Checklist

- [x] Unit tests added
- [x] Documentation updated

### DIFF
--- a/instrumentation/opentelemetry-instrumentation-genai-langchain/src/opentelemetry/instrumentation/genai/langchain/callback_handler.py
+++ b/instrumentation/opentelemetry-instrumentation-genai-langchain/src/opentelemetry/instrumentation/genai/langchain/callback_handler.py
@@ -440,11 +440,8 @@ class OpenTelemetryLangChainCallbackHandler(BaseCallbackHandler):
             if inputs is not None:
                 arguments = inputs
             else:
-                try:
-                    arguments = json.loads(input_str)
-                except (json.JSONDecodeError, ValueError):
-                    arguments = input_str
-
+                arguments = input_str
+                
             tool_invocation.arguments = (
                 arguments
                 if isinstance(arguments, str)
--- a/instrumentation/opentelemetry-instrumentation-genai-langchain/src/opentelemetry/instrumentation/genai/langchain/callback_handler.py
+++ b/instrumentation/opentelemetry-instrumentation-genai-langchain/src/opentelemetry/instrumentation/genai/langchain/callback_handler.py
@@ -40,6 +40,7 @@ from opentelemetry.util.genai.types import (
     Text,
     ToolCallRequest,
 )
+from opentelemetry.util.genai.utils import gen_ai_json_dumps
 
 
 class OpenTelemetryLangChainCallbackHandler(BaseCallbackHandler):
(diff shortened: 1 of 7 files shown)


instrumentation/opentelemetry-instrumentation-genai-langchain/src/opentelemetry/instrumentation/genai/langchain/callback_handler.py
```diff
@@ -431,19 +431,22 @@ def on_tool_start(
...
         )
-        tool_invocation.arguments = arguments
+        if self._telemetry_handler.should_capture_content():
+            arguments: Any
+            if inputs is not None:
+                arguments = inputs
+            else:
+                arguments = input_str
+
+            tool_invocation.arguments = (
```
lmolkova, line 444: they are json dumped only for the time being until we can populate complex attributes on spans. I'd prefer to either keep deserialization in instrumentations or make utils deserialize them when they come as string (since it's common part for all instrumentations). If we do the latter, we should also do this in other instrumentation libs and eventually util-genai will switch to complex attrs AND will start deserailizing these at the same time.

DylanRussell: How does complex attributes help here ? The hard problem here is converting arguments (which can be essentially anything) into the narrow otel attributes type, complex attributes just expands the type a tiny bit

At the moment I see only 3 instrumentations with tool invocations:

https://github.com/open-telemetry/opentelemetry-python-genai/blob/2d0d28fe3e553a3efc03c91dab042aa2f961c873/instrumentation/opentelemetry-instrumentation-genai-langchain/src/opentelemetry/instrumentation/genai/langchain/callback_handler.py#L436-L442

[instrumentation/opentelemetry-instrumentation-google-genai/src/opentelemetry/instrumentation/google_genai/tool_call_wrapper.py](https://github.com/open-telemetry/opentelemetry-python-genai/blob/2d0d28fe3e553a3efc03c91dab042aa2f961c873/instrumentation/opentelemetry-instrumentation-google-genai/src/opentelemetry/instrumentation/google_genai/tool_call_wrapper.py#L4)

https://github.com/open-telemetry/opentelemetry-python-genai/blob/2d0d28fe3e553a3efc03c91dab042aa2f961c873/instrumentation/opentelemetry-instrumentation-genai-openai-agents/src/opentelemetry/instrumentation/genai/openai_agents/processor.py#L106

^^ I don't really see what common code to pull out of this into the utils library.. I also still prefer pulling that serialization out of the utils library where it's kind of hidden.

So in general you prefer the tool arguments / results are not serialized and somehow better represented ?

lmolkova: semconv tell to have them not serialized 

https://github.com/open-telemetry/semantic-conventions-genai/blob/c26a2c21d1ee70d5231bd440c7b48d3c94ee506a/model/gen-ai/registry.yaml#L345

>       It's expected to be an object - in case a serialized string is available
>      to the instrumentation, the instrumentation SHOULD do the best effort to
>      deserialize it to an object.

how I see it can work:

- instrumentations either 
  - deserialize them into [`ToolCallArguments` ](https://github.com/open-telemetry/semantic-conventions-genai/blob/c26a2c21d1ee70d5231bd440c7b48d3c94ee506a/model/gen-ai/gen-ai-tool-call-arguments.json#L4) - we'll add a type even though it's just an alias to AnyValue
   - keep them as string since they've got string from model 
- utils in the future once complex attributes on spans are supported:
  - if it's string, try deserializing it into json object (AnyValue)
  - if it's a `ToolCallArguments` already, just set it
  - ignore otherwise
- utils now:
  - if it's a string, stamp as is
  - serialize into string otherwise

DylanRussell: Okay i think we are mostly aligned then, just leave it up to instrumentations to translate arguments into `AnyValue` and accept that as the param ?

Maybe lets just revisit once complex attributes is available as that should be pretty soon..

PR #272, opened 2026-07-15
(Openinference Migration: Langchain) - Add support for cache and reasoning token counts (PR #1)

rads-1996:
# Description

- PR 1 — Token counts. Sets `cache_creation_input_tokens, cache_read_input_tokens, and thinking_tokens` from LangChain usage_metadata.
- PR 2 — Legacy function_call. Parses the pre-tools OpenAI `additional_kwargs.function_call shape` into a ToolCallRequest on both the input path (_ai_message_parts) and the output path (on_llm_end), matching the modern tool_calls handling.
- PR 3 — Multimodal images. Converts OpenAI image_url (URL or data: base64) and Anthropic image/source blocks into Uri (URL) or Blob (base64 decoded to raw bytes). Base64 decoding matches the sibling genai-anthropic extract_base64_blob behavior so the same image yields identical Blob bytes across instrumentations.

## Type of change

Please delete options that are not relevant.

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update

# How has this been tested?

Please describe the tests that you ran to verify your changes. Provide
instructions so we can reproduce. List any relevant details for your test
configuration.

- [x] uv run tox -e py312-test-instrumentation-genai-langchain-conformance -- -q
- [x] uv run --python 3.12 tox -e lint-instrumentation-genai-langchain
- [x] uv run tox -e py312-test-instrumentation-genai-langchain -- -q

# Checklist

See [CONTRIBUTING.md](https://github.com/open-telemetry/opentelemetry-python-genai/blob/main/CONTRIBUTING.md)
for the style guide, changelog guidance, and more.

- [x] Followed the style guidelines of this project
- [x] Changelog updated if the change requires an entry
- [x] Unit tests added
- [ ] Documentation updated

PR #296, opened 2026-07-21
(Openinference Migration: Langchain): Capture multimodal image content (OpenAI `image_url` and Anthropic `image` blocks) as `Blob`/`Uri` message parts.

rads-1996:
# Description

Part of the langchain migration PRs - https://github.com/open-telemetry/opentelemetry-python-genai/pull/272

## Type of change

# How has this been tested?

configuration.

- [x] uv run tox -e py312-test-instrumentation-genai-langchain -- -q
- [x] uv run tox -e py312-test-instrumentation-genai-langchain-conformance -- -q
- [x] uv run --python 3.12 tox -e lint-instrumentation-genai-langchain

# Checklist

- [x] Unit tests added
- [ ] Documentation updated


AgentGymLeader:
@rads-1996 One thing I wanted to check on ordering. In `on_chat_model_start`, `to_input_messages()` runs before anything consults `TelemetryHandler.should_capture_content()`, so with the default `NO_CONTENT` the data URI and Anthropic base64 images still get decoded to `bytes` via `_decode_base64`.

That means a config that doesn't record content still copies large images. If that isn't intended, either skip the media decode when capture is off, or add a test asserting `_decode_base64` isn't called under `NO_CONTENT`.

If the parse is meant to run unconditionally so the completion hook has something to work with, that's fine, I just wanted to confirm that's the assumption.

rads-1996:
> @rads-1996 One thing I wanted to check on ordering. In `on_chat_model_start`, `to_input_messages()` runs before anything consults `TelemetryHandler.should_capture_content()`, so with the default `NO_CONTENT` the data URI and Anthropic base64 images still get decoded to `bytes` via `_decode_base64`.
> 
> That means a config that doesn't record content still copies large images. If that isn't intended, either skip the media decode when capture is off, or add a test asserting `_decode_base64` isn't called under `NO_CONTENT`.
> 
> If the parse is meant to run unconditionally so the completion hook has something to work with, that's fine, I just wanted to confirm that's the assumption.

@AgentGymLeader Thanks for the callout. I have gated decoding and will skipped when the content capture mode is `NO_CONTENT`.

lzchen (approved):

instrumentation/opentelemetry-instrumentation-genai-langchain/tests/test_callback_handler.py
lmolkova, line 1: can you please add conformance test? thanks!

instrumentation/opentelemetry-instrumentation-genai-langchain/.changelog/296.added
```diff
@@ -0,0 +1 @@
+(Openinference Migration: Langchain) - Capture multimodal image content (OpenAI ``image_url`` and Anthropic ``image`` blocks) as ``Blob``/``Uri`` message parts.
```
lmolkova: please update openai and anthropic instrumentations to hahve feature parity.  Ok to do in a separate PR. Thanks!

instrumentation/opentelemetry-instrumentation-genai-langchain/src/opentelemetry/instrumentation/genai/langchain/utils.py
```diff
@@ -75,6 +80,89 @@ def _normalize_role(message: BaseMessage) -> str:
     return _ROLE_MAP.get(message.type, message.type)
 
 
+def _decode_base64(data: str) -> bytes | None:
+    # Skip the decode entirely when message content is not being captured;
+    # the resulting bytes would never be emitted under ``NO_CONTENT``.
+    if get_content_capturing_mode() is ContentCapturingMode.NO_CONTENT:
```
lmolkova: this one actually checks env var and is now used on a hot path, please  use handler.should_capture_content() somewhere up the stack and don't call this method when content capturing is disabled

instrumentation/opentelemetry-instrumentation-genai-langchain/src/opentelemetry/instrumentation/genai/langchain/utils.py
```diff
@@ -75,6 +80,89 @@ def _normalize_role(message: BaseMessage) -> str:
...
+                content=decoded,
+            )
+        if source_type == "url":
+            source_url = source_dict.get("url")
+            if isinstance(source_url, str) and source_url:
+                return _image_from_url(source_url)
+    return None
+
+
+def _image_from_url(url: str) -> MessagePart | None:
```
lmolkova: should we have this whole method in utils? the data:// is pretty generic approach, nothing specific to openai https://developer.mozilla.org/en-US/docs/Web/URI/Reference/Schemes/data

instrumentation/opentelemetry-instrumentation-genai-langchain/src/opentelemetry/instrumentation/genai/langchain/utils.py
```diff
@@ -19,6 +20,8 @@
     gen_ai_attributes as GenAIAttributes,
 )
 from opentelemetry.util.genai.types import (
+    Blob,
```
lmolkova: it's the first time we're using this type in this repo, mind renaming it to BlobPart? Same with Uri, can we rename it to `UriPart` - that's how they are called in semconv and it seems we're inconsistent - https://github.com/open-telemetry/semantic-conventions-genai/blob/main/docs/gen-ai/non-normative/models.py

instrumentation/opentelemetry-instrumentation-genai-langchain/src/opentelemetry/instrumentation/genai/langchain/utils.py
rads-1996: Yeah make sense, I can then reuse it for openai and anthropic.

instrumentation/opentelemetry-instrumentation-genai-langchain/src/opentelemetry/instrumentation/genai/langchain/utils.py
rads-1996: Sounds good. I will make changes in the utils. I believe the same change is needed for some other data classes as well. I think I will make a separate PR to address that.

rads-1996: @lmolkova `Blob` and `Uri` both have previously been used in other instrumentations, `Blob` has been used in the anthropic and google-genai instrumentation and `Uri` in google-genai. I have a PR out which aligns the names with semconv - https://github.com/open-telemetry/opentelemetry-python-genai/pull/365. Should we have backward compatibility aliases for downstream users with deprecation messages?

instrumentation/opentelemetry-instrumentation-genai-langchain/.changelog/296.added
rads-1996: Tracking in this issue - https://github.com/open-telemetry/opentelemetry-python-genai/issues/348

instrumentation/opentelemetry-instrumentation-genai-langchain/src/opentelemetry/instrumentation/genai/langchain/utils.py
rads-1996: Addressed.

instrumentation/opentelemetry-instrumentation-genai-langchain/tests/test_callback_handler.py
rads-1996: Addressed.

instrumentation/opentelemetry-instrumentation-genai-langchain/src/opentelemetry/instrumentation/genai/langchain/callback_handler.py
```diff
@@ -162,7 +166,10 @@ def on_chain_end(
             self._invocation_manager.delete_invocation_state(run_id)
             return
 
-        invocation.output_messages = make_last_output_message(outputs)
+        capture_content = self._telemetry_handler.should_capture_content()
+        invocation.output_messages = make_last_output_message(
+            outputs, capture_content
+        )
```
lmolkova: why not 

```suggestion
        if capture_content: 
            invocation.output_messages = make_last_output_message(
                outputs
            )
```

?

instrumentation/opentelemetry-instrumentation-genai-langchain/src/opentelemetry/instrumentation/genai/langchain/callback_handler.py
```diff
@@ -270,7 +277,8 @@ def on_chat_model_start(
         # :func:`to_input_messages` produce spec-conformant ``InputMessage`` s
         # with proper roles, tool-call requests, tool results, and reasoning.
         flattened: list[BaseMessage] = [msg for sub in messages for msg in sub]
-        input_messages = to_input_messages(flattened)
+        capture_content = self._telemetry_handler.should_capture_content()
+        input_messages = to_input_messages(flattened, capture_content)
```
lmolkova:
```suggestion
        if capture_content:
            input_messages = to_input_messages(flattened)
```

?

util/opentelemetry-util-genai/src/opentelemetry/util/genai/utils.py
```diff
@@ -36,6 +41,44 @@ def get_content_capturing_mode() -> ContentCapturingMode:
         return ContentCapturingMode.NO_CONTENT
 
 
+def decode_base64(data: str, capture_content: bool = False) -> bytes | None:
```
lmolkova: does it need to be public? also it seems we can just avoid calling to_input / to_output if content is disabled and add a docstring for public utils methods that is expected from caller to check if content is enabled and not call this method if it's not.


```suggestion
def _decode_base64(data: str) -> bytes | None:
```

instrumentation/opentelemetry-instrumentation-genai-langchain/src/opentelemetry/instrumentation/genai/langchain/utils.py
lmolkova: oh, good catch! not sure how I missed them. Yes, let's do aliases

util/opentelemetry-util-genai/src/opentelemetry/util/genai/utils.py
rads-1996: I think it needs to be public because this method is called directly as well and since we have moved it to the utils, we would need another public method which then calls it internally.

rads-1996: For the second part, it makes sense, and I will make the changes you have suggested in the other comments above.

instrumentation/opentelemetry-instrumentation-genai-langchain/src/opentelemetry/instrumentation/genai/langchain/utils.py
rads-1996: @lmolkova This has been addressed in https://github.com/open-telemetry/opentelemetry-python-genai/pull/365. That PR needs to be reviewed and merged before the current one can be updated with the new names for the data classes. PTAL when you have a moment. Thanks!

instrumentation/opentelemetry-instrumentation-genai-langchain/src/opentelemetry/instrumentation/genai/langchain/utils.py
```diff
@@ -75,6 +77,59 @@ def _normalize_role(message: BaseMessage) -> str:
...
+        if isinstance(image_url, str):
+            url = image_url
+        elif isinstance(image_url, dict):
+            image_url_dict = cast(dict[str, Any], image_url)
+            raw_url = image_url_dict.get("url")
+            url = raw_url if isinstance(raw_url, str) else None
+        if not url:
+            return None
+        return image_from_url(url)
+    if block_type == "image":
```
lmolkova: LangChain's own standard image blocks are silently dropped - only the raw provider shapes are handled. `ImageContentBlock` in langchain-core 1.x is `{"type": "image", "url"/"base64", "mime_type"}`, and 0.3 uses `source_type` + `data`; neither has a `source` key. Since standard blocks are the documented v1 way to pass images, this misses the common case (the whole message is dropped, not just the image part).

This fails on the branch - all four cases:

```python
@pytest.mark.parametrize(
    "block,expected",
    [
        ({"type": "image", "base64": B64, "mime_type": "image/png"}, Blob),
        ({"type": "image", "url": "https://example.com/a.png"}, Uri),
        ({"type": "image", "source_type": "base64", "data": B64, "mime_type": "image/png"}, Blob),
        ({"type": "image", "source_type": "url", "url": "https://e/b.png"}, Uri),
    ],
)
def test_langchain_standard_image_blocks_are_captured(block, expected):
    messages = to_input_messages([HumanMessage(content=[block])])
    assert messages, "message dropped entirely - no parts extracted"
    assert any(isinstance(p, expected) for p in messages[0].parts)
```

Please handle these shapes and add the tests.

util/opentelemetry-util-genai/src/opentelemetry/util/genai/utils.py
```diff
@@ -36,6 +41,46 @@ def get_content_capturing_mode() -> ContentCapturingMode:
...
+    if url.startswith("data:"):
+        header, _, payload = url[len("data:") :].partition(",")
+        mime_type = header.split(";", 1)[0] or None
+        if ";base64" in header:
+            decoded = decode_base64(payload)
+            if decoded is None:
+                return None
+            content = decoded
+        else:
+            content = payload.encode("utf-8")
```
lmolkova: A non-base64 `data:` URL is percent-encoded (RFC 2397), so this stores the escaped text rather than the bytes. Use `urllib.parse.unquote_to_bytes(payload)`.

Fails on the branch:

```python
def test_image_from_url_percent_encoded_data_url():
    part = image_from_url("data:image/svg+xml,%3Csvg%2F%3E")
    assert part.content == b"<svg/>"   # got b"%3Csvg%2F%3E"
```

Please fix and add the test.

util/opentelemetry-util-genai/src/opentelemetry/util/genai/utils.py
```diff
@@ -36,6 +41,46 @@ def get_content_capturing_mode() -> ContentCapturingMode:
         return ContentCapturingMode.NO_CONTENT
 
 
+def decode_base64(data: str) -> bytes | None:
+    """Called only when content capture is enabled
```
lmolkova: nit: docstring of a public function should say what it does - this summary line is a caller precondition.

instrumentation/opentelemetry-instrumentation-genai-langchain/src/opentelemetry/instrumentation/genai/langchain/utils.py
```diff
@@ -29,6 +30,7 @@
     ToolCallResponse,
     ToolDefinition,
 )
+from opentelemetry.util.genai.utils import decode_base64, image_from_url
```
lmolkova, line 35: The langchain package still pins `opentelemetry-util-genai >= 1.1b0`, so the `oldest` job installs the released util-genai and fails on this import:

```
ImportError: cannot import name 'decode_base64' from 'opentelemetry.util.genai.utils'
```

Other packages that depend on unreleased util-genai use `>= 1.1b0.dev, <2` - langchain's pyproject needs the same bump.

instrumentation/opentelemetry-instrumentation-genai-langchain/src/opentelemetry/instrumentation/genai/langchain/utils.py
```diff
@@ -75,6 +77,88 @@ def _normalize_role(message: BaseMessage) -> str:
...
+      URL becomes a :class:`Blob`; any other URL becomes a :class:`Uri`.
+    - Anthropic style ``{"type": "image", "source": {...}}`` where ``source``
+      is either ``{"type": "base64", "media_type": ..., "data": ...}`` (→
+      :class:`Blob`) or ``{"type": "url", "url": ...}`` (→ :class:`Uri`).
+    - langchain-core 0.3 standard blocks ``{"type": "image", "source_type":
+      "base64"|"url", "data"/"url": ..., "mime_type": ...}``.
+    - langchain-core 1.x standard blocks ``{"type": "image",
+      "base64"|"url": ..., "mime_type": ...}``.
+    """
+    block_type = item.get("type")
```
lmolkova, line 139: Responses-API blocks are missed. langchain-openai passes `{"type": "input_image", "image_url": ...}` (and `input_text`) through verbatim, so with `use_responses_api=True` the image is dropped - and since `to_input_messages` skips messages with no parts, an image-only message disappears from `gen_ai.input.messages` entirely rather than losing just the image.

Same for standard blocks that carry `file_id` instead of `base64`/`url`.

All four fail on the branch:

```python
@pytest.mark.parametrize(
    "block,expected",
    [
        ({"type": "input_image", "image_url": f"data:image/png;base64,{B64}"}, Blob),
        ({"type": "input_image", "image_url": "https://example.com/a.png"}, Uri),
    ],
)
def test_responses_api_input_image_blocks_are_captured(block, expected):
    messages = to_input_messages([HumanMessage(content=[block])])
    assert messages, "message dropped entirely - no parts extracted"
    assert any(isinstance(p, expected) for p in messages[0].parts)


def test_message_survives_unconvertible_image_block():
    # provider-hosted image, no bytes and no url to record
    messages = to_input_messages(
        [HumanMessage(content=[{"type": "image", "file_id": "file-123"}])]
    )
    assert messages, "message dropped entirely"


def test_text_kept_when_image_block_is_unconvertible():
    messages = to_input_messages(
        [
            HumanMessage(
                content=[
                    {"type": "input_text", "text": "what is in this image?"},
                    {"type": "input_image", "image_url": "https://e/a.png"},
                ]
            )
        ]
    )
    assert messages, "message dropped entirely"
    assert any(isinstance(p, Text) for p in messages[0].parts)
```

Please handle `input_image`/`input_text`, stop dropping the whole message when no part can be built, and add the tests.

util/opentelemetry-util-genai/src/opentelemetry/util/genai/utils.py
```diff
@@ -4,15 +4,21 @@
...
 from opentelemetry.util.genai.environment_variables import (
     OTEL_INSTRUMENTATION_GENAI_CAPTURE_MESSAGE_CONTENT,
     OTEL_INSTRUMENTATION_GENAI_EMIT_EVENT,
 )
-from opentelemetry.util.genai.types import ContentCapturingMode
+from opentelemetry.util.genai.types import (
+    Blob,
+    ContentCapturingMode,
+    MessagePart,
+    Uri,
```
lmolkova: #365 landed, so we can start using`BlobPart`/`UriPart` here

instrumentation/opentelemetry-instrumentation-genai-langchain/tests/test_llm_call.py
```diff
@@ -196,6 +216,144 @@ def test_chat_openai_gpt_3_5_turbo_model_llm_call_with_error(
...
+            "(explicit n/temperature); only the modern cassette is recorded"
+        )
+
+    # The content-capture flag is cached when the handler is built at
+    # ``instrument()`` time, so set it before instrumenting (not after).
+    monkeypatch.setenv(
+        "OTEL_INSTRUMENTATION_GENAI_CAPTURE_MESSAGE_CONTENT", "SPAN_ONLY"
+    )
+    instrumentor = LangChainInstrumentor()
+    instrumentor.instrument(
```
lmolkova: nit: the three new tests instrument manually with monkeypatch + try/finally while this PR converts the rest of the file to the `instrument(..., content_capture=...)` helper. Use the helper here too.

instrumentation/opentelemetry-instrumentation-genai-langchain/tests/test_llm_call.py
```diff
@@ -56,6 +61,13 @@ def _langchain_openai_version() -> tuple:
 # cannot hold on those versions.
 _supports_reasoning_token_details = _langchain_openai_version() >= (0, 2, 1)
 
+_REAL_PNG_B64 = (
+    "iVBORw0KGgoAAAANSUhEUgAAADAAAAAwCAIAAADYYG7QAAAARklEQVR42u3X"
+    "QQ0AIAwAsSnZG4lInJxJwMRICGlyAvq9yF1PFUBAQEBAQBdAXWskICAgICAg"
+    "ICAgICAgIOcKBAQEBPQd6ACUHHNEU5qggAAAAABJRU5ErkJggg=="
+)
+_REAL_PNG_BYTES = base64.b64decode(_REAL_PNG_B64)
```
lmolkova: nit: `_REAL_PNG_BYTES` is unused.

instrumentation/opentelemetry-instrumentation-genai-langchain/src/opentelemetry/instrumentation/genai/langchain/utils.py
rads-1996: @lmolkova I don't believe that will work, because the oldest job will still pull 1.1b0 because of the `UV_RESOLUTION=lowest-direct`, env var setting. I think we should bump the pyproject to `>=1.2b0.dev, <2` and the oldest.requirements should still install the editable version of the utils to pull in the unreleased changes. If we don't bump the pyproject then the `check-deps` will fail.

Side Note: I am still working on this PR, will update the comments once it is ready for another look. Thanks a lot for the feedback so far.

instrumentation/opentelemetry-instrumentation-genai-langchain/tests/test_llm_call.py
rads-1996: It is being used in the updated tests.

instrumentation/opentelemetry-instrumentation-genai-langchain/tests/test_llm_call.py
rads-1996: Updated the tests and replaced the manual monkeypatch with the helper.

util/opentelemetry-util-genai/src/opentelemetry/util/genai/utils.py
rads-1996: Updated

instrumentation/opentelemetry-instrumentation-genai-langchain/src/opentelemetry/instrumentation/genai/langchain/utils.py
rads-1996: Addressed.

instrumentation/opentelemetry-instrumentation-genai-langchain/src/opentelemetry/instrumentation/genai/langchain/utils.py
rads-1996: @lmolkova The PR is ready for another look. Thanks!

instrumentation/opentelemetry-instrumentation-genai-langchain/.changelog/296.added
```diff
@@ -0,0 +1 @@
+(OpenInference Migration: LangChain) - Capture multimodal image content (OpenAI ``image_url`` and Responses API ``input_image``, Anthropic ``image``, and LangChain standard ``image`` blocks) as ``Blob``/``Uri``/``File`` message parts.
```
lmolkova:
```suggestion
(OpenInference Migration: LangChain) - Capture multimodal image content (OpenAI ``image_url`` and Responses API ``input_image``, Anthropic ``image``, and LangChain standard ``image`` blocks) as ``BlobPart``/``UriPart``/``FilePart`` message parts.
```

lmolkova (approved):

PR #302, opened 2026-07-22
Pass completion hook from all instrumentations, update docs

lmolkova:
Make all instrumentations set hook, update docs, agents.md, add tests


eternalcuriouslearner (approved):
LGTM!! Couple of questions:

1. do we need this in google genai?
2. Is security an intention behind this? I think having prompts under span attributes can be a real security concern if they contain PII so just got curious on intentions.

lmolkova:
> LGTM!! Couple of questions:
> 
> 1. do we need this in google genai?
> 2. Is security an intention behind this? I think having prompts under span attributes can be a real security concern if they contain PII so just got curious on intentions.

it's needed everywhere (and it's already in google-genai) - it's a way to redirect chat history to a separate storage (not so expensive and more secure than telemetry). It's a way NOT to store prompts on telemetry.

DylanRussell (approved):

PR #308, opened 2026-07-24
Add CrewAI instrumentation boilerplate

venkata-pavani:
# Description
This PR only adds the boilerplate. The actual CrewAI operation instrumentation and SDK patching will come in follow-up PRs.
# How has this been tested?

CrewAI lifecycle tests: 3 passed
Ruff lint and formatting: passed
README validation: passed
License-header validation: passed
Generated-file check: passed
uv lock --check: passed
CrewAI source type checking: passed

# Checklist

- [ ] Unit tests added
- [ ] Documentation updated

Note : Added a temporary typecheck-specific dependency override for CrewAI’s OpenTelemetry `<1.43` constraint; it can be removed once CrewAI supports the repository’s `~=1.43` line.

Because the typecheck environment installs `crewai[instruments]`, but CrewAI restricts OpenTelemetry below `1.43` while this repository requires `~=1.43`; without the override, dependency installation fails before Pyright can run.


venkata-pavani:
Hi All, I added a temporary typecheck-specific dependency override for CrewAI’s OpenTelemetry <1.43 constraint; it can be removed once CrewAI supports the repository’s ~=1.43 line.

Because the typecheck environment installs crewai[instruments], but CrewAI restricts OpenTelemetry below 1.43 while this repository requires ~=1.43; without the override, dependency installation fails before Pyright can run.


All checks gone through successfully now. Please let me know if you have any issues or feedback on this

instrumentation/opentelemetry-instrumentation-genai-crewai/src/opentelemetry/instrumentation/genai/crewai/version.py
```diff
@@ -0,0 +1,4 @@
+# Copyright The OpenTelemetry Authors
+# SPDX-License-Identifier: Apache-2.0
+
+__version__ = "1.0b0.dev"
```
lmolkova:
```suggestion
__version__ = "1.1b0.dev"
```

to align with other packages in the repo

instrumentation/opentelemetry-instrumentation-genai-crewai/README.rst
```diff
@@ -0,0 +1,28 @@
...
+
+    from opentelemetry.instrumentation.genai.crewai import CrewAIInstrumentor
+
+    CrewAIInstrumentor().instrument()
+
+Configuration
+-------------
+
+Message content capture can be configured with the
+``OTEL_INSTRUMENTATION_GENAI_CAPTURE_MESSAGE_CONTENT`` environment variable.
```
lmolkova: please copy-paste this from some other package where we list supported values

pyproject.toml
```diff
@@ -26,6 +27,15 @@ dependencies = [
 [tool.uv]
 package = false # https://docs.astral.sh/uv/reference/settings/#package
 required-version = ">=0.6.0"
+# crewai hard-pins the OpenTelemetry core stack to a narrow window
```
lmolkova, line 33: it seems crewai comes with its own native instrumentation - could you please share some research on it - why do we need another one in otel? if crewai doesn't follow semantic conventions, how far are they from following them? was there any communication with crewai on github to support otel semantic conventions properly? Can you please link issues / PRs?

pyproject.toml
```diff
@@ -26,6 +27,15 @@ dependencies = [
...
+# (opentelemetry-api/-sdk >=1.42,<1.43) that can't coexist with this
+# workspace's ~=1.43 floor. Override crewai's caps so the workspace lock
+# resolves against the versions the util packages require. This is the
+# lock-time analog of the conformance-only UV_OVERRIDE wired in tox.ini.
+override-dependencies = [
+  "opentelemetry-api ~= 1.43",
```
lmolkova, line 39: this contradicts above statement 

> opentelemetry-api/-sdk >=1.42,<1.43

lmolkova (dismissed):
thanks for the contribution @venkata-pavani !

Can you please check on crewai native otel support? If there is a native instrumentation, we should try fixing it instead of adding a new one. 

Thanks!

pyproject.toml
venkata-pavani: @lmolkova Yes, CrewAI has its own native OpenTelemetry instrumentation and it's not aligned with GENAI semantic conventions.

I felt adding OpenTelemetry GenAI semantic conventions to native library is a larger effort as CrewAI is currently focused adding additional telemetry as per conventions. Here is the open PR https://github.com/crewAIInc/crewAI/pull/6304.

Additionally while I was researching around the implementation I saw LangChain and LangGraph also provide native tracing through LangSmith but we have GENAI aligned instrumentation here. Because of these reasons I felt adding genai aligned instrumentation is better here.

However I am not completely sure about the opentelemetry best practices. If you feel we should work with crewai to get genai semantic conventions in their sdk I can open up an issue and work with them. Let me know your thoughts.

lmolkova: thanks for the context @venkata-pavani !
Trying crewai out, it indeed seems very far from OTel GenAI semconv 

image

and doesn't even propagate context properly so that underlying OpenAI instrumentation is not correlated.

The fact that https://github.com/crewAIInc/crewAI/pull/6304 sits there for a month without any comments is a good indication that it's not a priority for crewai. I think this is a good justification to provide instrumentation here.

Could you please still create an issue in crewAI and point to this PR and this repo and kindly ask crewai folks to reach out if they want to adopt conventional instrumentation?

venkata-pavani: @lmolkova Thanks for the review !
I raised an issue with crewAI team https://github.com/crewAIInc/crewAI/issues/6723 and meanwhile I will be working on other changes mentioned on this PR. 

If they comeback to us on this ...I will co-ordinate accordingly further on the implmentation

lmolkova: thanks @venkata-pavani ! one more question: is crewai native instrumentation enabled by default? is there a way for us to disable it when this instrumentation is enabled? We can support an option to keep their telemetry coming too, but not by default.

venkata-pavani: @lmolkova Yes, CrewAI’s native telemetry appears to be enabled by default in the current implementation, as seen in https://github.com/venkata-pavani/crewAI/blob/d52d0a162814624095e58116c7bb6f58924e52b2/lib/crewai-core/src/crewai_core/telemetry.py#L123 .

For our instrumentation, we can keep the behavior opt-in and allow users to retain CrewAI’s native telemetry if they choose, while not enabling it by default. 

I will make sure I will pick this up in succeeding PR's while implementing this!

eternalcuriouslearner (approved):

lmolkova (approved):

lmolkova:
@venkata-pavani thanks! could you please resolve conflicts and this one will be good to go!

venkata-pavani:
> @venkata-pavani thanks! could you please resolve conflicts and this one will be good to go!

@lmolkova I just now resolved all conflicts on this PR.Could you please check? ...Thanks !

instrumentation/opentelemetry-instrumentation-genai-crewai/.changelog/.gitignore
```diff
@@ -0,0 +1,5 @@
+!*.added
+!*.changed
+!*.deprecated
+!*.removed
+!*.fixed
```
lmolkova:
```suggestion
 !.gitignore
```

PR #311, opened 2026-07-26
Add QwenPaw instrumentation

123liuziming:
# Description

Adds `opentelemetry-instrumentation-genai-qwenpaw`, instrumentation for [QwenPaw](https://github.com/agentscope-ai/QwenPaw), a personal assistant application built on AgentScope. Each user turn handled by `AgentRunner.query_handler` is traced as one `invoke_agent` span (`gen_ai.agent.id`, `gen_ai.agent.name`, `gen_ai.conversation.id`, and opt-in input/output message content), plus the `gen_ai.client.operation.duration` metric, all through `opentelemetry-util-genai`.

QwenPaw delegates model (LLM) and tool execution to AgentScope, so this package intentionally emits no `chat` or `execute_tool` spans and the conformance suite covers none — those operations belong to the AgentScope instrumentation.

Supported range is `qwenpaw >= 1.1.0, < 2.0.0` (2.0 removed `AgentRunner.query_handler`).

## Type of change

# How has this been tested?

- [x] `tox -e py310-test-instrumentation-genai-qwenpaw-oldest` (qwenpaw 1.1.0 via lowest-direct)
- [x] `tox -e py312-test-instrumentation-genai-qwenpaw-latest` (qwenpaw 1.1.12.post3)
- [x] `tox -e py313-test-instrumentation-genai-qwenpaw-conformance` (weaver live-check, 0 violations)
- [x] `tox -e lint-instrumentation-genai-qwenpaw`, `precommit`, `typecheck`, `readme`

# Checklist

- [x] Unit tests added
- [x] Documentation updated

Made with [Cursor](https://cursor.com)


instrumentation/opentelemetry-instrumentation-genai-qwenpaw/src/opentelemetry/instrumentation/genai/qwenpaw/__init__.py
```diff
@@ -0,0 +1,117 @@
...
+
+__all__ = ["QwenPawInstrumentor"]
+
+_RUNNER_MODULE = "qwenpaw.app.runner.runner"
+
+
+class QwenPawInstrumentor(BaseInstrumentor):
+    """An instrumentor for the QwenPaw application runner.
+
+    Traces ``AgentRunner.query_handler`` as an ``invoke_agent`` span and
```
eternalcuriouslearner, line 76: praise: appreciate the doc comments. They’re really helpful in understanding context.

instrumentation/opentelemetry-instrumentation-genai-qwenpaw/src/opentelemetry/instrumentation/genai/qwenpaw/patch.py
```diff
@@ -0,0 +1,149 @@
...
+
+
+def make_query_handler_wrapper(
+    handler: TelemetryHandler,
+) -> Callable[..., Any]:
+    """Factory for the ``wrapt`` wrapper bound to *handler*."""
+
+    def query_handler_wrapper(
+        wrapped: Callable[..., AsyncGenerator[Any, None]],
+        instance: Any,
```
eternalcuriouslearner: question(non-blocking): can we avoid Any here?

eternalcuriouslearner (commented):
LGTM!! Minor nit and can you please resolve the merge conflicts?

eternalcuriouslearner (approved):

instrumentation/opentelemetry-instrumentation-genai-qwenpaw/tests/conformance/invoke_agent.py
```diff
@@ -0,0 +1,86 @@
...
+    expected_metrics = (
+        "gen_ai.client.operation.duration",
+        "gen_ai.client.operation.time_to_first_chunk",
+    )
+    # The streamed turn records time_to_first_chunk, whose required
+    # gen_ai.provider.name cannot be set: QwenPaw delegates model calls to
+    # AgentScope, so no provider applies to the invoke_agent invocation.
+    expected_violations = (
+        ExpectedViolation(
+            advice_id="required_attribute_not_present",
```
lmolkova: it's no longer a required attribute on invoke_agent internal span, can you please check if this suppression is necessary?

instrumentation/opentelemetry-instrumentation-genai-qwenpaw/src/opentelemetry/instrumentation/genai/qwenpaw/patch.py
```diff
@@ -0,0 +1,154 @@
...
+                exc_info=True,
+            )
+        return False
+
+
+def _build_invocation(
+    handler: TelemetryHandler,
+    instance: object,
+    msgs: Any,
+    request: Any,
```
lmolkova: can we be more precise with types here and avoid using `Any` ?

instrumentation/opentelemetry-instrumentation-genai-qwenpaw/src/opentelemetry/instrumentation/genai/qwenpaw/patch.py
```diff
@@ -0,0 +1,154 @@
...
+) -> AgentInvocation:
+    # `agent_name` (config display name with a built-in fallback) was added
+    # during the 1.1.x line, so probe for it defensively.
+    agent_name = non_empty_str(getattr(instance, "agent_name", None))
+    invocation = handler.invoke_local_agent(agent_name=agent_name)
+    agent_id = non_empty_str(getattr(instance, "agent_id", None))
```
lmolkova: agent id is not used for internal agents because for them it's usually some form of instance-specific identifier that changes between app restarts and gen_ai.agent.id is sticky - https://github.com/open-telemetry/semantic-conventions-genai/blob/8d3e4a0f3c34a46f6edb9c71e8666e02e6bf3958/docs/registry/attributes/gen-ai.md?plain=1#L75

what's agent_id here? do we need to record it?

instrumentation/opentelemetry-instrumentation-genai-qwenpaw/src/opentelemetry/instrumentation/genai/qwenpaw/patch.py
```diff
@@ -0,0 +1,154 @@
...
+    input_messages_from_msgs,
+    non_empty_str,
+    output_message_from_yield_item,
+    parse_query_handler_call,
+)
+
+_logger = logging.getLogger(__name__)
+
+
+class QueryHandlerStreamWrapper(AsyncStreamWrapper[object]):
```
lmolkova, line 34: Telemetry is finalized only through the proxy's own methods, but the real callers do not always call them. [`agents/acp/server.py:572`](https://github.com/agentscope-ai/QwenPaw/blob/v1.1.9/src/qwenpaw/agents/acp/server.py#L572) breaks out of the `async for` when the client cancels, and the runner treats `asyncio.CancelledError` as a normal path ([runner.py:863](https://github.com/agentscope-ai/QwenPaw/blob/v1.1.9/src/qwenpaw/app/runner/runner.py#L863)), so a half consumed turn is expected traffic. When the caller drops the proxy, GC hands the raw async generator to the event loop's finalizer, which calls `aclose()` on the underlying generator directly and never reaches `QueryHandlerStreamWrapper.aclose()`. The span is never ended and never exported.

This test fails today with zero spans collected:

```python
@pytest.mark.asyncio
async def test_break_mid_stream_finalizes_span(
    instrument_no_content, runner_module, span_exporter
):
    async def multi_chunk(request, msgs, runner):
        for i in range(3):
            yield assistant_reply(f"chunk-{i}"), i == 2

    runner = runner_module.AgentRunner(agent_id="entry-agent")
    with patched_command_path(runner_module, multi_chunk):
        async for _ in runner.query_handler(user_command_msgs(), make_request()):
            break
    gc.collect()

    (span,) = span_exporter.get_finished_spans()
    assert span.name.startswith("invoke_agent")
```

Finalization has to hang off the generator's own teardown, so that `GeneratorExit` reaches a `finally`, rather than off proxy methods the caller never calls. No proxy-based wrapper can close this hole, since GC finalizes the underlying generator without going through it. Filed as #386 against util-genai, since `qwen-agent` has the same shape and the fix belongs there.

Fine to land as is for now: it only triggers when a caller stops consuming early, and the leaked context dies with the request task. Worth knowing that it is not purely a lost span though, as later spans in that task nest under the span that never ended.

instrumentation/opentelemetry-instrumentation-genai-qwenpaw/src/opentelemetry/instrumentation/genai/qwenpaw/patch.py
```diff
@@ -0,0 +1,154 @@
...
+
+    def __init__(
+        self,
+        stream: AsyncGenerator[object, None],
+        invocation: AgentInvocation,
+        capture_content: bool,
+    ) -> None:
+        super().__init__(cast(Any, stream))
+        self._self_gen = stream
+        self._self_invocation = invocation
```
lmolkova: Assigning to the base class's `_self_invocation` after `super().__init__(stream)` opts this INTERNAL `invoke_agent` span into the streamed-call metrics. A two chunk turn emits `gen_ai.client.operation.time_to_first_chunk` and `gen_ai.client.operation.time_per_output_chunk`, which is why the conformance scenario has to list TTFC in `expected_metrics` and suppress a `gen_ai.provider.name` violation.

It also lands in a half state: since the invocation was not passed to `super().__init__`, `_request_stream` stays unset, so the span has no `gen_ai.request.stream` while streamed metrics are recorded.

`qwen-agent` keeps the agent invocation in a separate field (`_self_agent_invocation`) for this reason. Doing the same here lets you drop the TTFC expectation and the declared violation.

instrumentation/opentelemetry-instrumentation-genai-qwenpaw/src/opentelemetry/instrumentation/genai/qwenpaw/patch.py
```diff
@@ -0,0 +1,154 @@
...
+        if output_message is not None:
+            self._self_output_message = output_message
+
+    def _on_stream_end(self) -> None:
+        if self._self_output_message is not None:
+            self._self_invocation.output_messages = [self._self_output_message]
+        self._self_invocation.stop()
+
+    def _on_stream_error(self, error: BaseException) -> None:
+        self._self_invocation.fail(error)
```
lmolkova: The accumulated output is set in `_on_stream_end` but dropped here, so a mid-stream failure loses content already captured. `test_stream_side_error_reraises_and_fails_span` yields `"partial"` before raising and the error span carries no `gen_ai.output.messages`. OpenAI's wrapper applies accumulated state on both paths.

```suggestion
    def _on_stream_error(self, error: BaseException) -> None:
        if self._self_output_message is not None:
            self._self_invocation.output_messages = [self._self_output_message]
        self._self_invocation.fail(error)
```

Worth extending that test to assert the partial output survives.

instrumentation/opentelemetry-instrumentation-genai-qwenpaw/pyproject.toml
```diff
@@ -0,0 +1,89 @@
...
+  "opentelemetry-api ~= 1.43",
+  "opentelemetry-instrumentation >= 0.64b0, <1",
+  "opentelemetry-semantic-conventions >= 0.64b0, <1",
+  "opentelemetry-util-genai >= 1.0b0, <2",
+]
+
+[project.optional-dependencies]
+# qwenpaw itself requires Python <3.14, so the marker keeps the extra
+# installable (as a no-op) on newer interpreters used for tooling.
+instruments = ["qwenpaw >= 1.1.0, < 2.0.0; python_version < '3.14'"]
```
lmolkova, line 36: This pins to a line upstream has already replaced. QwenPaw is at 2.1.0 on PyPI, and the [Runtime architecture upgrade](https://github.com/agentscope-ai/QwenPaw/blob/main/website/public/blog/runtime-architecture-upgrade.en.md) refactored away the 650 line `AgentRunner.query_handler` god method this instrumentation patches, so `query_handler` no longer appears in `src/` on main. Worth confirming what the equivalent seam is in 2.x and whether it makes sense to land against 1.1.x only.

instrumentation/opentelemetry-instrumentation-genai-qwenpaw/src/opentelemetry/instrumentation/genai/qwenpaw/patch.py
123liuziming: Good catch — fixed in c955f42. The wrapper now keeps the agent invocation in its own `_self_agent_invocation` field (same as qwen-agent), so the base class's `_self_invocation` stays `None` and the INTERNAL `invoke_agent` span no longer opts into the streamed-call metrics or the half-set `gen_ai.request.stream` state. Dropped the TTFC expectation and the `gen_ai.provider.name` declared violation from the conformance scenario, and the unit test now asserts operation.duration is the only metric emitted.

instrumentation/opentelemetry-instrumentation-genai-qwenpaw/src/opentelemetry/instrumentation/genai/qwenpaw/patch.py
123liuziming: Fixed in c955f42 — `_on_stream_error` now applies the accumulated output message before `fail()`, mirroring the success path (extracted a shared `_apply_output_message` helper). Extended `test_stream_side_error_reraises_and_fails_span` to assert the "partial" output survives on the error span.

instrumentation/opentelemetry-instrumentation-genai-qwenpaw/pyproject.toml
123liuziming: Agreed on landing against 1.1.x only: the `instruments` constraint already caps at `< 2.0.0`, so the instrumentor never patches a 2.x install. In c955f42 I added a "Supported versions" section to the README stating 1.1.x-only support and why (the 2.x runtime architecture upgrade removed `AgentRunner.query_handler`). 2.x support needs to target the new runtime entry point and I'd prefer to do that in a follow-up PR once the 2.x seam settles.

instrumentation/opentelemetry-instrumentation-genai-qwenpaw/src/opentelemetry/instrumentation/genai/qwenpaw/patch.py
123liuziming: Thanks for the thorough analysis — agreed. Since GC finalizes the underlying generator directly, no proxy-based wrapper can intercept that path, so the fix belongs in util-genai's stream wrapper as you filed in #386 (qwen-agent shares the same shape). I'll adopt the util-genai fix here (and add your break-mid-stream test) once it lands; leaving this thread open until then.

123liuziming:
Any more comments on this PR? @lmolkova @eternalcuriouslearner ? Can we just merge it after those conflicts are resolved?

instrumentation/opentelemetry-instrumentation-genai-qwenpaw/src/opentelemetry/instrumentation/genai/qwenpaw/__init__.py
```diff
@@ -0,0 +1,118 @@
...
+                - completion_hook: CompletionHook instance, taking precedence
+                  over the one configured by environment variable
+        """
+        handler = TelemetryHandler(
+            tracer_provider=kwargs.get("tracer_provider"),
+            meter_provider=kwargs.get("meter_provider"),
+            logger_provider=kwargs.get("logger_provider"),
+            completion_hook=kwargs.get("completion_hook")
+            or load_completion_hook(),
+        )
```
lmolkova, line 101: TelemetryHandler requires the package scope name and version, and the package needs a scope test.

```suggestion
        handler = TelemetryHandler(
            tracer_provider=kwargs.get("tracer_provider"),
            meter_provider=kwargs.get("meter_provider"),
            logger_provider=kwargs.get("logger_provider"),
            completion_hook=kwargs.get("completion_hook")
            or load_completion_hook(),
            instrumentation_scope_name=__package__,
            instrumentation_scope_version=__version__,
        )
```

Also add `tests/test_instrumentation_scope.py`:

```python
from opentelemetry.instrumentation.genai.qwenpaw import QwenPawInstrumentor
from opentelemetry.instrumentation.genai.qwenpaw.version import __version__
from opentelemetry.test_util_genai.scope import TelemetryHandlerScopeTest


class TestInstrumentationScope(TelemetryHandlerScopeTest):
    instrumentor_class = QwenPawInstrumentor
    instrumentation_scope_name = "opentelemetry.instrumentation.genai.qwenpaw"
    instrumentation_scope_version = __version__
```

instrumentation/opentelemetry-instrumentation-genai-qwenpaw/src/opentelemetry/instrumentation/genai/qwenpaw/patch.py
```diff
@@ -0,0 +1,167 @@
...
+        # generator ourselves; failing to do so would leave it running.
+        self._finalize_failure(exc_val)
+        try:
+            await self._self_gen.aclose()
+        except Exception:  # pylint: disable=broad-exception-caught
+            _logger.debug(
+                "QwenPaw stream close error after user exception",
+                exc_info=True,
+            )
+        return False
```
lmolkova, line 117: AsyncStreamWrapper already delegates aclose and handles __aexit__. Overriding close leaks an unsupported attribute onto the async generator proxy.

Removing aclose, close, and __aexit__ here relies on AsyncStreamWrapper directly. This test fails today because close is defined:

```python
@pytest.mark.asyncio
async def test_wrapper_does_not_expose_close(runner_module):
    runner = runner_module.AgentRunner(agent_id="entry-agent")
    with patched_command_path(runner_module, fake_run_command_path("ok")):
        stream = runner.query_handler(user_command_msgs(), make_request())
        try:
            # Async generators expose aclose(), not close().
            assert not hasattr(stream, "close")
        finally:
            await stream.aclose()
```

instrumentation/opentelemetry-instrumentation-genai-qwenpaw/src/opentelemetry/instrumentation/genai/qwenpaw/patch.py
```diff
@@ -0,0 +1,167 @@
...
+    agent_name = non_empty_str(getattr(instance, "agent_name", None))
+    invocation = handler.invoke_local_agent(agent_name=agent_name)
+    # The runner's `agent_id` is a local config key (e.g. "default"), not a
+    # provider-assigned stable identifier, so `gen_ai.agent.id` is not
+    # recorded per its semconv guidance.
+    conversation_id = non_empty_str(getattr(request, "session_id", None))
+    if conversation_id is not None:
+        invocation.conversation_id = conversation_id
+    if handler.should_capture_content():
+        invocation.input_messages = input_messages_from_msgs(msgs)
```
lmolkova, line 137: invoke_local_agent() returns LocalAgentInvocation, and content capture can be read directly from invocation.should_capture_content. Also, reading _agent_name avoids eager config file I/O on turns that do not touch agent_name.

```suggestion
) -> LocalAgentInvocation:
    # `agent_name` (config display name with a built-in fallback) was added
    # during the 1.1.x line, so probe for it defensively.
    agent_name = non_empty_str(getattr(instance, "_agent_name", None))
    invocation = handler.invoke_local_agent(agent_name=agent_name)
    # The runner's `agent_id` is a local config key (e.g. "default"), not a
    # provider-assigned stable identifier, so `gen_ai.agent.id` is not
    # recorded per its semconv guidance.
    conversation_id = non_empty_str(getattr(request, "session_id", None))
    if conversation_id is not None:
        invocation.conversation_id = conversation_id
    if invocation.should_capture_content:
        invocation.input_messages = input_messages_from_msgs(msgs)
```

instrumentation/opentelemetry-instrumentation-genai-qwenpaw/src/opentelemetry/instrumentation/genai/qwenpaw/utils.py
```diff
@@ -0,0 +1,83 @@
...
+    if getattr(msg, "role", None) != "assistant":
+        return None
+    text = _msg_text(msg)
+    if text is None:
+        return None
+    return OutputMessage(
+        role="assistant",
+        parts=[Text(content=text)],
+        finish_reason="stop",
+    )
```
lmolkova, line 83: OutputMessage.finish_reason is deprecated in favor of gen_ai.response.finish_reasons on the span (invocation.finish_reasons = ["stop"]), and Text was replaced with TextPart:

```suggestion
    return OutputMessage(
        role="assistant",
        parts=[TextPart(content=text)],
    )
```

In `patch.py`, set finish_reasons on completion:

```python
    def _on_stream_end(self) -> None:
        self._apply_output_message()
        self._self_agent_invocation.finish_reasons = ["stop"]
        self._self_agent_invocation.stop()
```

PR #312, opened 2026-07-26
Add CoPawInstrumentor for QwenPaw's legacy copaw distribution

123liuziming:
# Description

Depends on #311 (stacked on its branch; only the last commit is new).

QwenPaw was originally published as `copaw` (last release `copaw 1.0.2`) with the same `AgentRunner.query_handler` surface. This adds a `CoPawInstrumentor` plugin next to `QwenPawInstrumentor` — a thin subclass pair over a shared private base — so installations still pinned to `copaw` get identical `invoke_agent` telemetry. Like QwenPaw, copaw delegates model and tool execution to AgentScope, so no `chat`/`execute_tool` spans are emitted by this package.

The existing test suite now runs against whichever distribution is installed; a dedicated `py310-test-instrumentation-genai-qwenpaw-copaw` tox factor pins `copaw==1.0.2`.

## Type of change

# How has this been tested?

- [x] `tox -e py310-test-instrumentation-genai-qwenpaw-copaw` (copaw 1.0.2)
- [x] `tox -e py310-test-instrumentation-genai-qwenpaw-oldest` / `py312-…-latest` (unchanged, still green)
- [x] `tox -e py313-test-instrumentation-genai-qwenpaw-conformance` (0 violations)
- [x] `precommit`, `typecheck`, `lint-instrumentation-genai-qwenpaw`, `readme`

# Checklist

- [x] Unit tests added
- [x] Documentation updated

Made with [Cursor](https://cursor.com)


123liuziming:
The failing changelog check is expected while this PR is stacked on #311: the workflow validates every changelog fragment added relative to main, so it sees both 311.added and 312.added and rejects 311.added for not matching this PR's number. Verified locally that both fragments render fine with towncrier (towncrier build --draft).It will resolve automatically once #311 merges and this branch is rebased onto main. I'd rather not drop 311.added from this branch since that would delete #311's fragment at merge time. Alternatively, a maintainer can add the Skip Changelog label while the stack is open.

instrumentation/opentelemetry-instrumentation-genai-qwenpaw/src/opentelemetry/instrumentation/genai/qwenpaw/patch.py
```diff
@@ -0,0 +1,149 @@
...
+            _logger.debug(
+                "QwenPaw stream close error after user exception",
+                exc_info=True,
+            )
+        return False
+
+
+def _build_invocation(
+    handler: TelemetryHandler,
+    instance: Any,
```
lmolkova, line 108: do we know the type here?

instrumentation/opentelemetry-instrumentation-genai-qwenpaw/src/opentelemetry/instrumentation/genai/qwenpaw/patch.py
```diff
@@ -0,0 +1,149 @@
...
+    msgs: Any,
+    request: Any,
+) -> AgentInvocation:
+    # `agent_name` (config display name with a built-in fallback) was added
+    # during the 1.1.x line, so probe for it defensively.
+    agent_name = non_empty_str(getattr(instance, "agent_name", None))
+    invocation = handler.invoke_local_agent(agent_name=agent_name)
+    agent_id = non_empty_str(getattr(instance, "agent_id", None))
```
lmolkova, line 116: is this a server-generated agent id or client-side generated? agent id is expected to be static per agent (and comes from the hosted agent server) https://github.com/open-telemetry/semantic-conventions-genai/blob/main/docs/registry/attributes/gen-ai.md#gen-ai-agent-id

instrumentation/opentelemetry-instrumentation-genai-qwenpaw/tests/conformance/invoke_agent.py
```diff
@@ -0,0 +1,109 @@
...
+        "gen_ai.client.operation.duration",
+        "gen_ai.client.operation.time_to_first_chunk",
+    )
+    # The streamed turn records time_to_first_chunk, whose required
+    # gen_ai.provider.name cannot be set: QwenPaw delegates model calls to
+    # AgentScope, so no provider applies to the invoke_agent invocation.
+    expected_violations = (
+        ExpectedViolation(
+            advice_id="required_attribute_not_present",
+            message_substring="gen_ai.provider.name",
```
lmolkova, line 77: gen_ai.provider.name is no longer required on internal invoke agent spans, I don't think this is necessary anymore

instrumentation/opentelemetry-instrumentation-genai-qwenpaw/tests/conformance/invoke_agent.py
```diff
@@ -0,0 +1,109 @@
...
+            content_capture="SPAN_ONLY",
+        ):
+            with patched_command_path(
+                runner_module, fake_run_command_path("Conformance reply.")
+            ):
+                asyncio.run(_drive_one_turn(runner_module))
+
+
+async def _drive_one_turn(runner_module: Any) -> None:
+    runner = runner_module.AgentRunner(agent_id="conformance-agent")
```
lmolkova, line 105: > agent_id="conformance-agent"

this looks like agent name - do we record as name or id? And if it's id, what goes into the name then?

PR #325, opened 2026-07-28
feat(google-genai): support event-only extra attributes

### LINKED ISSUE #22: [google-genai] Provide a way to attach extra attributes to the operation-details event but not to the span
### What problem do you want to solve?

`opentelemetry-instrumentation-google-genai` currently exposes a single channel
for callers to attach extra attributes to `generate_content` telemetry:
[`GENERATE_CONTENT_EXTRA_ATTRIBUTES_CONTEXT_KEY`][key] (added in #3961, #4103).
Whatever a caller puts there is written to **both**:

- the `generate_content {model}` span (via `span.set_attributes(...)`), and
- the `gen_ai.client.inference.operation.details` log event (via the
  `LogRecord` emitted in `_maybe_log_completion_details`).

There is no way to mark a caller-supplied attribute as event-only.

The existing `OTEL_INSTRUMENTATION_GENAI_CAPTURE_MESSAGE_CONTENT` env var
(modes: `NO_CONTENT` / `SPAN_ONLY` / `EVENT_ONLY` / `SPAN_AND_EVENT`) only gates
the standard semconv message-content fields (`gen_ai.input.messages`,
`gen_ai.output.messages`, etc.) — not arbitrary caller-supplied extras.

This matters when callers want to attach attributes that are acceptable for
audit-style log/event consumers but undesirable on broadly-sampled spans —
for example, an end-user identifier (`user.id`). A concrete instance: the
[Google ADK][adk] passes a `common_attributes` dict containing
`user.id` (the `USER_ID` semconv constant from
`opentelemetry.semconv._incubating.attributes.user_attributes`) through this
context key, which causes `user.id` to land on every `generate_content` span
as soon as `opentelemetry-instrumentation-google-genai` is installed.
ADK currently has no way to keep that value off the span without also losing
it from the operation-details event.

[key]: https://github.com/open-telemetry/opentelemetry-python-contrib/blob/main/instrumentation-genai/opentelemetry-instrumentation-google-genai/src/opentelemetry/instrumentation/google_genai/generate_content.py#L100
[adk]: https://github.com/google/adk-python

### Describe the solution you'd like

Add a sibling context key (`GENERATE_CONTENT_EVENT_ONLY_EXTRA_ATTRIBUTES_CONTEXT_KEY`) that is read by all four `generate_content` wrappers
(sync/async × streaming/non-streaming) but is **only** merged into the
`gen_ai.client.inference.operation.details` `LogRecord`, never into
`span.set_attributes(...)`.

Behavior:

- Keys attached under the new context key appear on the
  `gen_ai.client.inference.operation.details` event log record.
- Those keys never appear on the `generate_content {model}` span.
- On key collisions with the existing `GENERATE_CONTENT_EXTRA_ATTRIBUTES_CONTEXT_KEY`,
  the event-only value wins on the event; the span carries the existing
  extra-attributes value.
- The new constant is re-exported from `opentelemetry.instrumentation.google_genai`
  alongside `GENERATE_CONTENT_EXTRA_ATTRIBUTES_CONTEXT_KEY`.

### Describe alternatives you've considered

1. Have callers strip the attribute before calling the SDK and re-attach it
   to log records via a separate path. This is what ADK is being pushed
   toward today, but it requires duplicating the operation-details event
   construction outside contrib (or post-processing the LogRecord), defeating
   the point of having the contrib instrumentation own the event.
2. A new env var (e.g. `OTEL_INSTRUMENTATION_GENAI_DROP_SPAN_ATTRIBUTES`)
   listing attribute keys to strip from spans. Rejected because it puts the
   policy in deployment config rather than in the code that knows the
   attribute's semantics, and because keys are caller-defined there's no
   useful default list contrib could ship.

### Additional Context

_No response_

Yes

### Tip

tomatotomata:
## Summary

- add a public context key for event-only `generate_content` attributes
- preserve event-only attributes on `gen_ai.client.inference.operation.details` while keeping them off spans
- cover sync/async streaming and non-streaming paths, including collision precedence

Fixes #22

## Validation

- `python -m pytest instrumentation/opentelemetry-instrumentation-google-genai/tests/generate_content/test_sync_nonstreaming.py instrumentation/opentelemetry-instrumentation-google-genai/tests/generate_content/test_sync_streaming.py instrumentation/opentelemetry-instrumentation-google-genai/tests/generate_content/test_async_nonstreaming.py instrumentation/opentelemetry-instrumentation-google-genai/tests/generate_content/test_async_streaming.py -q` � 84 passed, 84 base skips
- `python -m pytest util/opentelemetry-util-genai/tests -q` � 239 passed
- Ruff checks and `git diff --check` pass

### DIFF
--- a/instrumentation/opentelemetry-instrumentation-google-genai/.changelog/22.added
+++ b/instrumentation/opentelemetry-instrumentation-google-genai/.changelog/22.added
@@ -1,0 +1,1 @@
+Add a context key for attaching attributes to `gen_ai.client.inference.operation.details` events without adding them to spans.
--- a/instrumentation/opentelemetry-instrumentation-google-genai/src/opentelemetry/instrumentation/google_genai/__init__.py
+++ b/instrumentation/opentelemetry-instrumentation-google-genai/src/opentelemetry/instrumentation/google_genai/__init__.py
@@ -30,12 +30,16 @@ API
 ---
 """
 
-from .generate_content import GENERATE_CONTENT_EXTRA_ATTRIBUTES_CONTEXT_KEY
+from .generate_content import (
+    GENERATE_CONTENT_EVENT_ONLY_EXTRA_ATTRIBUTES_CONTEXT_KEY,
+    GENERATE_CONTENT_EXTRA_ATTRIBUTES_CONTEXT_KEY,
+)
 from .instrumentor import GoogleGenAiSdkInstrumentor
 from .version import __version__
 
 __all__ = [
     "GoogleGenAiSdkInstrumentor",
     "GENERATE_CONTENT_EXTRA_ATTRIBUTES_CONTEXT_KEY",
+    "GENERATE_CONTENT_EVENT_ONLY_EXTRA_ATTRIBUTES_CONTEXT_KEY",
     "__version__",
 ]
--- a/instrumentation/opentelemetry-instrumentation-google-genai/src/opentelemetry/instrumentation/google_genai/generate_content.py
+++ b/instrumentation/opentelemetry-instrumentation-google-genai/src/opentelemetry/instrumentation/google_genai/generate_content.py
@@ -63,6 +63,11 @@ except ImportError:
 GENERATE_CONTENT_EXTRA_ATTRIBUTES_CONTEXT_KEY = context_api.create_key(
     "generate_content_extra_attributes_context_key"
 )
+GENERATE_CONTENT_EVENT_ONLY_EXTRA_ATTRIBUTES_CONTEXT_KEY = (
+    context_api.create_key(
+        "generate_content_event_only_extra_attributes_context_key"
+    )
+)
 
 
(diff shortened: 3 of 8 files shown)


lmolkova (changes requested):
This seems to be a duplicate of #24

A more generic solution to this problem is coming via https://github.com/open-telemetry/opentelemetry-specification/pull/4931

In the current scope this problem can be solved with custom context key and custom log record processor which would be a recommended approach until context-scoped attributes land in OTel spec and python implementation. 

Is there a strong use-case to stamp information on events only and avoid spans?

tomatotomata:
Closing this PR in response to maintainer guidance: it duplicates #24, and the generic context-scoped solution is being developed in OpenTelemetry Specification PR #4931. The current recommended workaround is a custom context key plus log-record processor, so this branch should not remain open as competing work.

PR #326, opened 2026-07-28
feat(google-genai): capture embedding config attributes

### LINKED ISSUE #170: Coverage gaps in `opentelemetry-instrumentation-google-genai` instrumentations
Using this to track some gaps as I go about monkey patching various APIs as part of the openinference donation stuff..

Gaps:
A generic way for the user to provide attributes that make it into the span.. We use a `ContextVar` to do this for the `GenerateContent` method, and could do the same for the other APIs..

Embeddings:

- We could allow capturing various `EmbedContentConfig` in attributes the same way we do `GenerateContentConfig` in the `generate_content` instrumentation.
- Openinference captures the [actual raw embeddings in their instrumentation](https://github.com/open-telemetry/donation-openinference/blob/6cdd644d79fccf50aedcb614187f924ddfcafb7b/python/instrumentation/openinference-instrumentation-google-genai/src/openinference/instrumentation/google_genai/_embedding_attributes_extractor.py#L1) -- that seems too big to put into an attribute..
- https://github.com/googleapis/python-genai/issues/2658 -- `EmbedContentResponse` has a bug that doesn't let us get tokens from `usage_metadata`

Interactions:

- Need to capture interaction ID:  https://github.com/open-telemetry/semantic-conventions-genai/pull/372
- Expand input / output messages to capture all inputs / outputs -- the SDK accepts so many different types (related sem conv issue: https://github.com/open-telemetry/semantic-conventions-genai/issues/380)

tomatotomata:
## Summary

- capture non-sensitive `EmbedContentConfig` fields on Google GenAI embedding spans
- support both `EmbedContentConfig` objects and dict configurations for sync and async calls
- omit `http_options` so headers and request settings are not copied into telemetry

Fixes #170

## Validation

- `python -m pytest instrumentation/opentelemetry-instrumentation-google-genai/tests/embeddings/test_embeddings.py -q` � 6 passed
- Full package run: 193 passed, 116 skipped; 8 credential-dependent setup errors and 1 async-plugin failure are pre-existing environment limitations
- Ruff and `git diff --check` pass

### DIFF
--- a/instrumentation/opentelemetry-instrumentation-google-genai/.changelog/170.added
+++ b/instrumentation/opentelemetry-instrumentation-google-genai/.changelog/170.added
@@ -1,1 +1,1 @@
-Capture non-sensitive `EmbedContentConfig` fields on Google GenAI embedding spans.
+Capture non-sensitive ``EmbedContentConfig`` fields on Google GenAI embedding spans.
--- a/instrumentation/opentelemetry-instrumentation-google-genai/.changelog/170.added
+++ b/instrumentation/opentelemetry-instrumentation-google-genai/.changelog/170.added
@@ -1,0 +1,1 @@
+Capture non-sensitive `EmbedContentConfig` fields on Google GenAI embedding spans.
--- a/instrumentation/opentelemetry-instrumentation-google-genai/src/opentelemetry/instrumentation/google_genai/embeddings.py
+++ b/instrumentation/opentelemetry-instrumentation-google-genai/src/opentelemetry/instrumentation/google_genai/embeddings.py
@@ -24,6 +24,7 @@ from opentelemetry.util.genai.invocation import (
     EmbeddingInvocation,
 )
 
+from .allowlist_util import AllowList
 from .custom_semconv import GCP_GENAI_OPERATION_CONFIG
 from .dict_util import flatten_dict
 
@@ -94,6 +95,7 @@ def _apply_embedding_response_attributes(
 
 def _apply_embedding_request_attributes(
     config: EmbedContentConfig | dict[str, Any] | None,
+    allow_list: AllowList,
     invocation: EmbeddingInvocation,
 ) -> None:
     if config is None:
@@ -111,11 +113,18 @@ def _apply_embedding_request_attributes(
         # telemetry attributes.
         exclude_keys={f"{GCP_GENAI_OPERATION_CONFIG}.http_options"},
     )
-    invocation.attributes.update(attributes)
+    invocation.attributes.update(
+        {
+            key: value
+            for key, value in attributes.items()
+            if allow_list.allowed(key)
+        }
+    )
 
 
 def _create_instrumented_embed_content(
     telemetry_handler: TelemetryHandler,
+    embed_content_config_key_allowlist: AllowList,
 ) -> Callable[
     [
         Callable[..., EmbedContentResponse],
@@ -142,7 +151,9 @@ def _create_instrumented_embed_content(
             server_address=server_address,
         ) as invocation:
             _apply_embedding_request_attributes(
-                kwargs.get("config"), invocation
+                kwargs.get("config"),
+                embed_content_config_key_allowlist,
+                invocation,
             )
             response = wrapped(*args, **kwargs)
             _apply_embedding_response_attributes(response, invocation)
@@ -154,6 +165,7 @@ def _create_instrumented_embed_content(
 
 def _create_instrumented_async_embed_content(
     telemetry_handler: TelemetryHandler,
+    embed_content_config_key_allowlist: AllowList,
 ) -> Callable[
     [
         Callable[..., Any],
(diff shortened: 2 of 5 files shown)


instrumentation/opentelemetry-instrumentation-google-genai/src/opentelemetry/instrumentation/google_genai/embeddings.py
```diff
@@ -89,6 +92,28 @@ def _apply_embedding_response_attributes(
             pass
 
 
+def _apply_embedding_request_attributes(
```
DylanRussell, line 96: For `GenerateContentConfig` we made it default off // opt-in whether to  include it..

Check out https://github.com/open-telemetry/opentelemetry-python-genai/blob/3b86f51593cf1a58169d6d76d08865c7082f2f44/instrumentation/opentelemetry-instrumentation-google-genai/README.rst#configuration-recording and https://github.com/open-telemetry/opentelemetry-python-genai/blob/main/instrumentation/opentelemetry-instrumentation-google-genai/src/opentelemetry/instrumentation/google_genai/generate_content.py#L306 -- EmbedContentConfig is smaller so maybe it's fine to just dump it though.. probably most of the time just 1 or 2 fields will be set on it..

tomatotomata: You are right: issue #170 explicitly calls for EmbedContentConfig to follow GenerateContentConfig, so treating small config values as safe by default was inconsistent. e258f09 now defaults embedding-config capture off and adds the matching `OTEL_GOOGLE_GENAI_EMBED_CONTENT_CONFIG_INCLUDES` / `..._EXCLUDES` controls. I also added coverage for default omission and explicit opt-in; the focused embedding and instrumentor tests pass (7 tests).

instrumentation/opentelemetry-instrumentation-google-genai/src/opentelemetry/instrumentation/google_genai/embeddings.py
```diff
@@ -89,8 +93,38 @@ def _apply_embedding_response_attributes(
...
+        except Exception:
+            return
+
+    attributes = flatten_dict(
+        config.model_dump(exclude_none=True),
+        key_prefix=GCP_GENAI_OPERATION_CONFIG,
+        # HTTP options can contain request headers and must not be copied into
+        # telemetry attributes.
+        exclude_keys={f"{GCP_GENAI_OPERATION_CONFIG}.http_options"},
+    )
```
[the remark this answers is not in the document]
tomatotomata: Addressed in d1915c3. The focused embedding tests now cover synchronous capture from an `EmbedContentConfig` instance and assert that no `http_options` attribute is emitted even with an unrestricted allowlist. The existing async dict-config test remains in place.

instrumentation/opentelemetry-instrumentation-google-genai/.changelog/170.added
```diff
@@ -0,0 +1 @@
+Capture non-sensitive `EmbedContentConfig` fields on Google GenAI embedding spans.
```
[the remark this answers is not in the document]
tomatotomata: Corrected in d1915c3 to use reStructuredText inline-literal markup.

instrumentation/opentelemetry-instrumentation-google-genai/src/opentelemetry/instrumentation/google_genai/embeddings.py
```diff
@@ -89,8 +93,38 @@ def _apply_embedding_response_attributes(
...
+        return
+    if isinstance(config, dict):
+        try:
+            config = EmbedContentConfig.model_validate(config)
+        except Exception:
+            return
+
+    attributes = flatten_dict(
+        config.model_dump(exclude_none=True),
+        key_prefix=GCP_GENAI_OPERATION_CONFIG,
```
aabmass, line 111: Can this collide with the other operation options?

tomatotomata:
I checked this against the current instrumentation paths. The embedding config keys use the shared `gcp.gen_ai.operation.config.*` namespace, which is also used for generation config, but the two operations create separate spans, so one operation cannot overwrite the other operation's attributes. The allowlist is applied to the flattened, prefixed keys on each span, and `http_options` stays excluded. I kept the shared namespace intentionally so the config fields have the same shape across operations. Let me know if you would prefer an operation-specific prefix instead.

instrumentation/opentelemetry-instrumentation-google-genai/src/opentelemetry/instrumentation/google_genai/embeddings.py
tomatotomata: I checked the span lifecycle: embeddings and generations create separate spans, so the shared prefixed keys stay scoped to their own operation. The include/exclude filter is applied to the flattened keys on that span, and `http_options` remains excluded. I kept the shared namespace so the config keys stay consistent across both operations.

PR #327, opened 2026-07-28
test(openai): enforce experimental embedding metric conformance

### LINKED ISSUE #38: OpenAI instrumentation: embedding instrumentation metrics conformance
It seems openai instrumentation emits legacy metrics when in experimental mode

```
opentelemetry.test.weaver_live_check.LiveCheckError: Semconv violations found:
  - [deprecated] Attribute 'gen_ai.system' is deprecated; reason = 'renamed', note = 'Replaced by `gen_ai.provider.name`.'. (1 occurrence(s) on metric 'gen_ai.client.operation.duration')
  - [deprecated] Attribute 'gen_ai.system' is deprecated; reason = 'renamed', note = 'Replaced by `gen_ai.provider.name`.'. (1 occurrence(s) on metric 'gen_ai.client.token.usage')
  - [required_attribute_not_present] Required attribute 'gen_ai.provider.name' is not present. (1 occurrence(s) on metric 'gen_ai.client.operation.duration')
  - [required_attribute_not_present] Required attribute 'gen_ai.provider.name' is not present. (1 occurrence(s) on metric 'gen_ai.client.token.usage')
```

To reproduce: unskip corresponding scenario in conformance testing: https://github.com/open-telemetry/opentelemetry-python-genai/pull/14

tomatotomata:
## Summary

- run the OpenAI embedding conformance scenario explicitly in latest experimental semantic-convention mode
- add a regression test that requires `gen_ai.provider.name` on both embedding metrics

Fixes #38

## Validation

- `python -m pytest instrumentation/opentelemetry-instrumentation-genai-openai/tests/test_embedding_invocation_unit.py -q` � 11 passed
- Ruff and `git diff --check` pass
- The live Weaver conformance scenario is configured but cannot run locally because the Weaver binary is not installed; CI conformance will execute it.

### DIFF
--- a/instrumentation/opentelemetry-instrumentation-genai-openai/tests/conformance/embedding.py
+++ b/instrumentation/opentelemetry-instrumentation-genai-openai/tests/conformance/embedding.py
@@ -38,9 +38,6 @@ class EmbeddingScenario(Scenario):
             logger_provider=logger_provider,
             meter_provider=meter_provider,
             content_capture="SPAN_ONLY",
-            extra_env={
-                "OTEL_SEMCONV_STABILITY_OPT_IN": "gen_ai_latest_experimental"
-            },
         ):
             with vcr.use_cassette("embedding_conformance.yaml"):
                 OpenAI().embeddings.create(
--- a/instrumentation/opentelemetry-instrumentation-genai-openai/tests/conformance/embedding.py
+++ b/instrumentation/opentelemetry-instrumentation-genai-openai/tests/conformance/embedding.py
@@ -38,6 +38,9 @@ class EmbeddingScenario(Scenario):
             logger_provider=logger_provider,
             meter_provider=meter_provider,
             content_capture="SPAN_ONLY",
+            extra_env={
+                "OTEL_SEMCONV_STABILITY_OPT_IN": "gen_ai_latest_experimental"
+            },
         ):
             with vcr.use_cassette("embedding_conformance.yaml"):
                 OpenAI().embeddings.create(
--- a/instrumentation/opentelemetry-instrumentation-genai-openai/tests/test_embedding_invocation_unit.py
+++ b/instrumentation/opentelemetry-instrumentation-genai-openai/tests/test_embedding_invocation_unit.py
@@ -140,40 +140,6 @@ def test_dimensions_omitted_when_not_provided(handler):
         invocation.stop()
 
 
-def test_embedding_metrics_use_provider_name_attribute(handler, metric_reader):
-    """Experimental metrics must use the current provider attribute."""
-    invocation = create_embedding_invocation(
-        handler, {"model": "text-embedding-3-small"}, _make_client()
-    )
-    invocation.input_tokens = 7
-    invocation.stop()
-
-    metrics = metric_reader.get_metrics_data()
-    metric_names = {
-        metric.name
-        for resource_metric in metrics.resource_metrics
-        for scope_metric in resource_metric.scope_metrics
-        for metric in scope_metric.metrics
-    }
-    assert {
-        "gen_ai.client.operation.duration",
-        "gen_ai.client.token.usage",
-    } <= metric_names
-
-    for resource_metric in metrics.resource_metrics:
-        for scope_metric in resource_metric.scope_metrics:
-            for metric in scope_metric.metrics:
-                if metric.name in {
-                    "gen_ai.client.operation.duration",
-                    "gen_ai.client.token.usage",
-                }:
-                    for point in metric.data.data_points:
-                        assert (
-                            point.attributes["gen_ai.provider.name"]
-                            == "openai"
-                        )
-
-
 def test_encoding_format_mapped_to_invocation(handler):
     invocation = create_embedding_invocation(
         handler,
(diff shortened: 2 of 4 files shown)


instrumentation/opentelemetry-instrumentation-genai-openai/tests/conformance/embedding.py
```diff
@@ -38,6 +38,9 @@ def run(
             logger_provider=logger_provider,
             meter_provider=meter_provider,
             content_capture="SPAN_ONLY",
+            extra_env={
```
lmolkova: this should not be necessary in this repo - we should have removed this flag here. If there are some leftovers that require it, could you please send a PR to switch to latest experimental by default and unconditionally ?

instrumentation/opentelemetry-instrumentation-genai-openai/tests/test_embedding_invocation_unit.py
```diff
@@ -140,6 +140,40 @@ def test_dimensions_omitted_when_not_provided(handler):
         invocation.stop()
 
 
+def test_embedding_metrics_use_provider_name_attribute(handler, metric_reader):
```
lmolkova: this is already covered in https://github.com/open-telemetry/opentelemetry-python-genai/blob/e92e604fef191c9df7a4568ba4deb03ab34206cc/util/opentelemetry-util-genai/tests/test_handler_metrics.py#L266

if the goal is to test all openai instrumentation e2e consider adding missing attributes in existing tests https://github.com/ahmadalguydi/opentelemetry-python-genai/blob/77dbb757b64025a90db5841eb6e79e8513c19be9/instrumentation/opentelemetry-instrumentation-genai-openai/tests/test_embeddings.py

tomatotomata: Moved the regression out of the unit test and into `test_embeddings_token_metrics` in `test_embeddings.py`, so it covers the end-to-end OpenAI instrumentation path. The test now checks both emitted metrics and their data points, including the provider attribute and absence of the deprecated system key.

instrumentation/opentelemetry-instrumentation-genai-openai/tests/test_embedding_invocation_unit.py
```diff
@@ -140,6 +140,40 @@ def test_dimensions_omitted_when_not_provided(handler):
...
+                if metric.name in {
+                    "gen_ai.client.operation.duration",
+                    "gen_ai.client.token.usage",
+                }:
+                    for point in metric.data.data_points:
+                        assert (
+                            point.attributes["gen_ai.provider.name"]
+                            == "openai"
+                        )
+
```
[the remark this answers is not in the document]
tomatotomata: Implemented in `8895d2f`: the regression now lives in the existing embedding integration test, selects the duration and token metrics explicitly, requires data points, uses `GenAIAttributes` constants/enums, asserts `gen_ai.provider.name == OPENAI`, and verifies the deprecated `gen_ai.system` attribute is absent.

instrumentation/opentelemetry-instrumentation-genai-openai/tests/conformance/embedding.py
tomatotomata: The conformance scenario no longer passes an experimental-mode flag; commit `c220d9d` removed that plumbing so it now exercises the repository default unconditionally.

instrumentation/opentelemetry-instrumentation-genai-openai/tests/test_embeddings.py
```diff
@@ -377,6 +377,15 @@ def test_embeddings_token_metrics(
...
     assert token_metric is not None
 
+    for metric in (duration_metric, token_metric):
+        assert metric.data.data_points, f"{metric.name} metric has no data points"
+        for point in metric.data.data_points:
+            assert (
+                point.attributes[GenAIAttributes.GEN_AI_PROVIDER_NAME]
+                == GenAIAttributes.GenAiProviderNameValues.OPENAI.value
+            )
+            assert GenAIAttributes.GEN_AI_SYSTEM not in point.attributes
```
lmolkova: there is no need to test it. the original issue was fixed and conformance test would flag usage of deprecated attributes


```suggestion
```

tomatotomata: Removed the redundant assertions in 488c735. The existing conformance coverage remains the guard for deprecated attributes. The affected `test_embeddings_token_metrics` test passes locally, along with Ruff, formatting, and `git diff --check`.

lmolkova (changes requested):
this PR is empty

tomatotomata:
I checked the current base comparison before replying. The implementation commits are already present on `main`, so this PR now contains only the changelog fragment. I was thinking the cleanest path is to treat it as superseded rather than add another no-op commit. If the fragment is still wanted separately, I can keep it here; otherwise I’ll close this stale remainder.

lmolkova:
thankd for the contribution @tomatotomata , I'm going to close this RR since original is resolved and this PR now consist of the changelog only.

PR #328, opened 2026-07-28
[`opentelemetry-instrumentation-genai-agno`] Add instrumentation for `agno.agent.run`/`arun` and `agno.tools.function.execute`/`aexecute`

DylanRussell:
# Description

Instruments agno's agent run (creates a client side invoke agent span) and tool execution (creates an execute tool span) functions.

Adds tool definition and metric count sem conv attributes, and various other attributes

I also added a way in the `/dev` folder to manually test out the instrumentation against the live code / model (along with google genai) and capture all the spans / logs locally, so you can manually review what the instrumentation actually generates..

Also adds weaver conformance tests..

## Type of change

- [x ] New feature (non-breaking change which adds functionality)

# How has this been tested?

Unit tests. Also manually tried out and looked at the spans / logs that are generated..

# Checklist

- [x] Unit tests added
- [x] Documentation updated


tox.ini
```diff
@@ -33,6 +33,7 @@ envlist =
     ; instrumentation-genai-agno
     py3{10,11,12,13,14}-test-instrumentation-genai-agno-latest
     py310-test-instrumentation-genai-agno-oldest
+    py314-test-instrumentation-genai-agno-conformance
```
lmolkova, line 36: there is something in release workflows that lists releasable instrumentations, since this one becomes functional, we should probably list it there too

instrumentation/opentelemetry-instrumentation-genai-agno/tests/conformance/tool.py
```diff
@@ -0,0 +1,52 @@
...
+
+from opentelemetry.instrumentation.genai.agno import AgnoInstrumentor
+from opentelemetry.sdk._logs import LoggerProvider
+from opentelemetry.sdk.metrics import MeterProvider
+from opentelemetry.sdk.trace import TracerProvider
+from opentelemetry.test_util_genai.conformance import Scenario
+from opentelemetry.test_util_genai.instrumentor import instrument
+
+
+class ToolScenario(Scenario):
```
lmolkova: can we merge it into agent scenario? Conformance tests are more like integration tests, so it's ok to have larger scenarios there and then we can validate a natural flow rather than synthetic one

instrumentation/opentelemetry-instrumentation-genai-agno/src/opentelemetry/instrumentation/genai/agno/patch.py
```diff
@@ -0,0 +1,303 @@
...
+    _set_invocation_input(invocation, instance, args, kwargs, capture_content)
+    invocation.tool_definitions = prepare_tool_definitions(
+        getattr(instance, "tools", None)
+    )
+    return invocation
+
+
+def _start_tool_invocation(
+    handler: TelemetryHandler,
+    instance: Any,
```
lmolkova, line 193: does it need to be any? it's always `ToolInvocation`, no?

instrumentation/opentelemetry-instrumentation-genai-agno/src/opentelemetry/instrumentation/genai/agno/patch.py
```diff
@@ -0,0 +1,303 @@
...
+        if (
+            hasattr(metrics, "output_tokens")
+            and getattr(metrics, "output_tokens") is not None
+        ):
+            invocation.output_tokens = int(getattr(metrics, "output_tokens"))
+
+
+def _start_agent_invocation(
+    handler: TelemetryHandler,
+    instance: Any,
```
lmolkova, line 177: it can be properly typed I think

DylanRussell:
> According to the official OpenTelemetry GenAI Semantic Conventions (0.65b0), gen_ai.tool.definitions is only defined on inference/chat spans (chat, text_completion), not on agent invocation (invoke_agent) spans.

This seems wrong -- why can't tool definitions be on invoke agent spans ? You can define tools on an agent

tox.ini
DylanRussell: Nice, added it in

instrumentation/opentelemetry-instrumentation-genai-agno/tests/conformance/tool.py
DylanRussell: Yeah SG merged them

instrumentation/opentelemetry-instrumentation-genai-agno/src/opentelemetry/instrumentation/genai/agno/patch.py
DylanRussell: Yeah good point.. fixed

lmolkova:
> > According to the official OpenTelemetry GenAI Semantic Conventions (0.65b0), gen_ai.tool.definitions is only defined on inference/chat spans (chat, text_completion), not on agent invocation (invoke_agent) spans.
> 
> This seems wrong -- why can't tool definitions be on invoke agent spans ? You can define tools on an agent

where did it come from? let's fix, I agree it's wrong

DylanRussell:
> > > According to the official OpenTelemetry GenAI Semantic Conventions (0.65b0), gen_ai.tool.definitions is only defined on inference/chat spans (chat, text_completion), not on agent invocation (invoke_agent) spans.
> > 
> > 
> > This seems wrong -- why can't tool definitions be on invoke agent spans ? You can define tools on an agent
> 
> where did it come from? let's fix, I agree it's wrong

nevermind, gen ai hallucination when debugging the failed weaver live check.. nothing needs fixing

instrumentation/opentelemetry-instrumentation-genai-agno/src/opentelemetry/instrumentation/genai/agno/patch.py
```diff
@@ -0,0 +1,295 @@
...
+            pass
+    return str(cast(object, args_val))
+
+
+def _set_tool_invocation_input(
+    invocation: ToolInvocation,
+    instance: Any,
+    capture_content: bool,
+) -> None:
+    if getattr(invocation, "should_capture_content_on_span", False):
```
lmolkova: why not 

```suggestion
    if capture_content:
```

?

instrumentation/opentelemetry-instrumentation-genai-agno/src/opentelemetry/instrumentation/genai/agno/patch.py
```diff
@@ -0,0 +1,295 @@
...
+            invocation.arguments = _extract_arguments_str(arguments)
+
+
+def _set_tool_invocation_output(
+    invocation: Any,
+    result: Any,
+    capture_content: bool,
+) -> None:
+    if (
+        getattr(invocation, "should_capture_content_on_span", False)
```
lmolkova:
```suggestion
        capture_content
```

lmolkova (approved):

instrumentation/opentelemetry-instrumentation-genai-agno/src/opentelemetry/instrumentation/genai/agno/patch.py
DylanRussell: Yep that works. done

PR #350, opened 2026-08-03
[`opentelemetry-instrumentation-genai-agno`] Add instrumentation for Agno Team, Workflow methods.

DylanRussell:
# Description

Add instrumentation for the following methods:

Agno Class & Method(s) | SemConv Concept / Kind
-- | --
Team.runTeam.arun | Agent (gen_ai.span.kind = AGENT)
Workflow.runWorkflow.arun | Workflow (gen_ai.span.kind = WORKFLOW)

I'm not certain if it makes sense to model `teams` function call as a `invoke agent` span or if we might want to add a new sem conv span name for it.. But that is how both open inference and  alibaba longsuite modeled it and it sort of makes sense. https://docs.agno.com/teams/overview#when-to-use-teams and https://docs.agno.com/workflows/overview#when-to-use-workflows describe the differences between workflows / agents / teams..

Next steps:

Add run_stream / arun_stream / response_stream / aresponse_stream using SyncStreamWrapper and AsyncStreamWrapper.
Add Parallel.execute and Parallel.aexecute wrapping for workflow parallel steps.
Instrument `Step` (awaiting on https://github.com/open-telemetry/semantic-conventions-genai/pull/188)
Instrument model invocations.. need to decide how to disable/prevent duplicate telemetry in this scenario..

## Type of change

# How has this been tested?

Unit tests

# Checklist
- [x] Unit tests added
- [x] Documentation updated


instrumentation/opentelemetry-instrumentation-genai-agno/src/opentelemetry/instrumentation/genai/agno/patch.py
```diff
@@ -135,14 +226,16 @@ def _set_tool_invocation_output(
...
-    invocation: AgentInvocation,
+    invocation: Any,
     instance: Any,
     args: tuple[Any, ...],
     kwargs: dict[str, Any],
     capture_content: bool,
 ) -> None:
-    agent_id = getattr(instance, "agent_id", None)
-    if agent_id:
+    agent_id = getattr(instance, "agent_id", None) or getattr(
```
lmolkova: sorry I didn't ask before, but is this remote agent id or a local instance identifier? In semconv we decided not to record in-proc instance id on this attribute - https://github.com/open-telemetry/semantic-conventions-genai/pull/242

instrumentation/opentelemetry-instrumentation-genai-agno/src/opentelemetry/instrumentation/genai/agno/patch.py
```diff
@@ -179,7 +272,8 @@ def _start_agent_invocation(
     kwargs: dict[str, Any],
     capture_content: bool,
 ) -> AgentInvocation:
-    agent_name = getattr(instance, "name", None) or "Agent"
+    default_name = "Team" if instance.__class__.__name__ == "Team" else "Agent"
+    agent_name = getattr(instance, "name", None) or default_name
```
lmolkova: do we need a static name? it shouldn't be required, I'd prefer it not to have name rather than using a static name. this one is proposing explicit guidance to avoid hardcoded name fallbacks for workflows - https://github.com/open-telemetry/semantic-conventions-genai/pull/354

instrumentation/opentelemetry-instrumentation-genai-agno/src/opentelemetry/instrumentation/genai/agno/patch.py
```diff
@@ -290,3 +384,209 @@ async def traced_method(
...
+def _start_workflow_invocation(
+    handler: TelemetryHandler,
+    instance: Any,
+    args: tuple[Any, ...],
+    kwargs: dict[str, Any],
+    capture_content: bool,
+    *,
+    default_name: str = "Workflow",
+) -> WorkflowInvocation:
+    workflow_name = getattr(instance, "name", None) or default_name
```
lmolkova: same (and drop default as a param)

```suggestion
    workflow_name = getattr(instance, "name", None)
```

instrumentation/opentelemetry-instrumentation-genai-agno/src/opentelemetry/instrumentation/genai/agno/patch.py
```diff
@@ -290,3 +384,209 @@ async def traced_method(
...
+                    role=str(getattr(result, "role", "assistant")),
+                    parts=[Text(content=output_str)],
+                    finish_reason=str(
+                        getattr(result, "finish_reason", "stop")
+                    ),
+                )
+            ]
+
+
+def _workflow_run(
```
lmolkova, line 366: can you please add conformance scenario for workflows ?

instrumentation/opentelemetry-instrumentation-genai-agno/src/opentelemetry/instrumentation/genai/agno/patch.py
```diff
@@ -290,3 +384,209 @@ async def traced_method(
...
+) -> Callable[..., Any]:
+    capture_content = handler.should_capture_content()
+
+    def traced_method(
+        wrapped: Callable[..., Any],
+        instance: Any,
+        args: tuple[Any, ...],
+        kwargs: dict[str, Any],
+    ) -> Any:
+        with _start_workflow_invocation(
```
lmolkova: is step a workflow? 

@RKest is defining a node / step span here https://github.com/open-telemetry/semantic-conventions-genai/pull/188 and we're bikeshedding on terminology.

JFYI: the ultimate terminology bikeshed - https://github.com/open-telemetry/semantic-conventions-genai/issues/418

instrumentation/opentelemetry-instrumentation-genai-agno/src/opentelemetry/instrumentation/genai/agno/patch.py
DylanRussell: local instance identifier. Alright SG I've removed it

instrumentation/opentelemetry-instrumentation-genai-agno/src/opentelemetry/instrumentation/genai/agno/patch.py
DylanRussell: Ack thats fine, no need for a static name, updated it to default to None

instrumentation/opentelemetry-instrumentation-genai-agno/src/opentelemetry/instrumentation/genai/agno/patch.py
DylanRussell: yep, done

instrumentation/opentelemetry-instrumentation-genai-agno/src/opentelemetry/instrumentation/genai/agno/patch.py
DylanRussell: step is a node in a workflow (see https://docs.agno.com/workflows/workflow-patterns/step-based-workflow and https://docs.agno.com/reference/workflows/step) -- so it sounds like I should hold off on it for now.. I've reverted monkey patching of STEP in this PR

instrumentation/opentelemetry-instrumentation-genai-agno/src/opentelemetry/instrumentation/genai/agno/patch.py
DylanRussell: FYI this turned up that the duration metric was missing from GenAi utils workflow code (which I've added in  this PR now)

Other workflow conformance scenarios didn't fail previously because the child operations (chat or invoke_agent) inside the workflows emitted gen_ai.client.operation.duration.

util/opentelemetry-util-genai/src/opentelemetry/util/genai/_workflow_invocation.py
```diff
@@ -97,4 +106,4 @@ def _apply_finish(self, error: Error | None = None) -> None:
             inputs=self.input_messages,
             outputs=self.output_messages,
         )
-        # TODO: Add workflow metrics when supported
+        self._metrics_recorder.record(self)
```
lmolkova: we have a special metric for workflow - https://github.com/open-telemetry/semantic-conventions-genai/blob/main/docs/gen-ai/gen-ai-metrics.md#metric-gen_aiinvoke_workflowduration

this one i think records the client duration

instrumentation/opentelemetry-instrumentation-genai-agno/src/opentelemetry/instrumentation/genai/agno/patch.py
```diff
@@ -135,16 +204,12 @@ def _set_tool_invocation_output(
 
 
 def _set_invocation_input(
-    invocation: AgentInvocation,
+    invocation: Any,
```
lmolkova: why? should it be workflow invocation instead?

util/opentelemetry-util-genai/.changelog/350.fixed
```diff
@@ -0,0 +1 @@
+Record `gen_ai.client.operation.duration` metric on workflow invocations (`WorkflowInvocation`).
```
lmolkova: mind updating conformance scenarios for other libs doing workflows and adding `gen_ai.invoke_workflow.duration` as expected metric?

lmolkova (commented):
LGTM, but the metric should be `gen_ai.invoke_workflow.duration`

util/opentelemetry-util-genai/.changelog/350.fixed
DylanRussell: Yes, ack done

util/opentelemetry-util-genai/src/opentelemetry/util/genai/_workflow_invocation.py
DylanRussell: Woops.. done

instrumentation/opentelemetry-instrumentation-genai-agno/src/opentelemetry/instrumentation/genai/agno/patch.py
DylanRussell: yes Agent or Workflow invocation.. fixed thanks

instrumentation/opentelemetry-instrumentation-genai-agno/src/opentelemetry/instrumentation/genai/agno/patch.py
```diff
@@ -291,3 +360,155 @@ async def traced_method(
...
+    kwargs: dict[str, Any],
+    capture_content: bool,
+) -> WorkflowInvocation:
+    workflow_name = getattr(instance, "name", None)
+    invocation = handler.workflow(name=workflow_name)
+    _set_invocation_input(invocation, instance, args, kwargs, capture_content)
+    return invocation
+
+
+def _start_model_invocation(
```
lmolkova: do we need to report inference from agno? does it rely on openai / anthropic / google genai to make model calls? if so, why not rely on the corresponding instrumentations to do it? 

Otherwise we end up with duplication if inference is recorded by both layers.

util/opentelemetry-test-util-genai/src/opentelemetry/test_util_genai/conformance.py
```diff
@@ -72,9 +72,7 @@ class ExpectedViolation:
     def matches(self, violation: dict[str, Any]) -> bool:
         return violation.get(
             "id"
-        ) == self.advice_id and self.message_substring in str(
-            violation.get("message", "")
-        )
+        ) == self.advice_id and self.message_substring in str(violation)
```
lmolkova: do we need to relax? I'd rather make it more strict

util/opentelemetry-test-util-genai/src/opentelemetry/test_util_genai/_setup_weaver.py
```diff
@@ -207,25 +207,22 @@ def _provision_genai_root() -> Path:
     )
 
     upstream_pins = _load_version_pins(genai_target / "versions.env")
-    try:
-        upstream_version = upstream_pins["SEMCONV_VERSION"]
-    except KeyError as missing:
-        raise RuntimeError(
-            f"genai repo's versions.env is missing {missing!s}"
-        ) from missing
+    upstream_version = upstream_pins.get("SEMCONV_VERSION")
```
lmolkova: intentional?

instrumentation/opentelemetry-instrumentation-genai-agno/src/opentelemetry/instrumentation/genai/agno/patch.py
DylanRussell: They support a ton of models including ones we don't have instrumenation for and stuff like vLLM where you can bring up your own model (https://docs.agno.com/examples/models/overview) ..

If we instrument at the agno layer the invoke model span will look the exact same for every model call, versus it will look slightly different when generated at the model layer (because the model API interfaces differ, and the instrumentations differ slightly).. 

It's easier to just install agno instrumentation, than install agno + X model provider instrumentations (probably not a big deal but still..)

ADK does the same thing FWIW..

Yeah duplication is a problem though.. I think probably the best way to handle it is for the agno instrumentation to NOT instrument the model call IF the models associated instrumentation is installed and the model has been monkey patched.. Before thinking about how to do that though, what do you think of my argument ?

util/opentelemetry-test-util-genai/src/opentelemetry/test_util_genai/conformance.py
DylanRussell: reverted this change

util/opentelemetry-test-util-genai/src/opentelemetry/test_util_genai/_setup_weaver.py
DylanRussell: Yes because as of https://github.com/open-telemetry/semantic-conventions-genai/pull/433/ we no longer specify `SEMCONV_VERSION` in `versions.env`.. instead it's directly specified in the manifest (not sure what this means or how that works exatly) but i'm pretty sure we can get rid of this workaround completely now.. PTAL though and verify what I did

instrumentation/opentelemetry-instrumentation-genai-agno/src/opentelemetry/instrumentation/genai/agno/patch.py
lmolkova: I think it makes sense to instrument layers that can't be instrumented otherwise. 
Priority-wise - we deliver more value and coverage by instrumenting layers that uniquely belong to this library.

I propose to postpone inference instrumentation in agentic libs until we have a reasonable coverage for agentic layers and get back to it later when polishing onboarding experience or covering less popular inference providers.

DylanRussell: I'm going to instrument all of `agno`, I don't think we have to worry about prioritization and ranking value of particular parts of this library.. I think it makes sense to instrument this layer and address the problem now.. But I'll instrument the rest of the library first, I don't have a problem doing that either..

DylanRussell: Reverted the model inference instrumentation for now.. PTAL

instrumentation/opentelemetry-instrumentation-genai-agno/src/opentelemetry/instrumentation/genai/agno/patch.py
```diff
@@ -156,20 +198,24 @@ def _set_invocation_input(
...
 ) -> None:
     if capture_content and result is not None:
         output_str = _extract_output_content(result)
         invocation.output_messages = [
             OutputMessage(
-                role="assistant",
+                role=str(getattr(result, "role", "assistant")),
                 parts=[Text(content=output_str)],
-                finish_reason="stop",
+                finish_reason=str(getattr(result, "finish_reason", "stop")),
```
lmolkova: `role` and `finish_reason` don't exist on `RunOutput`, `TeamRunOutput` or `WorkflowRunOutput` (agno 2.8.6), so both `getattr`s always take the default. And if either field does appear later as `None`, `str()` turns it into the literal string `"None"`:

```python
@dataclass
class _Result:
    content: str = "hi"
    role: str | None = None
    finish_reason: str | None = None

def test_none_role_becomes_string_none():
    invocation = TelemetryHandler().workflow(name="wf")
    _set_invocation_output(invocation, _Result(), capture_content=True)
    invocation.stop()
    msg = invocation.output_messages[0]
    assert msg.role == "assistant"
    assert msg.finish_reason == "stop"
```

```
AssertionError: OutputMessage(role='None', parts=[Text(content='hi', type='text')], finish_reason='None')
```

Please use an existing field or fallback to stop on success / error on failure. Please make sure there is a test coverage

instrumentation/opentelemetry-instrumentation-genai-agno/src/opentelemetry/instrumentation/genai/agno/patch.py
```diff
@@ -156,20 +198,24 @@ def _set_invocation_input(
...
             )
         ]
-    if hasattr(result, "session_id") and getattr(result, "session_id"):
+    if (
+        isinstance(invocation, AgentInvocation)
```
lmolkova: `WorkflowRunOutput` has a `session_id`, but this `isinstance` gate silently drops it because `WorkflowInvocation` has no `conversation_id`:

```python
def test_workflow_session_id(tracer_provider, logger_provider, meter_provider, span_exporter):
    with instrument(AgnoInstrumentor(), tracer_provider=tracer_provider,
                    logger_provider=logger_provider, meter_provider=meter_provider,
                    content_capture="SPAN_ONLY"):
        Workflow(name="wf", steps=[], session_id="session-1").run("hi")
    assert span.attributes.get(GenAIAttributes.GEN_AI_CONVERSATION_ID) == "session-1"
```

```
AssertionError: assert None == 'session-1'
  {'gen_ai.operation.name': 'invoke_workflow', 'gen_ai.workflow.name': 'wf', ...}
```

Please add `conversation_id` to `WorkflowInvocation` and set it here

instrumentation/opentelemetry-instrumentation-genai-agno/tests/test_agent.py
```diff
@@ -150,3 +152,113 @@ async def _run_async() -> None:
...
+    pytest.importorskip("fastapi")
+    pytest.importorskip("agno.workflow.workflow")
+    from agno.workflow.workflow import Workflow
+
+    workflow = Workflow(name="test-workflow", steps=[])
+    with patch.object(Workflow, "run", wraps=workflow.run):
+        workflow.run("test input")
+
+    spans = span_exporter.get_finished_spans()
+    assert any(
```
lmolkova: The two workflow tests don't assert much. `patch.object(Workflow, "run", wraps=workflow.run)` is a no-op - the bound method it wraps is already the instrumented one, and removing the patch produces the identical span. And `assert any(... == "invoke_workflow")` skips span count, span name, and `gen_ai.workflow.name`; the Team tests above assert all three.

Also missing across all four new tests: the error path (wrapped call raises -> span carries `error.type` and the exception re-raises unchanged).

util/opentelemetry-util-genai/src/opentelemetry/util/genai/metrics.py
```diff
@@ -58,6 +62,19 @@ def record(self, invocation: GenAIInvocation) -> None:
                 context=invocation._span_context,
             )
 
+    def record_workflow(self, invocation: GenAIInvocation) -> None:
```
lmolkova: Type this `WorkflowInvocation`, not `GenAIInvocation`. The base `_get_metric_attributes()` returns just `metric_attributes` - no `gen_ai.operation.name`, no `gen_ai.workflow.name` - so any other invocation type passed here records a silently wrong-shaped metric instead of failing.

util/opentelemetry-util-genai/src/opentelemetry/util/genai/instruments.py
```diff
@@ -48,6 +48,20 @@ def create_duration_histogram(meter: Meter) -> Histogram:
     )
 
 
+def create_workflow_duration_histogram(meter: Meter) -> Histogram:
+    name = getattr(
+        gen_ai_metrics,
+        "GEN_AI_INVOKE_WORKFLOW_DURATION",
```
lmolkova: nit: `getattr(gen_ai_metrics, "GEN_AI_INVOKE_WORKFLOW_DURATION", ...)` always takes the fallback - the constant doesn't exist in any released `opentelemetry-semantic-conventions`. It reads as forward-compat but is a hardcoded name in disguise, and if semconv-python ever defines that symbol with a different value the metric name flips silently. Please use the literal

instrumentation/opentelemetry-instrumentation-genai-agno/src/opentelemetry/instrumentation/genai/agno/patch.py
lmolkova: thanks! this is not about writing the code that does inference. Writing it is not a problem, but if we do it, we need to
- figure out suppression
- keep parity across all libs when we make changes

we already have this problem with openai and anthropic instrumented in langchain and in their own libs. 

It's readonable to add inference for providers that don't and won't have their own instrumentation - this is what we just did in smolagents. 

This is also now a part of the guidelines in this repo - https://github.com/open-telemetry/opentelemetry-python-genai/blob/main/CONTRIBUTING.md#which-operations-an-instrumentation-should-emit

instrumentation/opentelemetry-instrumentation-genai-agno/src/opentelemetry/instrumentation/genai/agno/patch.py
DylanRussell: ok good point.. I've updated the type to be the union of those 3 output types, and I just default role to assistant.. Similar changes made to the finish_reason

instrumentation/opentelemetry-instrumentation-genai-agno/src/opentelemetry/instrumentation/genai/agno/patch.py
DylanRussell: good point.. done

instrumentation/opentelemetry-instrumentation-genai-agno/tests/test_agent.py
DylanRussell: Ack, applied these changes

util/opentelemetry-util-genai/src/opentelemetry/util/genai/instruments.py
DylanRussell: Ack, done

util/opentelemetry-util-genai/src/opentelemetry/util/genai/metrics.py
DylanRussell: Ack..  good catch, fixed

instrumentation/opentelemetry-instrumentation-genai-agno/src/opentelemetry/instrumentation/genai/agno/patch.py
DylanRussell: Ack.. discussed offline we will figure out suppression first.

instrumentation/opentelemetry-instrumentation-genai-agno/tests/requirements.oldest.txt
```diff
@@ -20,6 +20,4 @@
 # factor, so they are NOT pinned here — pyproject.toml is the single source of truth. The
 # OpenTelemetry SDK and test utilities come transitively from opentelemetry-test-util-genai, which
 # every oldest env installs. Pin here only test-only deps that nothing else already provides.
-#
-# There is currently nothing to pin: agno has no test-only dependency that isn't already
-# provided by a declared bound or by the shared test fixtures.
+fastapi >= 0.100.0
```
lmolkova, line 24: why do we need it? doesn't seem like anything in tests depends on fastapi

util/opentelemetry-util-genai/src/opentelemetry/util/genai/_workflow_invocation.py
```diff
@@ -87,8 +88,19 @@ def _get_messages_for_span(self) -> dict[str, AttributeValue]:
...
+        }
+        if self._name is not None:
+            attrs[GenAI.GEN_AI_WORKFLOW_NAME] = self._name
+        attrs.update(self.metric_attributes)
+        return attrs
+
     def _apply_finish(self, error: Error | None = None) -> None:
         attributes: dict[str, AttributeValue] = self._get_messages_for_span()
+        if self.conversation_id is not None:
+            attributes[GenAI.GEN_AI_CONVERSATION_ID] = self.conversation_id
```
lmolkova, line 104: nice!

lmolkova (approved):

instrumentation/opentelemetry-instrumentation-genai-agno/tests/requirements.oldest.txt
DylanRussell: it is required for agno workflows -- so the workflow unit tests require it

PR #340, opened 2026-07-31
Add `opentelemetry-instrumentation-genai-smolagents` instrumentation package

alexander-akhmetov:
# Description

Ports `openinference-instrumentation-smolagents` into a new `opentelemetry-instrumentation-genai-smolagents` package. All telemetry is emitted through the public `opentelemetry-util-genai` API.

What is instrumented:

- `MultiStepAgent.run`, sync and streaming, as `invoke_agent`
- the model classes that define `generate`, as `chat`
- `Tool.__call__`, as `execute_tool`

It also wraps smolagents' `local_python_executor.timeout`, so tool spans started from agent-generated code keep the agent span as their parent instead of becoming root spans.

Known gaps, also listed in the package `README.rst`:

- No span for an agent step. Neither the semantic conventions nor `opentelemetry-util-genai` have a step or chain concept, so the OpenInference "Step N" CHAIN spans are dropped and `chat` and `execute_tool` spans nest directly under `invoke_agent`. See open-telemetry/semantic-conventions-genai#81.
- `Model.generate_stream` is not instrumented, so an agent created with `stream_outputs=True` reports no `chat` span and no token usage.
- A custom `Model` subclass that overrides `generate` shadows the patched base method and reports no `chat` span. A subclass that inherits `generate` is instrumented.
- A tool subclassing `PipelineTool`, including the shipped `SpeechToTextTool`, overrides `Tool.__call__` and reports no `execute_tool` span.
- `gen_ai.tool.call.id` is not recorded. `Tool.__call__` receives only the argument values, so the provider's tool call id never reaches the instrumentation.

Part of https://github.com/open-telemetry/opentelemetry-python-genai/issues/141

## Type of change

- [x] New feature (non-breaking change which adds functionality)

# How has this been tested?

Unit tests for the agent, model and tool patches, VCR-backed, and Weaver conformance scenarios for `invoke_agent`, `chat` and multimodal input. Run against smolagents 1.24.0:

```sh
uv run tox -e py310-test-instrumentation-genai-smolagents-oldest
uv run tox -e py314-test-instrumentation-genai-smolagents-latest
uv run tox -e py314-test-instrumentation-genai-smolagents-conformance
```

# Checklist

See [CONTRIBUTING.md](https://github.com/open-telemetry/opentelemetry-python-genai/blob/main/CONTRIBUTING.md)
for the style guide, changelog guidance, and more.

- [x] Followed the style guidelines of this project
- [x] Changelog updated if the change requires an entry
[...]

PR #352, opened 2026-08-03
[opentelemetry-instrumentation-genai-smolagents] Add chat instrumentation

alexander-akhmetov:
Split from #340. Agent runs (`invoke_agent`) and tool calls (`execute_tool`) will follow in separate PRs.

In smolagents, every LLM call goes through a `Model` class: `OpenAIModel`, `LiteLLMModel`, `InferenceClientModel`, `AmazonBedrockModel`, etc.

Each of these classes defines its own `generate` method, and so does the `Model` base class. The instrumentation wraps `generate` wherever it is defined. A call to a wrapped `generate` emits one `chat` span, an operation-duration metric and a token-usage metric, all through the public `opentelemetry-util-genai` API.

Known gaps, also listed in the package `README.rst`:
- If your own `Model` subclass overrides `generate`, the override hides the wrapped method and the call produces no `chat` span. A subclass that inherits `generate` is fine.
- A streamed `chat` span reports no `gen_ai.response.id` and no `gen_ai.response.model`, because a smolagents stream delta does not have them.

Part of #141

## Type of change

# How has this been tested?

```sh
uv run tox -e py310-test-instrumentation-genai-smolagents-oldest
uv run tox -e py314-test-instrumentation-genai-smolagents-latest
uv run tox -e py314-test-instrumentation-genai-smolagents-conformance
```

# Checklist

- [x] Unit tests added
- [x] Documentation updated


instrumentation/opentelemetry-instrumentation-genai-smolagents/src/opentelemetry/instrumentation/genai/smolagents/provider.py
```diff
@@ -0,0 +1,151 @@
...
+    parsed = urlparse(endpoint)
+    port = parsed.port
+    if port == 443:
+        port = None
+    return parsed.hostname or None, port
+
+
+def resolve_provider(instance: Any) -> str:
+    """Return the ``gen_ai.provider.name`` value for a smolagents model instance."""
+    class_names = _class_names(instance)
```
DylanRussell: _class_names is just called in 1 place it seems ? If thats the case I'd just inline the code

instrumentation/opentelemetry-instrumentation-genai-smolagents/src/opentelemetry/instrumentation/genai/smolagents/__init__.py
```diff
@@ -65,15 +114,51 @@ def _instrument(self, **kwargs: Any) -> None:
                 - logger_provider: LoggerProvider instance
                 - completion_hook: CompletionHook instance
         """
-        TelemetryHandler(
+        import smolagents  # pylint: disable=import-outside-toplevel  # noqa: PLC0415
```
DylanRussell: why is this outside the toplevel?

instrumentation/opentelemetry-instrumentation-genai-smolagents/src/opentelemetry/instrumentation/genai/smolagents/__init__.py
```diff
@@ -38,20 +44,63 @@
...
+    method and emits no ``chat`` span; that limitation is documented in
+    ``README.rst``.
+
+    Deduplicated by class object, because smolagents exports some classes under
+    two names (``OpenAIServerModel`` is ``OpenAIModel``) and wrapping the same
+    class twice would double every ``chat`` span.
+    """
+    from smolagents.models import (  # noqa: PLC0415  # pylint: disable=import-outside-toplevel
+        Model,
+    )
```
DylanRussell: same question.. we seem to always call this function so why not put this at the top ?

instrumentation/opentelemetry-instrumentation-genai-smolagents/src/opentelemetry/instrumentation/genai/smolagents/__init__.py
```diff
@@ -65,15 +114,51 @@ def _instrument(self, **kwargs: Any) -> None:
...
             logger_provider=kwargs.get("logger_provider"),
             completion_hook=kwargs.get("completion_hook")
             or load_completion_hook(),
         )
-        # Patching will be added in follow-up PRs.
+
+        wrapped_generate_classes: list[type] = []
+        self._wrapped_generate_classes = wrapped_generate_classes
+        wrapped_generate_stream_classes: list[type] = []
+        self._wrapped_generate_stream_classes = wrapped_generate_stream_classes
```
DylanRussell: why not just do `self._wrapped_generate_classes = []` and `self._wrapped_generate_classes.append()`

instrumentation/opentelemetry-instrumentation-genai-smolagents/src/opentelemetry/instrumentation/genai/smolagents/_messages.py
```diff
@@ -0,0 +1,316 @@
...
+            return Uri(mime_type=None, modality="image", uri=url)
+        return None
+    if content_type == "image":
+        image = element.get("image")
+        if image is not None:
+            return _image_blob(image)
+    return None
+
+
+def _parts_from_content(content: Any) -> list[Any]:
```
DylanRussell: `list[Any]` seems to broad.. looks like it should be `list[Text, ImagePart]` ?

instrumentation/opentelemetry-instrumentation-genai-smolagents/src/opentelemetry/instrumentation/genai/smolagents/_messages.py
```diff
@@ -0,0 +1,316 @@
...
+            if not isinstance(element, dict):
+                _logger.debug(
+                    "Unknown message content dropped from telemetry: %s",
+                    type(element).__name__,
+                )
+                continue
+            if element.get("type") == "text" and (text := element.get("text")):
+                parts.append(Text(content=text))
+                continue
+            image_part = _image_part_from_element(element)
```
DylanRussell: could be reduced to `if image_part := _image_part_from_element..`

instrumentation/opentelemetry-instrumentation-genai-smolagents/src/opentelemetry/instrumentation/genai/smolagents/patch.py
```diff
@@ -0,0 +1,436 @@
...
+
+    def wrapper(
+        wrapped: Callable[..., Any],
+        instance: Any,
+        args: tuple[Any, ...],
+        kwargs: dict[str, Any],
+    ) -> Any:
+        invocation = _start_inference(handler, wrapped, instance, args, kwargs)
+
+        try:
```
DylanRussell, line 391: do we want to generate telemetry if the stream fails to be initialized completely ??

I sort of think we don't want to emit telemetry in this case (it's just a user failing to use this library correctly and we should just skip telemetry in this case... just let the exception get raised..)

That's what I did for the google genai instrumentation anyway, happy to hear other opinion..

instrumentation/opentelemetry-instrumentation-genai-smolagents/src/opentelemetry/instrumentation/genai/smolagents/patch.py
```diff
@@ -0,0 +1,436 @@
...
+def model_generate(handler: TelemetryHandler) -> _Wrapper:
+    """Wrap a defining ``Model.generate`` to emit a ``chat`` span."""
+
+    def wrapper(
+        wrapped: Callable[..., Any],
+        instance: Any,
+        args: tuple[Any, ...],
+        kwargs: dict[str, Any],
+    ) -> Any:
+        invocation = _start_inference(handler, wrapped, instance, args, kwargs)
```
DylanRussell, line 298: i would use the context manager here.. `with _start_inference as invocation...` that way all the finalization and error capturing stuff is done for you automatically when you exit the `with` block..

instrumentation/opentelemetry-instrumentation-genai-smolagents/src/opentelemetry/instrumentation/genai/smolagents/patch.py
```diff
@@ -0,0 +1,436 @@
...
+        _apply_token_usage(invocation, output_message)
+        invocation.response_model_name = response_model_name(output_message)
+        invocation.response_id = response_id(output_message)
+        output = to_output_message(output_message)
+        # to_output_message leaves finish_reason empty when the provider
+        # reported none, and an empty value is dropped rather than guessed at.
+        if output.finish_reason:
+            invocation.finish_reasons = [output.finish_reason]
+        if handler.should_capture_content():
+            invocation.output_messages = [output]
```
DylanRussell: we shouldn't generate output messages at all if capture content is not enabled

instrumentation/opentelemetry-instrumentation-genai-smolagents/.changelog/352.added
```diff
@@ -0,0 +1 @@
+Add ``chat`` instrumentation for the smolagents model classes.
```
lmolkova: do we need to instrument chat operation in agent framework. 

we do it in langchain for historical reasons and because langchain is so popular and it's still quite controversial. Can we avoid doing it here? Does smolagents rely on other libraries like openai and anthropic to make model calls?

If so users that install both smalagents and openai / anthropic / etc would end up with duplicated spans.
We can spend more effort suppressing them, but I want to check if it's even necessary. 
If there is no strong reason to do it, let's just start with more interesting agentic layer

alexander-akhmetov: Oh, makes sense! I checked, most of the smolagents classes use other libraries that are instrumented already. There are just a few exceptions, like `TransformersModel`, `VLLMModel` and `MLXModel`, so I can keep `chat` only for them

instrumentation/opentelemetry-instrumentation-genai-smolagents/src/opentelemetry/instrumentation/genai/smolagents/patch.py
alexander-akhmetov: Updated: https://github.com/open-telemetry/opentelemetry-python-genai/pull/352/changes/5ba10c25fe81d93b2bee495c70e42c715f67479c

instrumentation/opentelemetry-instrumentation-genai-smolagents/src/opentelemetry/instrumentation/genai/smolagents/provider.py
alexander-akhmetov: Updated here https://github.com/open-telemetry/opentelemetry-python-genai/pull/352/changes/57afc6da2af61ad53a7bd5f717355f96bdb7be53

instrumentation/opentelemetry-instrumentation-genai-smolagents/src/opentelemetry/instrumentation/genai/smolagents/__init__.py
alexander-akhmetov: Updated: https://github.com/open-telemetry/opentelemetry-python-genai/pull/352/changes/57afc6da2af61ad53a7bd5f717355f96bdb7be53

instrumentation/opentelemetry-instrumentation-genai-smolagents/src/opentelemetry/instrumentation/genai/smolagents/_messages.py
alexander-akhmetov: Updated to `list[MessagePart]`: https://github.com/open-telemetry/opentelemetry-python-genai/pull/352/changes/57afc6da2af61ad53a7bd5f717355f96bdb7be53

instrumentation/opentelemetry-instrumentation-genai-smolagents/src/opentelemetry/instrumentation/genai/smolagents/_messages.py
alexander-akhmetov: Updated: https://github.com/open-telemetry/opentelemetry-python-genai/commit/57afc6da2af61ad53a7bd5f717355f96bdb7be53#diff-4e27c2add6463966cb455a3b3a133d5e374c0b977a6a8c73cf92a6d090db3647R138-R154

instrumentation/opentelemetry-instrumentation-genai-smolagents/src/opentelemetry/instrumentation/genai/smolagents/patch.py
alexander-akhmetov: I removed `invocation.fail(error)` and the try-except block

instrumentation/opentelemetry-instrumentation-genai-smolagents/src/opentelemetry/instrumentation/genai/smolagents/patch.py
alexander-akhmetov: Makes sense, updated: https://github.com/open-telemetry/opentelemetry-python-genai/commit/57afc6da2af61ad53a7bd5f717355f96bdb7be53

instrumentation/opentelemetry-instrumentation-genai-smolagents/.changelog/352.added
alexander-akhmetov: I updated the PR to instrument chat only for these 3 classes

instrumentation/opentelemetry-instrumentation-genai-smolagents/src/opentelemetry/instrumentation/genai/smolagents/__init__.py
alexander-akhmetov: I saw that it's done like this in other packages, when SDK imports happen inside `_instrument` and not than at the top level. My understanding was that it's for packages to be importable even when the dependency package is not installed. I assume for `opentelemetry-instrument`, but I might be wrong about this.

instrumentation/opentelemetry-instrumentation-genai-smolagents/src/opentelemetry/instrumentation/genai/smolagents/__init__.py
alexander-akhmetov: Same as above :)

instrumentation/opentelemetry-instrumentation-genai-smolagents/src/opentelemetry/instrumentation/genai/smolagents/_messages.py
```diff
@@ -0,0 +1,233 @@
...
+
+
+def _encode_image_base64(image: Any) -> str | None:
+    try:
+        from smolagents.utils import (  # pylint: disable=import-outside-toplevel
+            encode_image_base64,
+        )
+    except ImportError:
+        _logger.debug("smolagents.utils.encode_image_base64 is unavailable")
+        return None
```
lmolkova: can we do it once and cache the result so we don't do it every time on the hot path?

instrumentation/opentelemetry-instrumentation-genai-smolagents/src/opentelemetry/instrumentation/genai/smolagents/_messages.py
```diff
@@ -0,0 +1,233 @@
...
+        _logger.debug(
+            "Failed to encode image of type %s, dropping it from telemetry",
+            type(image).__name__,
+            exc_info=True,
+        )
+        return None
+    return encoded if isinstance(encoded, str) else None
+
+
+def _image_blob(image: Any) -> Blob | None:
```
lmolkova: does it need to be `Any`? can we use a more precise type?

instrumentation/opentelemetry-instrumentation-genai-smolagents/src/opentelemetry/instrumentation/genai/smolagents/_messages.py
```diff
@@ -0,0 +1,233 @@
...
+        decoded = (
+            _decode_base64_image(encoded) if encoded is not None else None
+        )
+    if decoded is None:
+        return None
+    content, mime_type = decoded
+    return Blob(mime_type=mime_type, modality="image", content=content)
+
+
+def _image_part_from_element(element: dict[str, Any]) -> Uri | Blob | None:
```
lmolkova: is there a more precise type than Any we can use here?

instrumentation/opentelemetry-instrumentation-genai-smolagents/src/opentelemetry/instrumentation/genai/smolagents/_messages.py
```diff
@@ -0,0 +1,233 @@
...
+                parts.append(image_part)
+            else:
+                _logger.debug(
+                    "Unknown message part dropped from telemetry: %s",
+                    element.get("type"),
+                )
+    return parts
+
+
+def _get_role_and_content(message: Any) -> tuple[Any, Any]:
```
lmolkova: same here, can we avoid using Any?

instrumentation/opentelemetry-instrumentation-genai-smolagents/src/opentelemetry/instrumentation/genai/smolagents/_messages.py
```diff
@@ -0,0 +1,233 @@
...
+    non-JSON-Schema ``"any"`` type. ``get_tool_json_schema`` builds exactly the
+    schema the provider receives.
+    """
+    try:
+        from smolagents.models import (  # pylint: disable=import-outside-toplevel
+            get_tool_json_schema,
+        )
+    except ImportError:
+        _logger.debug("smolagents.models.get_tool_json_schema is unavailable")
+        return None
```
lmolkova: please cache it at start time

instrumentation/opentelemetry-instrumentation-genai-smolagents/src/opentelemetry/instrumentation/genai/smolagents/_messages.py
```diff
@@ -0,0 +1,233 @@
...
+    definitions: list[ToolDefinition] = []
+    for tool in tools:
+        name = getattr(tool, "name", None)
+        if not name:
+            continue
+        definitions.append(
+            FunctionToolDefinition(
+                name=name,
+                description=getattr(tool, "description", None),
```
lmolkova: do we need getattr here?

instrumentation/opentelemetry-instrumentation-genai-smolagents/src/opentelemetry/instrumentation/genai/smolagents/patch.py
```diff
@@ -0,0 +1,364 @@
...
+    if token_usage is None:
+        return
+    invocation.input_tokens = token_usage.input_tokens
+    invocation.output_tokens = token_usage.output_tokens
+
+
+def _start_inference(
+    handler: TelemetryHandler,
+    wrapped: Callable[..., Any],
+    instance: Any,
```
lmolkova: can we use a precise type?

instrumentation/opentelemetry-instrumentation-genai-smolagents/src/opentelemetry/instrumentation/genai/smolagents/patch.py
```diff
@@ -0,0 +1,364 @@
...
+    ) -> None:
+        super().__init__(stream, invocation=invocation)
+        self._self_inference = invocation
+        self._self_capture_content = handler.should_capture_content()
+        self._self_content: list[str] = []
+        self._self_input_tokens = 0
+        self._self_output_tokens = 0
+        self._self_saw_token_usage = False
+
+    def _process_chunk(self, chunk: Any) -> None:
```
lmolkova: any chance chunk is typed?

instrumentation/opentelemetry-instrumentation-genai-smolagents/tests/conformance/inference.py
```diff
@@ -0,0 +1,151 @@
...
+from opentelemetry.test_util_genai.instrumentor import instrument
+
+from ..test_utils import (
+    MESSAGES,
+    GetWeatherTool,
+    transformers_model,
+)
+from ._helpers import attr, chat_spans
+
+# An in-process runtime returns only the generated text: no request id, no
```
lmolkova, line 35: not for this PR, just thinking out loud - should we have separate internal inference span defined for in-process inference?

lmolkova (commented):
Look great, just some concerns around `Any` - let's try to use precise types and avoid getattr whenever possible. Thanks!

instrumentation/opentelemetry-instrumentation-genai-smolagents/src/opentelemetry/instrumentation/genai/smolagents/_messages.py
alexander-akhmetov: Changed to `Image | str`. The caller passes either a PIL image or a base64 string or `data:`: https://github.com/open-telemetry/opentelemetry-python-genai/pull/352/changes/1bdbe04aaf2dffaaa8c24f39328c8801cd412c0c#diff-4e27c2add6463966cb455a3b3a133d5e374c0b977a6a8c73cf92a6d090db3647R100

instrumentation/opentelemetry-instrumentation-genai-smolagents/src/opentelemetry/instrumentation/genai/smolagents/_messages.py
alexander-akhmetov: Not really actually, it's always present. Updated.

instrumentation/opentelemetry-instrumentation-genai-smolagents/src/opentelemetry/instrumentation/genai/smolagents/patch.py
alexander-akhmetov: Changed to `Model`: https://github.com/open-telemetry/opentelemetry-python-genai/pull/352/changes/1bdbe04aaf2dffaaa8c24f39328c8801cd412c0c#diff-e41c1b208c7ec08aad90172ab5c41690bf74cff9f6234e35e8cc08aacc9c3663R228

instrumentation/opentelemetry-instrumentation-genai-smolagents/src/opentelemetry/instrumentation/genai/smolagents/patch.py
alexander-akhmetov: Updated, it's `ChatMessageStreamDelta` https://github.com/open-telemetry/opentelemetry-python-genai/pull/352/changes/1bdbe04aaf2dffaaa8c24f39328c8801cd412c0c#diff-e41c1b208c7ec08aad90172ab5c41690bf74cff9f6234e35e8cc08aacc9c3663R306

instrumentation/opentelemetry-instrumentation-genai-smolagents/src/opentelemetry/instrumentation/genai/smolagents/_messages.py
alexander-akhmetov: This one comes from smolagent, it's [defined](https://github.com/huggingface/smolagents/blob/v1.26.0/src/smolagents/models.py#L126) it as `dict[str, Any]`. I created a type alias instead, since we use it in a few places: https://github.com/open-telemetry/opentelemetry-python-genai/pull/352/changes/387bce5123fe8ce23e474450f9e2d47b56f0ad04

instrumentation/opentelemetry-instrumentation-genai-smolagents/src/opentelemetry/instrumentation/genai/smolagents/_messages.py
alexander-akhmetov: Updated: https://github.com/open-telemetry/opentelemetry-python-genai/pull/352/changes/387bce5123fe8ce23e474450f9e2d47b56f0ad04#diff-4e27c2add6463966cb455a3b3a133d5e374c0b977a6a8c73cf92a6d090db3647R162

instrumentation/opentelemetry-instrumentation-genai-smolagents/src/opentelemetry/instrumentation/genai/smolagents/_messages.py
alexander-akhmetov: I moved imports to the module level: https://github.com/open-telemetry/opentelemetry-python-genai/pull/352/changes/a8f8c33f3acf36082c698a302d81e5ee25fd1d6d. I see it's done in other packages too, and I was wrong that it needs to be here. Checked in an env without smolagents and autoinstrumentation skips it without any errors:

```
DEBUG:opentelemetry.instrumentation.auto_instrumentation._load:Distribution distro will be configured                                                                                                                                                        
DEBUG:opentelemetry.instrumentation.auto_instrumentation._load:Skipping instrumentation smolagents: DependencyConflict: requested: "smolagents>=1.24.0; extra == "instruments"" but found: "None"
```

instrumentation/opentelemetry-instrumentation-genai-smolagents/src/opentelemetry/instrumentation/genai/smolagents/patch.py
```diff
@@ -0,0 +1,357 @@
...
+    """
+
+    def wrapper(
+        wrapped: Callable[..., Generator[ChatMessageStreamDelta, None, None]],
+        instance: Model,
+        args: tuple[Any, ...],
+        kwargs: dict[str, Any],
+    ) -> _ModelStreamWrapper:
+        invocation = _start_inference(handler, wrapped, instance, args, kwargs)
+        stream = wrapped(*args, **kwargs)
```
lmolkova: please wrap in try / catch in case SDK code fails (end span with error and bubble exception up.

instrumentation/opentelemetry-instrumentation-genai-smolagents/tests/test_models.py
```diff
@@ -0,0 +1,750 @@
...
+    # Instrumentation observes; it must not change what generate_stream returns.
+    model = transformers_model(stream_chunks=["In Paris"])
+
+    stream = model.generate_stream(messages=MESSAGES)
+    assert isinstance(stream, Generator)
+    assert inspect.isgenerator(stream)
+    list(stream)
+
+
+def test_generate_stream_bad_call_emits_no_span(
```
lmolkova: Could you add the test below and fix the leak?

Test that proves both (2 fail, 2 pass on this head):

```python
class _LifecycleRecorder(SpanProcessor):
    def __init__(self) -> None:
        self.started: list[str] = []
        self.ended: list[str] = []

    def on_start(self, span, parent_context=None) -> None:
        self.started.append(span.name)

    def on_end(self, span) -> None:
        self.ended.append(span.name)

    @property
    def leaked(self) -> list[str]:
        return self.started[len(self.ended) :]


@pytest.fixture
def lifecycle(tracer_provider):
    recorder = _LifecycleRecorder()
    tracer_provider.add_span_processor(recorder)
    yield recorder


class _NotATool:
    """tools_to_call_from is a plain runtime kwarg; the annotation is a
    contract, not enforcement."""


def test_bad_tool_does_not_leak_a_span(
    instrument_with_content, span_exporter, lifecycle
) -> None:
    model = transformers_model()

    with pytest.raises(Exception):  # noqa: B017 - the SDK's own error
        model.generate(messages=MESSAGES, tools_to_call_from=[_NotATool()])

    assert span_exporter.get_finished_spans() == ()
    assert lifecycle.leaked == []


def test_generate_stream_bad_call_does_not_leak_a_span(
    instrument_with_content, span_exporter, lifecycle
) -> None:
    model = transformers_model(stream_chunks=["In Paris"])

    with pytest.raises(TypeError):
        model.generate_stream()

    assert span_exporter.get_finished_spans() == ()
    assert lifecycle.leaked == []


def test_generate_bad_call_records_an_error_span(
    instrument_with_content, span_exporter, lifecycle
) -> None:
    """Control: the sync path calls the SDK inside `with invocation`, so the
    failure ends the span instead of leaking it."""
    model = transformers_model()

    with pytest.raises(TypeError):
        model.generate()

    assert lifecycle.leaked == []


def test_recorder_sees_the_happy_path(
    instrument_with_content, span_exporter, lifecycle
) -> None:
    transformers_model().generate(messages=MESSAGES)

    assert len(lifecycle.started) == 1
    assert lifecycle.leaked == []
```

instrumentation/opentelemetry-instrumentation-genai-smolagents/src/opentelemetry/instrumentation/genai/smolagents/patch.py
alexander-akhmetov: Updated: https://github.com/open-telemetry/opentelemetry-python-genai/pull/352/commits/fedc6b1dbee97dfdfd8b2011c0cedca713cf5f13

instrumentation/opentelemetry-instrumentation-genai-smolagents/tests/test_models.py
alexander-akhmetov: Added the tests and fixed both: 

- `generate_stream`: https://github.com/open-telemetry/opentelemetry-python-genai/pull/352/changes/fedc6b1dbee97dfdfd8b2011c0cedca713cf5f13 
- request/response: https://github.com/open-telemetry/opentelemetry-python-genai/pull/352/changes/58f3072547ba3b6ee722bc6ddf099f853a2784a8

instrumentation/opentelemetry-instrumentation-genai-smolagents/src/opentelemetry/instrumentation/genai/smolagents/__init__.py
```diff
@@ -66,15 +134,45 @@ def _instrument(self, **kwargs: Any) -> None:
...
+                self._wrapped_generate_classes.append(model_cls)
+
+            for model_cls in _model_classes_defining("generate_stream"):
+                wrap_function_wrapper(
+                    model_cls,
+                    "generate_stream",
+                    model_generate_stream(handler),
+                )
+                self._wrapped_generate_stream_classes.append(model_cls)
+        except Exception:
```
lmolkova: nit, just in case

```suggestion
        except BaseException:
```

instrumentation/opentelemetry-instrumentation-genai-smolagents/src/opentelemetry/instrumentation/genai/smolagents/_messages.py
```diff
@@ -0,0 +1,243 @@
...
+def _unwrap_role(role: MessageRole | str | None) -> str | None:
+    if role is None:
+        return None
+    if isinstance(role, Enum):
+        role = role.value
+    name = str(role)
+    return _ROLE_MAP.get(name, name)
+
+
+def _decode_base64_image(image: str) -> tuple[bytes, str] | None:
```
lmolkova, line 75:
```suggestion
# TODO leverage util helpers
def _decode_base64_image(image: str) -> tuple[bytes, str] | None:
```

https://github.com/open-telemetry/opentelemetry-python-genai/pull/296 is adding some, @alexander-akhmetov mind giving it a review to make sure they'll address your needs here?

instrumentation/opentelemetry-instrumentation-genai-smolagents/src/opentelemetry/instrumentation/genai/smolagents/_messages.py
```diff
@@ -0,0 +1,243 @@
...
+        if media_type:
+            mime_type = media_type
+    try:
+        return base64.b64decode(image, validate=True), mime_type
+    except (binascii.Error, ValueError):
+        _logger.debug("Failed to decode a base64 image", exc_info=True)
+        return None
+
+
+def _encode_base64_image(image: Image) -> str | None:
```
lmolkova, line 90: this would also be a good candidate for util helper (as a follow up)

instrumentation/opentelemetry-instrumentation-genai-smolagents/src/opentelemetry/instrumentation/genai/smolagents/_messages.py
```diff
@@ -0,0 +1,243 @@
...
+    try:
+        encoded = encode_image_base64(image)
+    except Exception:  # pylint: disable=broad-except
```
lmolkova:
```suggestion
    except BaseException:  # pylint: disable=broad-except
```

lmolkova (approved):
Could you please use BaseException instead of exception? Then we should be good to go, thanks!

PR #358, opened 2026-08-05
openai: convert list-of-parts message content to semconv message parts

### LINKED ISSUE #357: openai: chat message content recorded as empty parts when it is a list of content parts
**Package**: `opentelemetry-instrumentation-genai-openai` (current `main`; the same code ships in opentelemetry-python-contrib as `opentelemetry-instrumentation-openai-v2` ≤ 2.4b0, which per that repo's guidelines now only takes security patches — hence filing here).

## Bug

When a Chat Completions message's `content` is the **list-of-content-parts** form (the standard OpenAI shape for multi-part text and for multimodal requests), the instrumentation records the message with **empty parts** in `gen_ai.input.messages`:

```json
{"role": "user", "parts": []}
```

The message content is silently dropped from telemetry, even though content capture is enabled and the actual request carried the content.

## Reproduction

```python
from opentelemetry.instrumentation.genai.openai.utils import _prepare_input_messages

messages = [
    {"role": "system", "content": "You are a fact extractor."},
    {
        "role": "user",
        "content": [
            {"type": "text", "text": "user: remember, my name is Tom"},
            {"type": "text", "text": "assistant: got it, Tom!"},
        ],
    },
]
for m in _prepare_input_messages(messages):
    print(m.role, len(m.parts))
# system 1
# user 0    <-- content dropped
```

End-to-end this shows up as GENERATION observations with an empty user message in any OTLP backend that renders `gen_ai.input.messages` (we hit it in production tracing multi-part-text extraction calls into Langfuse and spent a while chasing a phantom "empty prompt" bug in our own pipeline).

## Root cause

`_prepare_input_messages` gates content on `_is_text_part`, which only accepts `str` (or an iterable of `str`):

```python
def _is_text_part(content: Any) -> bool:
    return isinstance(content, str) or (
        isinstance(content, Iterable)
        and all(isinstance(part, str) for part in content)
    )
```

A content-parts array is a list of dicts (or typed part objects), so it fails the check and nothing is appended. `_prepare_output_messages` shares the same helper.

Note the semconv message models already have the right part types for these shapes (`Text`, `Uri`, `Blob`, `File` in `opentelemetry-util-genai`), and the sibling `opentelemetry-instrumentation-genai-anthropic` package already converts content blocks per-part (`convert_content_to_parts`) — the OpenAI package just predates that treatment.

## Expected

- `{"type": "text", ...}` parts → one `Text` part each
- `{"type": "image_url", ...}` → `Uri` (modality `image`)
- `{"type": "input_audio", ...}` → `Blob` (modality `audio`)
- `{"type": "file", ...}` → `File`
- unrecognized part types skipped rather than nuking the whole message

I have a fix with unit tests ready and will open a PR referencing this issue.

HQidea:
# Description

Follow-up to #540, which taught `_content_to_parts` about text and image parts. The remaining OpenAI content-part shapes are still dropped, which is the same failure mode as #357: an unrecognized part costs that part, and a message left without any is omitted from `gen_ai.input.messages` entirely.

Building on the converter #540 landed (its `image_url` / `input_image` branch is untouched):

- `{"type": "input_audio"}` → an audio `BlobPart`, decoded with `decode_base64()` so a malformed payload is skipped instead of recorded as garbage bytes. `wav` → `audio/wav`, `mp3` → `audio/mpeg` (`audio/mp3` is not a registered media type); an unknown format yields no media type rather than a made-up one.
- `{"type": "file"}` (Chat Completions, descriptor nested under `file`) and `{"type": "input_file"}` (Responses API, fields on the part itself) → a `FilePart` for `file_id`, or a document `BlobPart` for an inline `file_data` data: URL. Both shapes are covered for symmetry with the way `input_image` already handles both APIs.
- `{"type": "refusal"}` → a `TextPart`.

`_prepare_output_messages` now goes through `_content_to_parts` as well, and records `choice.message.refusal`. A refused completion carries `content=None` with the text in that field, so its output message used to come out empty. `_is_text_part` was the last caller of the old gate and is removed.

Fixes #357

## Type of change

# How has this been tested?

Recorded coverage, in the shape of the tests #540 added:

- `test_chat_completion_captures_audio_and_file_input` — one request carrying text, `input_audio`, a `file` by `file_id`, and a `file` with inline `file_data`, asserting `gen_ai.input.messages` on the exported span
- `test_chat_completion_captures_refusal_output` — a cassette with `"content": null, "refusal": "..."`, asserting `gen_ai.output.messages`

`test_content_parts_unit.py` pins what is awkward to record: the `mp3` and unknown-format media types, a malformed base64 payload, both Responses API `input_file` shapes, a `file` descriptor with neither `file_id` nor `file_data`, and a `refusal` part.

Seven of the nine new tests fail against `main` as it stands (the other two are guard tests that pass today only because the whole part is dropped). Full package suite: 311 passed, 12 skipped. `ruff check` and `ruff format --check` clean with the repo-pinned ruff (0.16.7).

# Checklist

- [x] Unit tests added
- [ ] Documentation updated (n/a)


instrumentation/opentelemetry-instrumentation-genai-openai/tests/test_prepare_input_messages_unit.py
```diff
@@ -0,0 +1,234 @@
...
+    Text,
+    ToolCallRequest,
+    Uri,
+)
+
+
+def test_string_content_is_single_text_part():
+    messages = [{"role": "user", "content": "Say this is a test"}]
+
+    result = _prepare_input_messages(messages)
```
lmolkova: please write realistic tests against real instrumentation flow - this is a private method that might or might not be called

lmolkova (commented):
LGTM, but let's write more realistic tests. Thanks!

instrumentation/opentelemetry-instrumentation-genai-openai/tests/test_prepare_input_messages_unit.py
HQidea: Rewrote the coverage to drive the real instrumentation flow: `test_chat_completion_multiturn_content_parts` and `test_chat_completion_multimodal_content_parts` (in `test_chat_completions.py`) make instrumented `chat.completions.create` calls against cassettes and assert `gen_ai.input.messages` / `gen_ai.output.messages` on the exported span. Both fail against the unpatched code. A slim unit module remains only for degenerate content shapes a well-formed request wouldn't carry (unknown part types, mapping-shaped content, invalid base64, `None`).

instrumentation/opentelemetry-instrumentation-genai-openai/src/opentelemetry/instrumentation/genai/openai/utils.py
```diff
@@ -165,11 +170,87 @@ def get_value(v: Any):
...
-def _is_text_part(content: Any) -> bool:
-    return isinstance(content, str) or (
-        isinstance(content, Iterable)
-        and all(isinstance(part, str) for part in content)
-    )
+def _decode_base64(data: str) -> bytes | None:
+    try:
+        return base64.b64decode(data)
+    except Exception:  # pylint: disable=broad-exception-caught
+        return None
```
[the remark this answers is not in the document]
HQidea: Narrowed to `except ValueError` (`binascii.Error` is a `ValueError` subclass). For context, the blanket form was copied from the anthropic package's `_decode_base64`; that one may deserve the same narrowing separately.

instrumentation/opentelemetry-instrumentation-genai-openai/src/opentelemetry/instrumentation/genai/openai/utils.py
```diff
@@ -165,11 +170,87 @@ def get_value(v: Any):
...
+    if isinstance(content, str):
+        return [Text(content=content)]
+    if isinstance(content, Iterable):
+        parts: list[MessagePart] = []
+        for item in content:
+            part = _convert_content_part(item)
+            if part is not None:
+                parts.append(part)
+        return parts
+    return []
```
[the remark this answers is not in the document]
HQidea: Added an explicit `Mapping` guard that records no parts — a bare dict is not a valid `content` shape, and iterating it would have produced `Text` parts for its keys. Covered by a unit test.

instrumentation/opentelemetry-instrumentation-genai-openai/.changelog/358.fixed
```diff
@@ -0,0 +1 @@
+fix chat message content being dropped from `gen_ai.input.messages`/`gen_ai.output.messages` when it is a list of content parts
```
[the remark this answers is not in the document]
HQidea: Reworded to match the existing fragments' style: capitalized sentence with RST double-backtick literals.

HQidea:
@lmolkova sorry for the slow follow-up, and thanks for the review — updated as suggested. Ready for another look whenever you have a chance.

instrumentation/opentelemetry-instrumentation-genai-openai/src/opentelemetry/instrumentation/genai/openai/utils.py
```diff
@@ -185,11 +190,91 @@ def get_value(v: Any):
...
+        url = (
+            get_property_value(image_url, "url")
+            if image_url is not None
+            else None
+        )
+        if not isinstance(url, str):
+            return None
+        # Both remote (http/https) and inline (data:) URLs are URIs; record
+        # them as sent rather than decoding inline payloads.
+        return Uri(mime_type=None, modality="image", uri=url)
```
lmolkova: `data:` URLs put the full base64 payload inline in a `Uri` — semconv wants inline data as a `Blob`, and this can push multi-MB strings into a span attribute while `mime_type` stays `None` even though the data URL carries it.

#296 adds `image_from_url()` to `opentelemetry.util.genai.utils` which does exactly this split. Suggest waiting for #296 to merge and then reusing it instead of adding a second implementation here:

```python
if part_type == "image_url":
    ...
    return image_from_url(url, modality="image", capture_content=True)
```

@rads-1996 could you take a look at this one, given the overlap with #296?

instrumentation/opentelemetry-instrumentation-genai-openai/src/opentelemetry/instrumentation/genai/openai/utils.py
```diff
@@ -185,11 +190,91 @@ def get_value(v: Any):
...
 
 
-def _is_text_part(content: Any) -> bool:
-    return isinstance(content, str) or (
-        isinstance(content, Iterable)
-        and all(isinstance(part, str) for part in content)
-    )
+def _decode_base64(data: str) -> bytes | None:
+    try:
+        return base64.b64decode(data)
```
lmolkova: Without `validate=True`, non-alphabet characters are silently discarded, so malformed audio decodes to garbage instead of being skipped. `test_invalid_base64_audio_part_is_skipped` only passes because `"!!not-base64!!"` happens to leave a bad length — e.g. `"$$$$abcd!!!!"` decodes to `b'i\xb7\x1d'`.

#296 lands `decode_base64()` in util-genai with `validate=True`; reuse that once it merges. A case that fails today:

```python
{"type": "input_audio", "input_audio": {"data": "$$$$abcd!!!!", "format": "wav"}}
# expect the part to be skipped, not a Blob of 3 junk bytes
```

instrumentation/opentelemetry-instrumentation-genai-openai/src/opentelemetry/instrumentation/genai/openai/utils.py
```diff
@@ -185,11 +190,91 @@ def get_value(v: Any):
...
+        audio_format = get_property_value(input_audio, "format")
+        return Blob(
+            mime_type=f"audio/{audio_format}"
+            if isinstance(audio_format, str)
+            else None,
+            modality="audio",
+            content=decoded,
+        )
+
+    if part_type == "file":
```
lmolkova: Only `file_id` is handled; `{"type": "file", "file": {"file_data": ..., "filename": ...}}` (the inline PDF upload shape) still drops the part entirely — same bug class as #357.

```python
file_data = get_property_value(file_obj, "file_data")
if isinstance(file_data, str):
    # file_data is a data: URL — decode to a Blob(modality="document")
    ...
```

instrumentation/opentelemetry-instrumentation-genai-openai/src/opentelemetry/instrumentation/genai/openai/utils.py
```diff
@@ -185,11 +190,91 @@ def get_value(v: Any):
...
+        input_audio = get_property_value(part, "input_audio")
+        if input_audio is None:
+            return None
+        data = get_property_value(input_audio, "data")
+        decoded = _decode_base64(data) if isinstance(data, str) else None
+        if decoded is None:
+            return None
+        audio_format = get_property_value(input_audio, "format")
+        return Blob(
+            mime_type=f"audio/{audio_format}"
```
lmolkova: OpenAI accepts `wav` and `mp3`; `f"audio/{format}"` yields `audio/mp3`, which isn't a registered mime type — it should be `audio/mpeg`. Only `wav` is covered by tests.

```python
_AUDIO_MIME_TYPES = {"wav": "audio/wav", "mp3": "audio/mpeg"}
mime_type = _AUDIO_MIME_TYPES.get(audio_format) if isinstance(audio_format, str) else None
```

instrumentation/opentelemetry-instrumentation-genai-openai/src/opentelemetry/instrumentation/genai/openai/utils.py
```diff
@@ -274,8 +357,7 @@ def _prepare_output_messages(choices) -> list[OutputMessage]:
             if tool_calls:
                 parts += extract_tool_calls_new(tool_calls)
             content = get_property_value(choice.message, "content")
-            if _is_text_part(content):
-                parts.append(Text(content=str(content)))
+            parts += _content_to_parts(content)
```
lmolkova, line 406: The new `refusal` branch only fires for request-side content lists. The actual API shape is `choice.message.refusal` with `content=None`, and that path is still unrecorded — both new cassettes have `"refusal": null`, so nothing covers it.

```python
parts += _content_to_parts(content)
refusal = get_property_value(choice.message, "refusal")
if isinstance(refusal, str):
    parts.append(Text(content=refusal))
```

Worth a cassette with `"content": null, "refusal": "..."`.

lmolkova (commented):
The core fix looks right. A few things below, mostly around part types that are still dropped and overlap with #296.

instrumentation/opentelemetry-instrumentation-genai-openai/src/opentelemetry/instrumentation/genai/openai/utils.py
HQidea: Now that #296 is merged, this uses `image_from_url()` as suggested, so a `data:` URL becomes a `BlobPart` carrying the mime type from the URL header and only remote URLs stay as a `UriPart`. (The merged signature is `image_from_url(url, *, modality="image")` — no `capture_content` argument — and this call site already runs under the capture-content check.)

The multimodal cassette covers both: a remote `https://` URL and an inline `data:image/png;base64,...` one.

instrumentation/opentelemetry-instrumentation-genai-openai/src/opentelemetry/instrumentation/genai/openai/utils.py
HQidea: Switched to `decode_base64()` from util-genai, so a malformed payload is skipped instead of decoding to garbage. Your exact counterexample is now a test: `"$$$$abcd!!!!"` (which used to decode to `b'i\xb7\x1d'`) leaves only the sibling text part.

instrumentation/opentelemetry-instrumentation-genai-openai/src/opentelemetry/instrumentation/genai/openai/utils.py
HQidea: Handled. `file_id` still becomes a `FilePart`, and an inline `file_data` data: URL is decoded through `image_from_url(..., modality="document")` into a document `BlobPart` (mime type from the URL header). A `file` part carrying neither is skipped, with a test to pin that.

The multimodal cassette sends both shapes in one request; `filename` has no place in the semconv part models, so it is dropped.

instrumentation/opentelemetry-instrumentation-genai-openai/src/opentelemetry/instrumentation/genai/openai/utils.py
HQidea: Fixed with the mapping you suggested (`wav` -> `audio/wav`, `mp3` -> `audio/mpeg`); an unknown format yields `mime_type=None` rather than a made-up type. `mp3` now has its own test.

instrumentation/opentelemetry-instrumentation-genai-openai/src/opentelemetry/instrumentation/genai/openai/utils.py
HQidea: Good catch — that path was indeed unrecorded. `_prepare_output_messages` now appends `choice.message.refusal` as a `TextPart`, and `test_chat_completion_refusal` drives it through a new cassette with `"content": null, "refusal": "..."`. It fails without the fix.

instrumentation/opentelemetry-instrumentation-genai-openai/src/opentelemetry/instrumentation/genai/openai/utils.py
```diff
@@ -186,11 +190,95 @@ def get_value(v: Any):
...
-    return isinstance(content, str) or (
-        isinstance(content, Iterable)
-        and all(isinstance(part, str) for part in content)
-    )
+# OpenAI accepts "wav" and "mp3" for input_audio; "audio/mp3" is not a
+# registered media type, the payload is MPEG audio.
+_AUDIO_MIME_TYPES = {"wav": "audio/wav", "mp3": "audio/mpeg"}
+
+
+def _convert_content_part(part: Any) -> MessagePart | None:
```
lmolkova: PR #540 just merged into `main` and introduced `_content_to_parts` supporting `input_image` (Responses API), `input_text`, and `output_text`. Please rebase on `main` and build on top of the existing `_content_to_parts` implementation to avoid merge conflicts and preserve `input_image` support.

HQidea: Rebased on `main` and rewritten on top of your `_content_to_parts`. The `image_url` / `input_image` branch is untouched — the new part types are handled before it, so `input_image` (including the `file_id` form) keeps working, and the text handling is now yours rather than a second implementation.

#540 also means the original `{"type": "text"}` case from #357 is already fixed on `main`, so this PR is now scoped to what is still dropped:

- `input_audio` → an audio `BlobPart` via `decode_base64()`, with `wav` → `audio/wav` and `mp3` → `audio/mpeg`
- `file` (Chat Completions, descriptor nested under `file`) and `input_file` (Responses API, fields on the part) → a `FilePart` for `file_id` or a document `BlobPart` for an inline `file_data` data: URL. I covered both shapes for symmetry with how `input_image` handles both APIs — happy to drop `input_file` if you would rather keep that for a separate change.
- `refusal` parts → `TextPart`
- `_prepare_output_messages` now goes through `_content_to_parts` and records `choice.message.refusal`; a refused completion carries `content=None`, so that output message used to come out empty. `_is_text_part` was its last caller and is gone.

Tests follow the shape of the ones #540 added: `test_chat_completion_captures_audio_and_file_input` and `test_chat_completion_captures_refusal_output` against new cassettes, plus a small unit module for the mp3 mapping, the Responses `input_file` shapes, and the malformed descriptors. Seven of the nine fail against `main` as it stands. Full suite: 311 passed, 12 skipped.

instrumentation/opentelemetry-instrumentation-genai-openai/src/opentelemetry/instrumentation/genai/openai/utils.py
```diff
@@ -216,6 +254,34 @@ def _content_to_parts(content: Any) -> list[MessagePart]:
...
+            continue
+
+        if part_type in ("file", "input_file"):
+            # Chat Completions nests the descriptor under "file"; the
+            # Responses API carries the same fields on the part itself.
+            file_part = _document_to_part(
+                get_property_value(item, "file")
+                if part_type == "file"
+                else item
+            )
```
lmolkova, line 271: Handles un-nested `file` parts defensively and removes the ternary.
```suggestion
            file_part = _document_to_part(
                get_property_value(item, "file") or item
            )
```

instrumentation/opentelemetry-instrumentation-genai-openai/src/opentelemetry/instrumentation/genai/openai/utils.py
```diff
@@ -337,8 +403,12 @@ def _prepare_output_messages(choices) -> list[OutputMessage]:
...
                 parts += extract_tool_calls_new(tool_calls)
             content = get_property_value(choice.message, "content")
-            if _is_text_part(content):
-                parts.append(TextPart(content=str(content)))
+            parts += _content_to_parts(content)
+            # A refused completion carries content=None and puts the text
+            # in its own `refusal` field.
+            refusal = get_property_value(choice.message, "refusal")
+            if isinstance(refusal, str):
+                parts.append(TextPart(content=refusal))
```
lmolkova, line 414: In multi-turn chat history where a prior assistant turn has `content: null, refusal: "..."`, `_prepare_input_messages` still drops the message because `message.refusal` is not extracted there.

Please extract `refusal` on assistant messages in `_prepare_input_messages` as well:
```python
if role == Role.ASSISTANT.value:
    tool_calls = get_property_value(message, "tool_calls")
    if tool_calls:
        parts += extract_tool_calls_new(tool_calls)
    parts += _content_to_parts(content)
    refusal = get_property_value(message, "refusal")
    if isinstance(refusal, str):
        parts.append(TextPart(content=refusal))
```

And cover with a test:
```python
def test_prepare_input_messages_captures_assistant_refusal():
    messages = [
        {"role": "user", "content": "disallowed request"},
        {"role": "assistant", "content": None, "refusal": "I cannot help with that."},
    ]
    input_messages = _prepare_input_messages(messages)
    assert len(input_messages) == 2
    assert input_messages[1].parts == [TextPart(content="I cannot help with that.")]
```

lmolkova (approved):
LGTM with a few comments. Thanks!

instrumentation/opentelemetry-instrumentation-genai-openai/src/opentelemetry/instrumentation/genai/openai/utils.py
HQidea: Applied in 37ffc2c — `get_property_value(item, "file") or item`, so an un-nested `file` part works too and the ternary is gone.

instrumentation/opentelemetry-instrumentation-genai-openai/src/opentelemetry/instrumentation/genai/openai/utils.py
HQidea: Good catch, that history case was still lost. Applied in 37ffc2c, with your test as `test_prepare_input_messages_captures_assistant_refusal` in `tests/test_utils.py`; it fails without the change. Full suite: 312 passed, 12 skipped.

PR #93, opened 2026-05-26
feat(bedrock): Migrate AWS Bedrock Runtime instrumentation

williazz:
## Summary

- Adds `opentelemetry-instrumentation-genai-bedrock` as a new package in this monorepo
- Instruments the Bedrock Runtime Converse and ConverseStream APIs using the modern `TelemetryHandler`/`InferenceInvocation` lifecycle from `opentelemetry-util-genai`
- Patches `botocore.client.BaseClient._make_api_call` directly (no dependency on `opentelemetry-instrumentation-botocore`)

### Provenance

This is extracted from the existing Bedrock instrumentation in [`opentelemetry-python-contrib`](https://github.com/open-telemetry/opentelemetry-python-contrib), originally authored by @riccardo-magliocchetti:

**Source modules:**
- [`extensions/bedrock.py`](https://github.com/open-telemetry/opentelemetry-python-contrib/blob/main/instrumentation/opentelemetry-instrumentation-botocore/src/opentelemetry/instrumentation/botocore/extensions/bedrock.py)
- [`extensions/bedrock_utils.py`](https://github.com/open-telemetry/opentelemetry-python-contrib/blob/main/instrumentation/opentelemetry-instrumentation-botocore/src/opentelemetry/instrumentation/botocore/extensions/bedrock_utils.py)
- [`tests/test_botocore_bedrock.py`](https://github.com/open-telemetry/opentelemetry-python-contrib/blob/main/instrumentation/opentelemetry-instrumentation-botocore/tests/test_botocore_bedrock.py)

**Upstream PRs:**
- [#3161](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/3161) — Basic instrumentation for Bedrock runtime Converse API
- [#3258](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/3258) — Add Bedrock GenAI user events and lazy initialize tracers

The code has been rewritten to use the shared GenAI utilities (`TelemetryHandler`, `InferenceInvocation`, semconv enums) in this repo rather than the botocore extension pattern.

### What's included

- Package scaffold with `pyproject.toml`, entry points, and changelog support
- Non-streaming (`Converse`) instrumentation with full attribute extraction
- Streaming (`ConverseStream`) instrumentation with event accumulation
- Request/response attribute extraction (model, tokens, finish reasons, server address)
- Input/output message capture (text and tool use)
- Basic tests using botocore Stubber
- tox environment and workspace integration

### What's NOT included yet (follow-up work)

[...]

PR #359, opened 2026-08-05
Add initial `opentelemetry-instrumentation-genai-bedrock` boilerplate

DylanRussell:
# Description

Add initial `opentelemetry-instrumentation-genai-bedrock` boilerplate. `boto3` doesn't have any reference to open telemetry in it's repo, so no native instrumentation.

Created the package structure targeting "boto3 >= 1.40.46" (matching openinference-instrumentation-bedrock),

## Type of change

- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [x] This change requires a documentation update

# How has this been tested?
 Unit tests

# Checklist

- [x] Unit tests added
- [x] Documentation updated


eternalcuriouslearner:
@lmolkova Should we migrate bedrock? I was under the assumption that it is going to be left in contrib.

DylanRussell:
i dont think boto3 is instrumented in contrib.. well I see boto3 sqs is instrumented: https://github.com/open-telemetry/opentelemetry-python-contrib/blob/e5e3af0ccdb9c3d85728f6d7cb03d77c72196425/instrumentation/opentelemetry-instrumentation-boto3sqs/README.rst but that isn't the API we care about

eternalcuriouslearner (approved):

instrumentation/opentelemetry-instrumentation-genai-bedrock/.changelog/359.added
```diff
@@ -0,0 +1 @@
+Add the initial Amazon Bedrock instrumentation package setup.
```
lmolkova, line 1: how is this related to opentelemetry-instrumentation-botocore that contains bedrock runtime?

https://github.com/open-telemetry/opentelemetry-python-contrib/tree/main/instrumentation/opentelemetry-instrumentation-botocore#bedrock-runtime-botocore-only

see also 

https://cloud-native.slack.com/archives/C06KR7ARS3X/p1784057333165819

>  I think the benefits of potentially splitting the package out would be that it could integrate better into the GenAI repo/have more GenAI folks looking after it like @rm mentioned. However, if we perform a split as done in https://github.com/open-telemetry/opentelemetry-python-genai/pull/93 this would likely create issues with double instrumentation if you attempted to use both the gen AI package and the contrib botocore package.

> if we split it, one should be a dependency of the other to not break user

xrmx:
> i dont think boto3 is instrumented in contrib.. well I see boto3 sqs is instrumented: https://github.com/open-telemetry/opentelemetry-python-contrib/blob/e5e3af0ccdb9c3d85728f6d7cb03d77c72196425/instrumentation/opentelemetry-instrumentation-boto3sqs/README.rst but that isn't the API we care about

The one in contrib instruments botocore that is what powers boto3. So the layer would be different but you'll still get duplicated telemetry.

~The current instrumentation in -contrib instruments two classes of APIs: the old `InvokeModel` and the newer `Converse`. Both of these have a sync and stream version of the API. Does the openinference one supports both? Yes~

DylanRussell:
Hmm ok I missed that, so there's a gen ai instrumentation of the botocore library (which underlies boto3) in contrib..

It seems not great to have just 1 gen ai instrumentation in contrib and the rest in this repo going forward.. I also think it has some gaps: it doesn't use gen ai utils (is missing stuff like the completion hook and the content capture flag), and doesn't instrument aiobotocore for the gen AI stuff..

So my proposal is we deprecate that one and add one atop of boto3 here in this library.. WDYT ?

Also I think it's mostly fine that we instrument boto3 here and in the contrib repo, since the one contrib monkey patches just the SQS methods: https://github.com/open-telemetry/opentelemetry-python-contrib/blob/b4e7e098f81c3713798a5747c76e87f8c03b1e7e/instrumentation/opentelemetry-instrumentation-boto3sqs/src/opentelemetry/instrumentation/boto3sqs/__init__.py#L338-L342 -- we will instrument a different set of methods in this repo, so there's no conflict..

I think we also need to be careful of the entry point name we use for auto instrumentation boto3 has been taken --- https://github.com/open-telemetry/opentelemetry-python-contrib/blob/b4e7e098f81c3713798a5747c76e87f8c03b1e7e/instrumentation/opentelemetry-instrumentation-boto3sqs/pyproject.toml#L40 -- so we have to use something different there.. I have "bedrock" here now but we can use something else..

Proposal for how to deprecate the existing bedrock instrumentation:

Add a deprecation warning:

```
class _BedrockExtension:
    def __init__(self, ...):
        warnings.warn(
            "The Bedrock extension in 'opentelemetry-instrumentation-botocore' is deprecated "
            "and will be removed in a future release. "
            "Please use 'opentelemetry-instrumentation-genai-bedrock' instead.",
            category=DeprecationWarning,
            stacklevel=2,
        )
```

Disable `BedrockExtension` if `opentelemetry-instrumentation-genai-bedrock` is installed and enabled:

```
def _is_genai_bedrock_active() -> bool:
    if importlib.util.find_spec("opentelemetry.instrumentation.genai.bedrock") is None:
        return False
    try:
        from opentelemetry.instrumentation.genai.bedrock import BedrockInstrumentor
        return BedrockInstrumentor().is_instrumented_by_opentelemetry
    except Exception:
        return False
class _BedrockExtension:
    def __init__(self, ...):
        self._disabled = _is_genai_bedrock_active()
        if self._disabled:
            logger.debug(
                "opentelemetry-instrumentation-genai-bedrock is active; "
                "suppressing botocore BedrockExtension to avoid duplicate spans."
            )
    def extract_attributes(self, ...):
        if self._disabled:
            return  # No-op: let the new instrumentor handle GenAI telemetry
        ...
```

xrmx:
> Hmm ok I missed that, so there's a gen ai instrumentation of the botocore library (which underlies boto3) in contrib..
> 
> It seems not great to have just 1 gen ai instrumentation in contrib and the rest in this repo going forward.. I also think it has some gaps: it doesn't use gen ai utils (is missing stuff like the completion hook and the content capture flag), and doesn't instrument aiobotocore for the gen AI stuff..
> 
> So my proposal is we deprecate that one and add one atop of boto3 here in this library.. WDYT ?
> 
> Proposal for how to deprecate the existing bedrock instrumentation:
> 
> Add a deprecation warning:
> ...
> 
> Disable `BedrockExtension` if `opentelemetry-instrumentation-genai-bedrock` is installed and enabled:
> ...

Sounds good to me but the question is when we deprecate that? I would like to do so only once this has feature parity. And also when `opentelemetry-bootstrap` will start offering python-genai instrumentations. Have you already discussed when to substitute the -contrib ones there?

DylanRussell:
Yeah definitely don't deprecate until feature parity + release.

I think https://github.com/open-telemetry/opentelemetry-python-contrib/pull/4927/changes will fix that issue with `opentelemetry-bootstrap`

instrumentation/opentelemetry-instrumentation-genai-bedrock/.changelog/359.added
DylanRussell: Yeah I missed this initially. Please see the discussion below -- I think we should proceed with this PR now..

lmolkova:
Can we document a plan on how we approach bedrock and discuss on the Python SIG call? I think having an issue in python-contrib repo that captures steps would be great.

instrumentation/opentelemetry-instrumentation-genai-bedrock/tests/requirements.oldest.txt
```diff
@@ -0,0 +1,5 @@
+# Copyright The OpenTelemetry Authors
+# SPDX-License-Identifier: Apache-2.0
+
+# The oldest tox factor resolves declared dependencies from pyproject.toml
+# with UV_RESOLUTION=lowest-direct. There are no additional test-only pins.
```
xrmx, line 5: Shouldn't this pin boto3==1.40.46?

instrumentation/opentelemetry-instrumentation-genai-bedrock/src/opentelemetry/instrumentation/genai/bedrock/version.py
```diff
@@ -0,0 +1,4 @@
+# Copyright The OpenTelemetry Authors
+# SPDX-License-Identifier: Apache-2.0
+
+__version__ = "1.2b0.dev"
```
xrmx, line 4:
```suggestion
__version__ = "1.0b0.dev"
```

xrmx (approved):

instrumentation/opentelemetry-instrumentation-genai-bedrock/tests/requirements.oldest.txt
DylanRussell: `oldest` already resolves to the floor In tox.ini, the oldest factor sets `UV_RESOLUTION=lowest-direct`. uv resolves all package dependencies to their lower bounds.

instrumentation/opentelemetry-instrumentation-genai-bedrock/src/opentelemetry/instrumentation/genai/bedrock/version.py
DylanRussell: We are requiring all versions match exactly (major.minor must match) in opentelemetry-python-genai-1/scripts/version_utils.py -- not sure the reaosning behind this though.. Probably @lmolkova knows

DylanRussell:
Opened up https://github.com/open-telemetry/opentelemetry-python-contrib/issues/4999 .. happy to discuss on sig call

DylanRussell:
Re-opening this under https://github.com/open-telemetry/opentelemetry-python-genai/pull/486 so I can use gh stack feature..

Issue #169, opened 2026-06-25
Full type checking pyright coverage


aabmass:
Adding type annotations and getting pyright passing is not in scope for migrated OpenInference libs. The skill we use tells the agent to opt the package out of pyright checking.

This is good for the initial pass, but it would be much better to have static coverage for instrumentations, and required any new code to pass type checking.

_Originally posted by @lmolkova in https://github.com/open-telemetry/opentelemetry-python-genai/pull/143#discussion_r3463754653_

bhumikadangayach:
Hey @aabmass, I can pick this up.
Heads up — #153 already closed most of #41 (removing Any from the invocation classes), except the dataclass fields in types.py, which were called out as out of scope there. So, util-genai's types.py seems like the obvious starting point, and it'd close #41 fully too.
Before I touch anything else — do you want Pyright gating new code from day one with existing files exempted (# pyright: ignore or a per-file allowlist), and we burn that down package by package? Or strict mode required everywhere immediately?
I'll open a small PR for util-genai/types.py first either way, just want to agree on the exemption strategy before doing the rest of the packages.

bhumikadangayach:
Hey @aabmass, just following up on the gating strategy question above (gate-new-code vs. strict-everywhere) — no rush, but happy to go with whichever you prefer. If it's easier, I can default to gate-new-code (seems like the more common convention) and start on util-genai/types.py unless you'd rather I wait for your input.

PR #366, opened 2026-08-07
opentelemetry-instrumentation-genai-qwen-agent: support strict typing

Yigtwxx:
# Description

Adds `opentelemetry-instrumentation-genai-qwen-agent` to the `[tool.pyright]` `include` list, per the progressive rollout the config already describes, and excludes its `tests/` and `examples/` like the other covered packages.

Strict mode reported eight errors, all in `utils.py` and all the same shape: `isinstance(x, dict)` / `isinstance(x, list)` narrowing to `dict[Unknown, Unknown]` / `list[Unknown]`. They are narrowed with `cast(dict[str, Any], ...)` and `cast(list[Any], ...)`, matching the idiom already used in `opentelemetry-instrumentation-genai-agno`. No behavior change and no new public surface.

The package was already listed in the `typecheck` tox env deps, so no `tox.ini` change is needed.

Related to #169

## Type of change

# How has this been tested?

- [x] `tox -e typecheck` — 0 errors (was 8 with the package included and unfixed)
- [x] `tox -e py311-test-instrumentation-genai-qwen-agent-latest` — 22 passed
- [x] `tox -e py310-test-instrumentation-genai-qwen-agent-oldest` — 21 passed, 1 skipped
- [x] `tox -e lint-instrumentation-genai-qwen-agent` and `tox -e precommit`

The existing tests are unchanged, which is the evidence that the narrowing did not alter behavior.

# Checklist

- [ ] Unit tests added
- [ ] Documentation updated


instrumentation/opentelemetry-instrumentation-genai-qwen-agent/src/opentelemetry/instrumentation/genai/qwen_agent/utils.py
```diff
@@ -81,7 +82,7 @@ def find_tool_call_id(messages: Any, tool_name: str) -> str | None:
 
     responded: set[str] = set()
     pending: list[str] = []
-    for msg in messages:
+    for msg in cast(list[Any], messages):
```
lmolkova: can we do even better and specify the type of elements?

lmolkova (approved):
nice!

instrumentation/opentelemetry-instrumentation-genai-qwen-agent/src/opentelemetry/instrumentation/genai/qwen_agent/utils.py
Yigtwxx: Done in 1131e05 — the casts are gone entirely.

Instead of `cast(list[Any], ...)` I annotated the signatures with
qwen-agent's own `Message` / `ContentItem`, imported under
`TYPE_CHECKING` since qwen-agent is an optional dependency. Pyright
narrows the `isinstance` checks on its own from there, so all five
list casts dropped out.

The only cast left is the `dict[str, Any]` one in `_field_value`,
where the lookup genuinely is a heterogeneous mapping.

PR #381, opened 2026-08-12
fix(anthropic): handle with_raw_response results without crashing

### LINKED ISSUE #380: Anthropic instrumentation raises AttributeError on messages.create(...) called via with_raw_response
### What happened

With `AnthropicInstrumentor` enabled, calling `messages.create(...)` through `with_raw_response` crashes the application with an `AttributeError` — *after* the (billed) API call has already succeeded. Instrumentation must never break the app it observes.

### Repro

```python
from opentelemetry.instrumentation.genai.anthropic import AnthropicInstrumentor
import anthropic

AnthropicInstrumentor().instrument()
client = anthropic.Anthropic()

raw = client.messages.with_raw_response.create(
    model="claude-sonnet-4-20250514",
    max_tokens=100,
    messages=[{"role": "user", "content": "Say hello in one word."}],
)
# -> AttributeError: 'LegacyAPIResponse' object has no attribute 'model'
```

### Root cause

`with_raw_response.create(...)` returns a `LegacyAPIResponse`, not a `Message`. In `patch.py`, `traced_method` only special-cases `AnthropicStream`; everything else is passed to `MessageWrapper(result).extract_into(invocation)`:

- `instrumentation/opentelemetry-instrumentation-genai-anthropic/src/opentelemetry/instrumentation/genai/anthropic/patch.py` (sync `messages_create` and async `async_messages_create`)
- which calls `set_invocation_response_attributes`, reading `message.model` at `messages_extractors.py:155` → `AttributeError`, re-raised into the caller via the `except Exception: invocation.fail(exc); raise` block.

### Precedent for the fix

The sibling OpenAI package already handles this. `openai/_raw_response.py` defines a `ParsableResponse` protocol, and `openai/patch.py` (`_set_response_properties`) / `response_extractors.py` (`_parse_raw_response`) call `result.parse()` to unwrap the raw response for telemetry while returning the original raw response to the caller. The Anthropic package needs the same treatment.

### Impact

Any user of `with_raw_response` (common for reading rate-limit / request-id headers) gets a hard crash once instrumentation is enabled — violating the core "instrumentation must not break the application" guarantee. Affects sync and async clients, and the `parse()` wrappers that reuse the create path.

I have a fix + sync/async regression tests ready to submit.

Shailendra005:
Fixes #380

### Problem

`messages.create(...)` invoked via `with_raw_response` returns a `LegacyAPIResponse`, but the instrumentation assumed a `Message` and read `message.model` during extraction, raising `AttributeError` into the caller *after* a successful (billed) API call. Instrumentation must not break the app.

### Fix

Mirror the OpenAI sibling package's `ParsableResponse` pattern:

- Add a `_RawResponse` runtime-checkable Protocol (shape match on `parse()`, so we don't import the SDK's private response classes) and a `_message_for_extraction()` helper.
- In both sync `messages_create` and async `async_messages_create`, parse the raw response to obtain the `Message` for telemetry, then **return the original result untouched** so raw-response callers still get their headers and `parse()`.
- Streaming raw responses (`stream=True`) and any unexpected payload are left uninstrumented instead of crashing.
- The existing `Messages.parse` / `AsyncMessages.parse` wrappers reuse the create path, so they are covered too.

### Tests

Added `test_sync_messages_create_with_raw_response` and `test_async_messages_create_with_raw_response` (with cassettes). They fail before the fix (`AttributeError: 'LegacyAPIResponse' object has no attribute 'model'`) and pass after. Full package suite: 108 passed, 3 skipped. `ruff check` / `ruff format` clean.

### Changelog

Added a towncrier `fixed` fragment; I'll rename it to `<PR_NUMBER>.fixed` to match the convention.


Shailendra005:
Thanks for the review. Addressing both comments:

**1. `AnthropicMessage | None` return type (patch.py):** No change needed. The module has `from __future__ import annotations` (line 6), so all annotations are strings and never evaluated at runtime, and the package's `requires-python` is `>=3.10`. The surrounding code already uses PEP 604 unions in its annotations (e.g. the existing `messages_create` return type), so this stays consistent with the module.

**2. Duplicate interaction in the async cassette:** Good catch — fixed in 4d28f80. The async test issues a single `with_raw_response.create(...)` call, so I trimmed the cassette to the one expected interaction (the sync cassette already had just one). Full package suite still passes (108 passed, 3 skipped).

instrumentation/opentelemetry-instrumentation-genai-anthropic/src/opentelemetry/instrumentation/genai/anthropic/patch.py
```diff
@@ -52,6 +52,39 @@
...
+    ``Messages.create`` normally returns a ``Message``. When called through
+    ``with_raw_response`` it returns a raw-response object instead; parsing it
+    yields the ``Message`` (safe for non-streaming calls, mirroring the OpenAI
+    instrumentation). Streaming raw responses and any other unexpected payload
+    are left uninstrumented rather than crashing the caller's application.
+    """
+    if isinstance(result, AnthropicMessage):
+        return result
+    if not stream_requested and isinstance(result, _RawResponse):
+        parsed = result.parse()
```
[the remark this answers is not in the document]
Shailendra005: Fixed in 80032e2. The `parse()` is now guarded: any failure is swallowed and the raw response is returned untouched, so a headers-only caller is never broken (added a regression test). Parsing is kept only on the non-streaming path, mirroring the OpenAI sibling `openai/patch.py::_set_response_properties` ("safe to parse() here since this is the non-streaming path"): the body is already fully received, so it is a pure memoized deserialization, no stream consumption or early I/O. Streaming raw responses are never parsed here. Happy to switch to a deferred `wrapt.ObjectProxy` if you prefer strict lazy parity.

instrumentation/opentelemetry-instrumentation-genai-anthropic/src/opentelemetry/instrumentation/genai/anthropic/patch.py
```diff
@@ -52,6 +52,56 @@
...
+    telemetry must never break the application (AGENTS.md: "Do not raise new
+    exceptions in instrumentation/telemetry code"). Streaming raw responses are
+    never parsed here (that would consume the stream); they are left
+    uninstrumented.
+    """
+    if isinstance(result, AnthropicMessage):
+        return result
+    if not stream_requested and isinstance(result, _RawResponse):
+        try:
+            parsed = result.parse()
```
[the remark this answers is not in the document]
Shailendra005: Good catch, fixed in 36455c8. Detection now also checks the Stainless `x-stainless-raw-response: stream` header (not just the `stream` kwarg), mirroring the OpenAI sibling `is_streamed_raw_response` in `response_extractors.py`. `with_streaming_response` results are left un-parsed, so the streaming body is never read before the caller receives it. Added a regression test asserting `parse()` is not called for a raw response carrying that header.

instrumentation/opentelemetry-instrumentation-genai-anthropic/src/opentelemetry/instrumentation/genai/anthropic/patch.py
```diff
@@ -86,10 +160,13 @@ def traced_method(
...
                 )
 
-            wrapper = MessageWrapper(result, capture_content)
-            wrapper.extract_into(invocation)
+            message = _message_for_extraction(result, kwargs)
+            if message is not None:
+                MessageWrapper(message, capture_content).extract_into(
+                    invocation
+                )
             invocation.stop()
```
lmolkova: Not a blocker for this fix, but worth a follow-up.

With `with_raw_response.create(stream=True)` (and `with_streaming_response`) the SDK
returns a `LegacyAPIResponse`, not a `Stream`, so it falls past the
`isinstance(result, AnthropicStream)` check and we call `stop()` right away. That gives a
success span with no response attributes, ended before the caller has even called `parse()`.

The openai package solves this in `_raw_response.py` (#278): the raw response is returned
wrapped in a transparent `ObjectProxy` so metadata still resolves natively, `parse()` is
deferred and returns the instrumented stream wrapper so the span ends when the stream is
drained, and a `close`/`aclose` fallback finalizes the span if the caller never parses.
Would be good to do the same here.

lmolkova (approved):
Thanks for the fix!

Shailendra005:
Thanks @lmolkova, that was a real gap. Addressed in d2fbf1f (this PR).

Added `_raw_response.py` mirroring the OpenAI package (#278):

- `RawResponseStreamProxy(wrapt.ObjectProxy)` wraps the raw response transparently, so `headers` / `request_id` / `http_response` still resolve natively and `isinstance` / `__class__` see the original type.
- `parse()` is deferred and memoized. When called it returns the instrumented stream wrapper (`MessagesStreamWrapper` for a `Stream`, `AsyncMessagesStreamWrapper` for an `AsyncStream`), so the span ends when the stream is drained rather than at create time. A non-stream / awaitable `parse()` result is returned untouched with a debug log.
- A `close` / `aclose` fallback on `http_response` finalizes the span exactly once when the caller drains the body or never parses.

Wired into both `messages_create` and `async_messages_create`: a streaming `with_raw_response` / `with_streaming_response` result now goes through the proxy instead of the immediate `stop()`. Non-streaming raw responses are unchanged.

Tests: sync + async, covering metadata still resolving, `parse()` returning a drainable stream, the span carrying response model / tokens / finish reason after draining, and the never-parse-then-close path finalizing once. Full package suite is green (114 passed, 3 skipped); ruff check + format clean.

Matching on `parse()` + `http_response` shape (not private SDK types) to stay version-robust. Would appreciate your review when you have a moment.


Can you re-review it?

instrumentation/opentelemetry-instrumentation-genai-anthropic/src/opentelemetry/instrumentation/genai/anthropic/patch.py
```diff
@@ -52,6 +53,80 @@
...
+    shape rather than importing private names.
+    """
+
+    def parse(self, *args: Any, **kwargs: Any) -> Any: ...
+
+
+_RAW_RESPONSE_HEADER = "x-stainless-raw-response"
+
+
+def _is_streaming_request(kwargs: Mapping[str, Any]) -> bool:
```
lmolkova: Thanks for the update!


Could you add a test that goes through `with_streaming_response`? Neither the streaming
nor the non-streaming path through it is covered right now, and the non-streaming one
is broken.

`_is_streaming_request` keys off the `x-stainless-raw-response: stream` header, but the SDK sets that header on every `with_streaming_response` call, including non-streaming ones. The call gets routed into the stream proxy and the response is never extracted, so the span ends up with request attributes only.

The snippet below passes on the current head, which is the bug: the assertions say the response attributes are absent, and they are.

```python
    with client.messages.with_streaming_response.create(
        model=model, max_tokens=100, messages=messages,
    ) as raw_response:
        message = raw_response.parse()   # -> Message, not a Stream

    assert "gen_ai.response.model" not in span.attributes
    assert "gen_ai.usage.input_tokens" not in span.attributes
```

`gen_ai.client.token.usage` is missing for the same reason; only
`gen_ai.client.operation.duration` is emitted. Routing on what `parse()` returns rather than on the header would fix all of it.

Separately, `AsyncAPIResponse.parse()` is a coroutine, so async `with_streaming_response` streams are handed back uninstrumented and get a request-only span too. The async variant
of the same test would catch it.

Shailendra005: Fixed in 3e419b1. Routing now keys off what `parse()` returns, not the header.

The proxy defers `parse()` and dispatches on the value: a `Message` (non-streaming `with_raw_response` or `with_streaming_response`) extracts response telemetry and finalizes the span; a `Stream` / `AsyncStream` returns the instrumented wrapper that finalizes on drain; `AsyncAPIResponse.parse()` returns a coroutine, which the proxy awaits before dispatching by the same rules. The `x-stainless-raw-response` header is no longer consulted.

One subtlety I hit: a non-streaming `parse()` reads and closes the httpx body synchronously, which fired the `close`/`aclose` fallback and finalized the span before dispatch could extract. Added a parse-in-flight guard so the fallback defers while a parse is running and the parse owns finalization.

Tests (sync and async, fail-before / pass-after): your `with_streaming_response` non-streaming snippet with the assertions inverted (response model, input tokens, and the `gen_ai.client.token.usage` metric are now present), plus streaming `with_streaming_response`, and the existing `with_raw_response` paths. Full suite 118 passed, 3 skipped; ruff check and format clean.

Caveat: the async paths are exercised via VCR cassettes, not a live key.

instrumentation/opentelemetry-instrumentation-genai-anthropic/src/opentelemetry/instrumentation/genai/anthropic/_raw_response.py
```diff
@@ -0,0 +1,275 @@
...
+            finalize, self._self_finalize = self._self_finalize, None
+            finalize()
+
+    def parse(self, *args: Any, **kwargs: Any) -> object:
+        # Memoize the first parse regardless of the arguments it was called
+        # with, so calling parse() again returns the same value rather than
+        # re-parsing. This is deliberate: one raw response backs one span, and
+        # re-parsing a stream would consume it twice anyway.
+        if self._self_parsed is not None:
+            return self._self_parsed
```
lmolkova: Calling `parse()` twice on an async response raises. The SDK supports repeated `parse()` (it memoizes internally), but `RawResponseProxy.parse` returns the memoized *parsed value* on the second call, where the async SDK returns a coroutine. The caller awaits a `Message` and gets `TypeError: object Message can't be used in 'await' expression`.

Fails today (cassette from `test_async_messages_with_streaming_response_nonstreaming` replays fine):

```python
@pytest.mark.vcr()
@pytest.mark.asyncio
async def test_async_messages_with_streaming_response_parse_twice(
    span_exporter, async_anthropic_client, instrument_no_content
):
    async with async_anthropic_client.messages.with_streaming_response.create(
        model="claude-sonnet-4-20250514",
        max_tokens=100,
        messages=[{"role": "user", "content": "Say hello in one word."}],
    ) as raw_response:
        first = await raw_response.parse()
        second = await raw_response.parse()

    assert isinstance(first, Message)
    assert isinstance(second, Message)
    assert first.id == second.id
```

The memoized return needs to match the shape the caller expects, so on the async path the repeat call should hand back an awaitable rather than the value.

instrumentation/opentelemetry-instrumentation-genai-anthropic/src/opentelemetry/instrumentation/genai/anthropic/_raw_response.py
```diff
@@ -0,0 +1,275 @@
...
+    return RawResponseProxy(
+        result,
+        dispatch_message=lambda message: MessageWrapper(
+            message, capture_content
+        ).extract_into(invocation),
+        wrap_stream=lambda stream: _wrap_parsed_stream(
+            stream, invocation, capture_content
+        ),
+        finalize=invocation.stop,
+    )
```
lmolkova: A `with_raw_response` caller that never calls `parse()` gets no span at all.

For a non-streaming call the SDK sends the request with httpx `stream=False`, so httpx reads the body to completion and closes the response inside `send()`. By the time `RawResponseProxy.__init__` runs, the response is already consumed, and the two entry points are cleanly distinguishable at that moment:

```
with_raw_response (non-streaming):  LegacyAPIResponse  is_closed=True
with_streaming_response:            APIResponse        is_closed=False
```

So on the first one nothing will ever call `close()` again, the fallback never fires, `_self_parsed` stays `None`, and the invocation is never finalized. (The hook itself is fine: calling `http_response.close()` by hand does produce the span, but no real caller does that.) The unit tests miss it because `_FakeHTTPResponse` starts unclosed, which a real response never is on this path.

Deferring only makes sense while the body is still open. When it is already complete the operation is over, `parse()` is a memoized deserialization over bytes in memory, and we can just finalize:

```suggestion
def wrap_raw_response(
    result: Any,
    invocation: InferenceInvocation,
    capture_content: bool,
) -> Any:
    """Wrap a ``with_raw_response`` / ``with_streaming_response`` result.

    When the SDK has already read the body to completion (the non-streaming
    ``with_raw_response`` path), the operation is over: ``parse()`` is a memoized
    deserialization with nothing left to wait for, so the span is finalized here
    and the caller gets its raw response untouched.

    While the body is still open, the caller owns when it is read, so the result
    is wrapped so its metadata resolves natively and ``parse()`` routes on the
    value it returns: a ``Message`` finalizes the span immediately with response
    telemetry, a ``Stream`` / ``AsyncStream`` becomes an instrumented wrapper
    that finalizes on drain, and a coroutine is awaited before dispatching. When
    the caller never parses, the close fallback finalizes the span once.
    """
    http_response = getattr(result, "http_response", None)
    if getattr(http_response, "is_closed", False):
        try:
            parsed = result.parse()
        except Exception:  # pylint: disable=broad-exception-caught
            # Telemetry must never break the caller; skip response attributes.
            _logger.debug(
                "raw-response parse() failed; skipping response telemetry",
                exc_info=True,
            )
        else:
            if isinstance(parsed, AnthropicMessage):
                MessageWrapper(parsed, capture_content).extract_into(
                    invocation
                )
        invocation.stop()
        return result

    return RawResponseProxy(
        result,
        dispatch_message=lambda message: MessageWrapper(
            message, capture_content
        ).extract_into(invocation),
        wrap_stream=lambda stream: _wrap_parsed_stream(
            stream, invocation, capture_content
        ),
        finalize=invocation.stop,
    )
```

I tried this locally: the existing 119 tests still pass and the test below goes green. Note it also makes `_self_parsing` unnecessary, since that flag only exists to survive the close firing *during* `parse()` on the already-complete path.

Eagerly parsing the open-body case would not work, for the record: reading the body there would break a caller that wanted `iter_bytes()`, which is why this is a split rather than parsing everything up front.

Test that fails today with `assert 0 == 1`:

```python
@pytest.mark.vcr()
def test_sync_messages_with_raw_response_never_parsed(
    span_exporter, anthropic_client, instrument_no_content
):
    raw_response = anthropic_client.messages.with_raw_response.create(
        model="claude-sonnet-4-20250514",
        max_tokens=100,
        messages=[{"role": "user", "content": "Say hello in one word."}],
    )
    assert raw_response.headers is not None

    assert len(spans) == 1
```

instrumentation/opentelemetry-instrumentation-genai-anthropic/src/opentelemetry/instrumentation/genai/anthropic/_raw_response.py
Shailendra005: Fixed in 78556f3. On the async path the repeat `parse()` now returns a fresh awaitable that resolves to the already-parsed value, instead of the memoized `Message` itself, so a second `await raw_response.parse()` no longer raises. Sync repeats still return the value directly. Added your `test_async_messages_with_streaming_response_parse_twice` (both parses return the same `Message`); it fails before the change with the `TypeError` you described and passes after.

instrumentation/opentelemetry-instrumentation-genai-anthropic/src/opentelemetry/instrumentation/genai/anthropic/_raw_response.py
Shailendra005: Fixed in 78556f3, adopting your suggestion. When `http_response.is_closed` is already true (the non-streaming `with_raw_response` path, where httpx read the body to completion in `send()`), `wrap_raw_response` now parses inline, extracts the `Message` telemetry, finalizes the span, and returns the raw result untouched (no proxy). Added your `test_sync_messages_with_raw_response_never_parsed`; it fails before with no span and passes after.

One note on removing `_self_parsing`: it turned out to still be needed, but only for the `with_streaming_response` non-streaming path. There the body is still open at `wrap_raw_response` (`is_closed` is False), so it goes through the proxy, and `parse()` reads and closes the body during the call, which fires the close fallback before dispatch can extract. Removing the flag regressed both `with_streaming_response` non-streaming tests (span finalized with request-only attributes). Your reasoning holds for the `with_raw_response` path you moved to the `is_closed` branch, which no longer needs it, so I kept a minimal guard scoped to the still-open-body parse. Happy to revisit if you'd prefer a different split.

Also cleared the `misc / typecheck` pyright errors (0 now); full suite 120 passed, 3 skipped; ruff clean.

instrumentation/opentelemetry-instrumentation-genai-anthropic/src/opentelemetry/instrumentation/genai/anthropic/_raw_response.py
```diff
@@ -0,0 +1,332 @@
...
+        except Exception:  # pylint: disable=broad-exception-caught
+            _logger.debug(
+                "raw-response parse() failed; skipping response telemetry",
+                exc_info=True,
+            )
+        else:
+            if isinstance(parsed, AnthropicMessage):
+                MessageWrapper(parsed, capture_content).extract_into(
+                    invocation
+                )
```
lmolkova: This reintroduces the bug the PR is named after, on the primary `with_raw_response` path.

`extract_into()` sits in the `else` of the `try`/`except`, so it is unguarded: if extraction raises, the exception propagates out of `wrap_raw_response`, gets caught by `traced_method`, and is re-raised into the caller — after the API call already succeeded.

Fails today, with `ValueError: boom` escaping to the caller:

```python
@pytest.mark.vcr()
def test_sync_messages_raw_response_extraction_failure_not_raised(
    span_exporter, anthropic_client, instrument_no_content, monkeypatch
):
    def boom(self, invocation):
        raise ValueError("boom")

    monkeypatch.setattr(
        _raw_response.MessageWrapper, "extract_into", boom, raising=True
    )

    raw_response = anthropic_client.messages.with_raw_response.create(**KW)
    assert raw_response.parse().id
```

Moving the `isinstance` check and `extract_into()` inside the `try` fixes it. `_dispatch` already guards its equivalent call for exactly this reason.

instrumentation/opentelemetry-instrumentation-genai-anthropic/src/opentelemetry/instrumentation/genai/anthropic/_raw_response.py
```diff
@@ -0,0 +1,332 @@
...
+        except BaseException:
+            # The caller's own ``parse()`` failed; let the close fallback
+            # finalize the span and propagate the caller's exception unchanged.
+            self._self_parsing = False
+            raise
+        if inspect.iscoroutine(parsed):
+            # Async client: ``parse()`` is a coroutine and the body is only read
+            # when it is awaited, so keep suppressing until the awaitable runs.
+            self._self_async_parse = True
+            return self._dispatch_awaitable(parsed)
```
lmolkova: An async `parse()` coroutine that is never awaited leaves `_self_parsing = True` forever, which permanently suppresses the close fallback, so the span never ends. Callers do abandon coroutines, e.g. on an early return or a cancelled task.

Fails today with `assert 0 == 1`:

```python
@pytest.mark.vcr()
@pytest.mark.asyncio
async def test_async_messages_with_streaming_response_parse_never_awaited(
    span_exporter, async_anthropic_client, instrument_no_content
):
    async with async_anthropic_client.messages.with_streaming_response.create(
        **KW
    ) as raw_response:
        coro = raw_response.parse()
        coro.close()

```

The flag guards against the close firing mid-parse, but nothing resets it when the parse never runs. Finalizing from the proxy's own close / `__aexit__` rather than relying on the suppression flag would avoid the whole class of leak.

instrumentation/opentelemetry-instrumentation-genai-anthropic/src/opentelemetry/instrumentation/genai/anthropic/_raw_response.py
Shailendra005: Fixed in fe11cf4. Moved the `isinstance` + `extract_into()` inside the `try`, so an extraction failure is swallowed and logged at debug just like a `parse()` failure, matching `_dispatch`'s guard. The raw response is returned untouched and the caller is never broken. Added your `test_sync_messages_raw_response_extraction_failure_not_raised` (monkeypatches `extract_into` to raise, asserts `raw_response.parse().id`); it fails before with `ValueError: boom` escaping and passes after.

instrumentation/opentelemetry-instrumentation-genai-anthropic/src/opentelemetry/instrumentation/genai/anthropic/_raw_response.py
Shailendra005: Fixed in fe11cf4. The flag now only lives across an actual body read, so it can't get stranded. `parse()` holds it around the synchronous `parse()` call in a `try/finally` (the sync body read, whose close fires mid-parse), and releases it before handing back the async coroutine, which reads nothing until awaited. `_dispatch_awaitable` re-arms it immediately before `await` and resets it in its `finally`. So a coroutine that is created but never awaited never sets the flag, and the `aclose` fallback finalizes the span normally on `__aexit__`, exactly as you wanted finalization to come from close rather than a suppression flag that can stick. Added your `test_async_messages_with_streaming_response_parse_never_awaited` (creates the coroutine, `coro.close()`, asserts one finished span); it fails before and passes after.

typecheck is 0, full suite 122 passed / 3 skipped, ruff clean.

lmolkova:
@Shailendra005 thanks for all the fixes! there are still a few problems and test gaps left, I've sent https://github.com/Shailendra005/opentelemetry-python-genai/pull/1 against your branch to cover them. PTAL

Shailendra005:
Thanks so much @lmolkova, this is a much better design. Merged your `anthropic-raw-response-review` into the branch, so it is now part of this PR. Letting the response body drive finalization (instead of deferring everything to `parse()`) is clearly the right call, and it closes the abandoned-stream, `parse(to=...)`, and read-without-parse gaps cleanly. Agreed on dropping the extraction catch-all too: the mappers already tolerate `None`, so a raise there is a real bug that should be loud.

Validated locally: pyright 0, ruff check/format clean, full anthropic suite green (the `#389` case is the skipped test as you noted). Happy to take the `MessagesWithStreamingResponse.create` patch-target follow-up for #389 as a separate PR if that is useful.

PR #363, opened 2026-08-07
adding crew ai instrumentation

venkata-pavani:
# Description

## Description

| What I changed | What this means |
| --- | --- |
| Added CrewAI LLM call instrumentation | CrewAI LLM calls now produce telemetry that follows the OpenTelemetry GenAI semantic conventions. |
| Added GenAI attributes | The telemetry includes available model, provider, request settings, messages, tool definitions, response details, token usage, and error information. |
| Added spans and metrics | The instrumentation emits inference spans, client duration metrics, and token usage metrics through `opentelemetry-util-genai`. |
| Added success and failure handling | Successful calls complete their telemetry, while failed calls record and re-raise the original error without changing CrewAI's behavior. |
| Disabled CrewAI native telemetry by default | This prevents two unrelated sets of telemetry from being emitted while this instrumentation is active. |
| Allowed users to retain CrewAI native telemetry | Users can set `CREWAI_DISABLE_TELEMETRY=false`. Existing values are preserved and restored correctly. ref: https://github.com/open-telemetry/opentelemetry-python-genai/pull/308#discussion_r3696976573 |
| Added completion-hook support | Captured prompts and completions can be forwarded using the shared completion-hook mechanism. |
| Added tests | Tests cover completed calls, failed calls, content capture, completion hooks, semantic-convention attributes, and native telemetry configuration. |
| Updated documentation and examples | The README explains the new behavior and includes a custom completion-hook example. |

## Type of change

Please delete options that are not relevant.
- [ ] New feature (non-breaking change which adds functionality)
- [ ] This change requires a documentation update

# Future PR 

In follow-up PRs, I plan to add CrewAI workflow, agent, and tool instrumentation using the `invoke_workflow`, `invoke_agent`, and `execute_tool` GenAI semantic conventions, along with proper context propagation and streaming support.

PR #384, opened 2026-08-14
Add CrewAI agent and workflow instrumentation

venkata-pavani:
# Description

I implemented and tested invoke_agent() and invoke_workflow() mechanism. I was able to reproduce and test this via Conformance test , VCR tests and was able confirm this by installing local Grafana tempo test 

# Features 

1. invoke_agent () and invoke_workflow()
##### Grafana Logs

image

image

2. Also added the feature for enabling or disabling native telemetry as mentioned in the the comments below (in a less invasive way) using instrument() and uninstrument()

https://github.com/open-telemetry/opentelemetry-python-genai/pull/308#pullrequestreview-4779945012

https://github.com/open-telemetry/opentelemetry-python-genai/pull/363#discussion_r3742630142

### Why I did above?

I followed the same pattern that we have for OpenAI-agents framework 

https://github.com/open-telemetry/opentelemetry-python-genai/blob/main/instrumentation/opentelemetry-instrumentation-genai-openai-agents/src/opentelemetry/instrumentation/genai/openai_agents/__init__.py

Please let me know if this works 

# Future PR

Planning to implement other features like plan() and execute() for this instrumentation


instrumentation/opentelemetry-instrumentation-genai-crewai/tests/cassettes/failed_kickoff.yaml
```diff
@@ -0,0 +1,38 @@
+# TODO: this is generated by AI, re-record
```
xrmx, line 1: 👀

instrumentation/opentelemetry-instrumentation-genai-crewai/README.rst
```diff
@@ -1,9 +1,33 @@
...
+CrewAI delegates model calls to LLM client libraries. This instrumentation does
+not emit inference spans for those calls; instrument the underlying LLM client
+to capture model-call telemetry without duplicate spans.
+
+The instrumentation adds its own handlers to CrewAI's event bus. It does not
+disable CrewAI telemetry, change CrewAI environment variables, or remove event
+handlers registered by the application or other integrations.
+
+To emit only OpenTelemetry GenAI telemetry, users can disable CrewAI's native
+telemetry before importing CrewAI while leaving this instrumentation enabled::
```
xrmx, line 17:
```suggestion
telemetry before importing CrewAI while leaving this instrumentation enabled:

::
```

instrumentation/opentelemetry-instrumentation-genai-crewai/src/opentelemetry/instrumentation/genai/crewai/event_listener.py
```diff
@@ -0,0 +1,282 @@
...
+        ):
+            message = _output_message(event.output)
+            if message is not None:
+                invocation.output_messages = [message]
+            invocation.stop()
+        elif isinstance(event, AgentExecutionErrorEvent) and isinstance(
+            invocation, AgentInvocation
+        ):
+            invocation.fail(
+                _error(event.error, "CrewAI agent execution failed")
```
lmolkova: > "CrewAI agent execution failed"

why? we don't need to invent anything that lib does not report

instrumentation/opentelemetry-instrumentation-genai-crewai/src/opentelemetry/instrumentation/genai/crewai/event_listener.py
```diff
@@ -0,0 +1,282 @@
...
+
+def _completion_key(event: BaseEvent) -> str | None:
+    return event.started_event_id
+
+
+def _error(value: object, fallback: str) -> BaseException:
+    return (
+        value
+        if isinstance(value, BaseException)
+        else RuntimeError(str(value or fallback))
```
lmolkova: no need to invent exception type, you can use Error object from utils that has error type and error message - the type of the value is always available and no need for fallback

instrumentation/opentelemetry-instrumentation-genai-crewai/src/opentelemetry/instrumentation/genai/crewai/event_listener.py
```diff
@@ -0,0 +1,282 @@
...
+    )
+
+
+def _input_message(value: object) -> InputMessage | None:
```
lmolkova: is there a more precise type than object?

instrumentation/opentelemetry-instrumentation-genai-crewai/src/opentelemetry/instrumentation/genai/crewai/event_listener.py
```diff
@@ -0,0 +1,282 @@
...
+    if value is None:
+        return None
+    if isinstance(value, str):
+        content = value
+    else:
+        try:
+            content = json.dumps(value, default=str)
```
lmolkova: this does not look right. we need to understand the structure of this value to properly convert it into input message part

instrumentation/opentelemetry-instrumentation-genai-crewai/src/opentelemetry/instrumentation/genai/crewai/event_listener.py
```diff
@@ -0,0 +1,282 @@
...
+        except (TypeError, ValueError):
+            content = str(value)
+    return InputMessage(role="user", parts=[Text(content=content)])
+
+
+def _output_message(value: object) -> OutputMessage | None:
+    if value is None:
+        return None
+    content = value if isinstance(value, str) else str(value)
```
lmolkova, line 76: same here, we should support all modalities - please take a look at how openai anthropic or google-genai do it.

instrumentation/opentelemetry-instrumentation-genai-crewai/src/opentelemetry/instrumentation/genai/crewai/event_listener.py
```diff
@@ -0,0 +1,282 @@
...
+        event_type: type[EventT],
+        handler: Callable[[object, EventT], None],
+    ) -> None:
+        if self._event_bus is None:
+            return
+        registered = self._event_bus.on(event_type)(handler)
+        self._handlers.append((event_type, registered))
+
+    def _remember(self, event: BaseEvent, invocation: GenAIInvocation) -> None:
+        with self._lock:
```
lmolkova, line 116: do we have to have a lock here?

instrumentation/opentelemetry-instrumentation-genai-crewai/src/opentelemetry/instrumentation/genai/crewai/event_listener.py
```diff
@@ -0,0 +1,282 @@
...
+            previous = self._invocations.setdefault(event.event_id, invocation)
```
lmolkova, line 117: do we have to accumulate state? can we instrument without it?

instrumentation/opentelemetry-instrumentation-genai-crewai/src/opentelemetry/instrumentation/genai/crewai/event_listener.py
```diff
@@ -0,0 +1,282 @@
...
+            )
+
+    def _on_crew_started(
+        self, source: object, event: CrewKickoffStartedEvent
+    ) -> None:
+        del source
+        if not is_instrumentation_enabled():
+            return
+        invocation = self._telemetry_handler.workflow(event.crew_name)
+        message = _input_message(event.inputs)
```
lmolkova: why just one? there is clearly more than one input

instrumentation/opentelemetry-instrumentation-genai-crewai/src/opentelemetry/instrumentation/genai/crewai/event_listener.py
```diff
@@ -0,0 +1,282 @@
...
+        self, source: object, event: AgentExecutionStartedEvent
+    ) -> None:
+        del source
+        if not is_instrumentation_enabled():
+            return
+        agent = event.agent
+        invocation = self._telemetry_handler.invoke_local_agent(
+            agent_name=agent.role
+        )
+        invocation.agent_id = str(agent.id)
```
lmolkova, line 246: agent id should not be used for local agents, it's only for remote ones

instrumentation/opentelemetry-instrumentation-genai-crewai/tests/conformance/orchestration.py
```diff
@@ -0,0 +1,93 @@
...
+                )
+            )
+            _emit(
+                CrewKickoffCompletedEvent(
+                    crew_name="research crew",
+                    crew=None,
+                    output="report",
+                    started_event_id=crew.event_id,
+                )
+            )
```
lmolkova, line 93: this is supposed to be top level api like crew kickoff, not hand-made combination of events

lmolkova (changes requested):

instrumentation/opentelemetry-instrumentation-genai-crewai/src/opentelemetry/instrumentation/genai/crewai/event_listener.py
venkata-pavani: hi @lmolkova ,

Instead of using manually written error messages, I now use the error information reported by CrewAI

image

instrumentation/opentelemetry-instrumentation-genai-crewai/src/opentelemetry/instrumentation/genai/crewai/event_listener.py
venkata-pavani: “I replaced object with CrewAI’s actual input type...This was updated in commit https://github.com/open-telemetry/opentelemetry-python-genai/commit/7971eda

instrumentation/opentelemetry-instrumentation-genai-crewai/src/opentelemetry/instrumentation/genai/crewai/event_listener.py
venkata-pavani: Thanks for raising the questions about the lock and accumulated state. 

After reviewing the current design, I agree that we may have to try to avoid **_invocations and _pending_completions**.
The listener currently receives start and completion events through CrewAI’s asynchronously dispatched event bus. Because those callbacks _may execute concurrently or out of order_, the instrumentation stores active invocations, temporarily stores unmatched completion events, and uses a lock to make the lookup/remove operations safe. 

The lock itself  https://github.com/open-telemetry/opentelemetry-python-genai/pull/384#discussion_r3849651652 ; it is a consequence of reconstructing an operation lifecycle from separate events.

I propose instrumenting CrewAI-owned API boundaries instead:

- Wrap Crew.kickoff() and its async equivalent for workflow invocations.
- Wrap the stable public agent execution entry points for agent invocations.
- Create the GenAIInvocation before calling the original method.
- On success, record the result and stop the invocation.
- On failure, fail the invocation with the original exception and re-raise it unchanged.

@lmolkova please let me know your thoughts on this design . Agno PR #590 demonstrates a similar pattern: wrapping SDK-owned operation methods and keeping non-streaming invocations local to each call, while using util-genai stream wrappers to finish streaming invocations.

lmolkova: this seems like a reasonable approach if we can't get what we need from callbacks. I'd recommend exploring what https://github.com/open-telemetry/donation-openinference/blob/main/python/instrumentation/openinference-instrumentation-crewai/src/openinference/instrumentation/crewai and prototyping mokeypatching approach before committing to one or another.

some instrumentations (langchain) accumulate state and it's ok for the first version. So don't consider it as a hard requirement, but make sure to explore other options.

venkata-pavani:
Closing this as I reimplemented it with monkey patching. This avoids the use of locking and monkey patching is the preferred approach as per openinference documentation.

CC: @lmolkova

PR #391, opened 2026-08-15
fix(langchain): emit invoke_agent for create_agent, including nested agents

AgentGymLeader:
## What

Emits invoke_agent spans for LangChain create_agent graphs, including nested agents, which now get their own spans.

Callback metadata cannot distinguish a nested agent from its caller: LangChain merges the forwarded config over the inner agent's own, so the inherited fields describe the outer agent. create_agent root classification therefore no longer relies on callback metadata alone. Pregel.stream / astream are patched so a compiled graph announces itself on a context stack, and the callback handler claims the top unclaimed announcement when a root run starts. An inherited agent_name matching an enclosing agent is treated as inherited config, so deeply nested agents keep their own declared names. Deprecated create_react_agent graphs compile without the marker, so they are detected at runtime by inspecting the agent node.

## Tests

Classification tests drive real invocations through instrument() against a real tracer provider and assert on exported spans: names and parent/child nesting. Sync nested, three-level and renamed cases pass. The async nested case is a strict xfail exposing the pre-existing context-propagation gap above. create_agent tests run on the latest supported LangChain; the oldest supported version predates create_agent, so those tests skip there.


instrumentation/opentelemetry-instrumentation-genai-langchain/tests/test_operation_mapping.py
```diff
@@ -172,6 +172,64 @@ def test_otel_agent_span_true_is_agent(self):
         )
         assert result == OperationName.INVOKE_AGENT
 
+    def test_top_level_langchain_agent_is_agent(self):
```
lmolkova: can you please add conformance test or update existing one to expect agent spans?

instrumentation/opentelemetry-instrumentation-genai-langchain/src/opentelemetry/instrumentation/genai/langchain/operation_mapping.py
```diff
@@ -96,14 +97,30 @@ def resolve_agent_name(
...
         or metadata.get(_META_AGENT_NAME)
         or metadata.get(_META_AGENT_TYPE)
+    ):
+        return True
+
+    # Compare only the callback's explicit name; resolved names may fall back
+    # to an inherited langgraph_node and misclassify a nested agent.
+    langchain_agent_name = metadata.get(_META_LANGCHAIN_AGENT_NAME)
+    return bool(
+        langchain_agent_name
```
lmolkova: `lc_agent_name` is only present in metadata when `create_agent(name=...)` is passed (`langchain/agents/factory.py` guards on `if name:`). The default `create_agent(model, tools)` emits only `{'ls_integration': 'langchain_create_agent'}`, so it still isn't classified as an agent - at top level it becomes `invoke_workflow LangGraph`, and nested it gets no span at all. `ls_integration == "langchain_create_agent"` is always present and is the reliable marker.

Fails on this branch (langchain 1.3.14 / langgraph 1.2.9):

```python
def test_unnamed_create_agent_emits_agent_span(spans):
    agent = create_agent(FakeModel(), [noop])
    agent.invoke({"messages": [("user", "hi")]})
    assert [s.name for s in agent_spans(spans)] == ["invoke_agent"]
# 0 agent spans; only "invoke_workflow LangGraph"
```

Same result when the unnamed agent is invoked from inside a graph node: 0 agent spans.

instrumentation/opentelemetry-instrumentation-genai-langchain/src/opentelemetry/instrumentation/genai/langchain/operation_mapping.py
```diff
@@ -96,14 +97,30 @@ def resolve_agent_name(
...
+        and (
+            LANGGRAPH_NODE_KEY not in metadata
+            or explicit_run_name is None
+            or str(explicit_run_name) == str(langchain_agent_name)
```
lmolkova: A `run_name` override defeats this comparison. LangChain's own subagent rule is name-independent - a nested run whose `lc_agent_name` differs from the parent namespace's. The handler already tracks parent runs, so that comparison is available and more robust.

Fails on this branch:

```python
agent = create_agent(FakeModel(), [noop], name="planner_agent")
# invoked from inside a graph node named "planner":
agent.invoke({"messages": [...]}, {"run_name": "step1"})
# name='step1', langgraph_node='planner', lc_agent_name='planner_agent' -> 0 agent spans
```

instrumentation/opentelemetry-instrumentation-genai-langchain/src/opentelemetry/instrumentation/genai/langchain/operation_mapping.py
```diff
@@ -96,14 +97,30 @@ def resolve_agent_name(
...
+    if (
         metadata.get(_META_AGENT_SPAN)
         or metadata.get(_META_AGENT_NAME)
         or metadata.get(_META_AGENT_TYPE)
+    ):
+        return True
+
+    # Compare only the callback's explicit name; resolved names may fall back
+    # to an inherited langgraph_node and misclassify a nested agent.
+    langchain_agent_name = metadata.get(_META_LANGCHAIN_AGENT_NAME)
```
lmolkova: `resolve_agent_name` never consults `lc_agent_name`, so a run detected as an agent here gets named from `kwargs["name"]` - the run name, not the agent name. Its `langgraph_node` fallback also feeds the ancestor dedup in `callback_handler.py`.

Fails on this branch:

```python
agent = create_agent(FakeModel(), [noop], name="my_agent")
RunnableLambda(lambda _: agent.invoke(..., {"run_name": "custom"})).invoke({})
# gen_ai.agent.name == "custom", expected "my_agent"
```

AgentGymLeader:
@lmolkova Pushed 42aaaf4.

- Conformance: updated the existing scenario. It uses create_agent now and expects invoke_agent, green under weaver 0.25.1.
- ls_integration: agreed, that's the marker. A run is a root when it carries it and no ancestor does, so the default form emits invoke_agent now.
- run_name override: dropped the name comparison. It compares against the nearest marked ancestor now, using the parent runs the handler already tracks.
- resolve_agent_name: checks lc_agent_name before the run name now.

Not fixed: agents nested inside another agent I couldn't do from callback metadata; the measurement and three options are in the PR body.

I resolved your four threads and the two Copilot ones. Reopen anything I closed too early.

instrumentation/opentelemetry-instrumentation-genai-langchain/src/opentelemetry/instrumentation/genai/langchain/callback_handler.py
```diff
@@ -70,8 +70,21 @@ def on_chain_start(
         metadata: dict[str, Any] | None = None,
         **kwargs: Any,
     ) -> Any:
+        parent_agent = self._find_nearest_agent(parent_run_id)
+        has_create_agent_marker = bool(
+            metadata
+            and metadata.get("ls_integration") == "langchain_create_agent"
```
lmolkova: Please reuse the constants from `operation_mapping` (`_META_LANGCHAIN_INTEGRATION` / `_LANGCHAIN_CREATE_AGENT`) instead of hardcoding the strings here - they're already defined there, let's keep following that pattern.

Better yet, expose a small helper in `operation_mapping` (e.g. `has_create_agent_marker(metadata)`) and call it from both places. Right now the marker set is encoded in two spots: if it ever grows, classification and ancestry marking silently drift apart and nested agents start emitting spurious spans.

instrumentation/opentelemetry-instrumentation-genai-langchain/src/opentelemetry/instrumentation/genai/langchain/operation_mapping.py
```diff
@@ -190,6 +215,8 @@ def classify_chain_run(
     metadata: dict[str, Any] | None,
     kwargs: dict[str, Any],
     parent_run_id: UUID | None = None,
+    parent_agent_name: str | None = None,
+    create_agent_ancestry: bool | None = False,
```
lmolkova: The default should be `None`, not `False`. `False` is the permissive value ("this run is a create_agent root"), so any call site that forgets to pass the argument will misclassify nested agents as roots. `None` (unknown) is the safe default.

instrumentation/opentelemetry-instrumentation-genai-langchain/src/opentelemetry/instrumentation/genai/langchain/operation_mapping.py
```diff
@@ -96,14 +104,31 @@ def resolve_agent_name(
...
+    if metadata.get(_META_LANGCHAIN_INTEGRATION) == _LANGCHAIN_CREATE_AGENT:
+        return create_agent_ancestry is False
+
+    langchain_agent_name = metadata.get(_META_LANGCHAIN_AGENT_NAME)
+    return bool(
+        langchain_agent_name
+        and (
+            parent_agent_name is None
+            or str(parent_agent_name) != str(langchain_agent_name)
+        )
```
lmolkova: This branch looks unreachable with real LangChain: `lc_agent_name` is only set by `create_agent`, which also sets the `ls_integration` marker, so the check above always returns first. If it's meant to cover some other producer, please add a test that exercises it; otherwise drop it.

If it stays: the comparison is case-sensitive, while the dedup in `callback_handler` compares agent names case-insensitively, so the two disagree.

instrumentation/opentelemetry-instrumentation-genai-langchain/src/opentelemetry/instrumentation/genai/langchain/callback_handler.py
lmolkova: Please reuse the constants from `operation_mapping` (`_META_LANGCHAIN_INTEGRATION` / `_LANGCHAIN_CREATE_AGENT`) instead of hardcoding the strings here - they're already defined there, let's keep following that pattern.

Better yet, expose a small helper in `operation_mapping` (e.g. `has_create_agent_marker(metadata)`) and call it from both places. Right now the marker set is encoded in two spots: if it ever grows, classification and ancestry marking silently drift apart and nested agents start emitting spurious spans.

instrumentation/opentelemetry-instrumentation-genai-langchain/src/opentelemetry/instrumentation/genai/langchain/operation_mapping.py
lmolkova: The default should be `None`, not `False`. `False` is the permissive value ("this run is a create_agent root"), so any call site that forgets to pass the argument will misclassify nested agents as roots. `None` (unknown) is the safe default.

instrumentation/opentelemetry-instrumentation-genai-langchain/src/opentelemetry/instrumentation/genai/langchain/operation_mapping.py
lmolkova: This branch looks unreachable with real LangChain: `lc_agent_name` is only set by `create_agent`, which also sets the `ls_integration` marker, so the check above always returns first. If it's meant to cover some other producer, please add a test that exercises it; otherwise drop it.

If it stays: the comparison is case-sensitive, while the dedup in `callback_handler` compares agent names case-insensitively, so the two disagree.

instrumentation/opentelemetry-instrumentation-genai-langchain/src/opentelemetry/instrumentation/genai/langchain/operation_mapping.py
```diff
@@ -96,14 +104,31 @@ def resolve_agent_name(
...
-    return bool(
+    if (
         metadata.get(_META_AGENT_SPAN)
         or metadata.get(_META_AGENT_NAME)
         or metadata.get(_META_AGENT_TYPE)
+    ):
+        return True
+
+    if metadata.get(_META_LANGCHAIN_INTEGRATION) == _LANGCHAIN_CREATE_AGENT:
+        return create_agent_ancestry is False
```
lmolkova: Bigger-picture concern about the approach, and I'd like it addressed in this PR rather than after.

Callback metadata cannot identify a `create_agent` root. When an agent is invoked from inside another agent's tool with the config forwarded, LangChain merges the caller's config over the inner agent's own, so `lc_agent_name` and `ls_integration` here describe the **outer** agent. That is why this PR has to reason about ancestry at all, and why nested agents end up unsupported. I checked every other field on the callback (`serialized` is `None`, `langgraph_checkpoint_ns` does not deepen at the nested root, `tags` and `kwargs["name"]` do not separate an agent from any other named runnable), and there is no signal that distinguishes them. LangSmith hits the same wall and works around it by reading pregel task events instead of callbacks, accepting a documented false positive.

The compiled graph's *own* bound config is never shadowed, though:

```python
create_agent(model, tools, name="inner").config
# {'metadata': {'ls_integration': 'langchain_create_agent', 'lc_agent_name': 'inner'}, ...}
```

So patching `Pregel.stream`/`astream` lets the graph announce itself on a context stack that the handler claims when the root run starts. That is ground truth instead of inference, and it removes the need for the ancestry heuristics entirely.

I put together a prototype: https://github.com/lmolkova/opentelemetry-python-genai/commit/81515bf6811da59e8f5c02e30492da2bbd186b93

With it, three-level nesting, same-named nested agents, async, streaming, unnamed agents and agents invoked without config forwarding all layer correctly, with no false positives for a named `RunnableLambda` or a plain subgraph invoked from a tool. It also deletes `create_agent_ancestry`, the `has_create_agent_marker` plumbing and these `ls_integration` / `lc_agent_name` branches, so it is close to net-neutral in size.

Please fold this into this PR. Merging the metadata-inference version first means we delete most of it in the follow-up, and the ancestry logic changes classification behavior in the meantime, so we would be shipping a regression risk for code we already plan to remove.

Note the prototype is exactly that - the tests for the removed inference still need dropping and the corpus tests reworking (separate comment).

instrumentation/opentelemetry-instrumentation-genai-langchain/tests/test_agent_classification_corpus.py
```diff
@@ -0,0 +1,257 @@
...
+        return self
+
+
+@tool
+def noop() -> str:
+    """Do nothing."""
+    return "ok"
+
+
+def _handler() -> tuple[OpenTelemetryLangChainCallbackHandler, mock.MagicMock]:
```
lmolkova, line 94: These tests are at the wrong level. They construct the callback handler directly and hand it a `MagicMock` telemetry handler, so what they assert is the internal shape of `invoke_local_agent` calls, not the telemetry a user gets. That makes them pass for classifications that produce no usable span, and it makes them break on any refactor of the handler's internals even when the emitted spans are unchanged.

Please drive them through `instrument()` with a real tracer provider and assert on the exported spans - names, parent/child nesting and `gen_ai.agent.name` - the way the other test modules in this package do. The nesting cases in particular are only meaningful as an assertion about span parentage, which a mock cannot show.

AgentGymLeader:
@lmolkova Folding your prototype in, and reworking the corpus tests through `instrument()` against exported spans.

Hadn't clocked that the mock made those tests pass with no span at all.

instrumentation/opentelemetry-instrumentation-genai-langchain/src/opentelemetry/instrumentation/genai/langchain/callback_handler.py
```diff
@@ -103,7 +114,14 @@ def on_chain_start(
                     if agent_invocation_name
                     else None
                 )
-                if suggested_agent_name_lower != agent_invocation_name_lower:
+                # A declared agent is its own layer even when it shares its
+                # parent's name, unless the user renamed it - then the rename
+                # already opened the layer and this run would duplicate it.
+                if (
+                    suggested_agent_name == declared_agent_name
```
lmolkova: A nested agent still gets no span when the outer agent is named via `metadata["agent_name"]` - the inner run inherits that name, so it is neither equal to `declared_agent_name` nor different from the parent's, and this branch drops the layer.

```python
outer = create_agent(model, [delegate], name="outer").with_config(
    metadata={"agent_name": "math_agent"}
)
outer.invoke({"messages": [("user", "start")]})
# execute_tool delegate, invoke_agent math_agent - inner agent missing
```

```suggestion
                    declared_agent_name
```

That restores the layer with all existing tests still passing, but the inner span is then named `math_agent`. The name precedence in `resolve_agent_name` needs the same look: inherited metadata outranks the graph's own declared name.

instrumentation/opentelemetry-instrumentation-genai-langchain/tests/test_agent_classification_corpus.py
```diff
@@ -0,0 +1,249 @@
...
+    builder = StateGraph(dict[str, int])
+    builder.add_node("increment", lambda state: {"value": state["value"] + 1})
+    builder.add_edge(START, "increment")
+    builder.add_edge("increment", END)
+    builder.compile().invoke({"value": 1}, {"callbacks": [handler]})
+
+    telemetry.invoke_local_agent.assert_not_called()
+
+
+@pytest.mark.xfail(
```
lmolkova: This strict xfail asserts nested agents do not work, while `test_nested_agent_gets_its_own_span_under_the_calling_tool` asserts they do. It only fails because it never starts the instrumentor - add `start_instrumentation` and it XPASSes. Drop it.

instrumentation/opentelemetry-instrumentation-genai-langchain/src/opentelemetry/instrumentation/genai/langchain/agent_context.py
```diff
@@ -0,0 +1,177 @@
...
+
+    ``Pregel.invoke`` runs through ``stream``, so this covers both entry points.
+    """
+    name = _agent_name(instance)
+    if name is None:
+        return wrapped(*args, **kwargs)
+    return _announce_at_stream_start(wrapped(*args, **kwargs), name)
+
+
+def wrap_astream(
```
lmolkova, line 131: The async path has no tests - every new test drives `invoke()`. `wrap_astream` is the trickier half (contextvars across `__anext__`), so it needs at least the root, nested and concurrent-`gather` cases the sync side has.

lmolkova (commented):
thanks! a few more minor comments. Please update PR description to describe what PR does, it does not need to describe a history of changes it went through

instrumentation/opentelemetry-instrumentation-genai-langchain/src/opentelemetry/instrumentation/genai/langchain/callback_handler.py
AgentGymLeader: Fixed in 986b3a6. Announced roots now always open their own layer, and resolve_agent_name treats a metadata name matching any enclosing agent as inherited config, so the nested graph's declared name wins — a plain precedence swap would have dropped the outer rename in your repro. Added your repro and a three-level case as tests.

instrumentation/opentelemetry-instrumentation-genai-langchain/tests/test_agent_classification_corpus.py
AgentGymLeader: Dropped.

instrumentation/opentelemetry-instrumentation-genai-langchain/src/opentelemetry/instrumentation/genai/langchain/agent_context.py
AgentGymLeader: Added root, nested and concurrent cases through ainvoke. They surfaced a pre-existing issue: on the async callback path the context token attached at span start isn't propagated, so spans come out without parentage (names are correct). The nested case is a strict xfail documenting it. Happy to file a separate issue for the fix.

instrumentation/opentelemetry-instrumentation-genai-langchain/src/opentelemetry/instrumentation/genai/langchain/callback_handler.py
AgentGymLeader: Fixed in bfa7b59. Announced roots now always open their own layer, and `resolve_agent_name` treats a metadata name matching any enclosing agent as inherited config, so the nested graph's declared name wins — a plain precedence swap would have dropped the outer rename in your repro. Added your repro and a three-level case as tests.

instrumentation/opentelemetry-instrumentation-genai-langchain/tests/test_agent_classification_corpus.py
AgentGymLeader: Dropped.

instrumentation/opentelemetry-instrumentation-genai-langchain/src/opentelemetry/instrumentation/genai/langchain/agent_context.py
AgentGymLeader: Added root, nested and concurrent cases through `ainvoke`. They surfaced a pre-existing issue: on the async callback path the context token attached at span start isn't propagated, so spans come out without parentage (names are correct). The nested case is a strict xfail documenting it. Happy to file a separate issue for the fix.

instrumentation/opentelemetry-instrumentation-genai-langchain/.changelog/391.fixed
```diff
@@ -0,0 +1 @@
+Emit ``invoke_agent`` spans for top-level LangChain ``create_agent`` graph roots without misclassifying known descendants; nested agents remain unsupported.
```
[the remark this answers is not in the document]
AgentGymLeader: Updated — the fragment now says nested agents get their own spans.

instrumentation/opentelemetry-instrumentation-genai-langchain/src/opentelemetry/instrumentation/genai/langchain/__init__.py
```diff
@@ -80,12 +85,50 @@ def _instrument(self, **kwargs: Any):
...
+        except ImportError:
+            pass
+        else:
+            for method in ("stream", "astream"):
+                unwrap(langgraph.pregel.Pregel, method)
+            for symbol in (
+                "create_react_agent",
+                "create_tool_calling_executor",
+            ):
+                unwrap(langgraph.prebuilt, symbol)
```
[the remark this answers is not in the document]
AgentGymLeader: Fixed — `_uninstrument` now catches `AttributeError` alongside `ImportError`.

AgentGymLeader:
@lmolkova Updated the description to just describe what the PR does. The other comments are addressed and resolved too.

instrumentation/opentelemetry-instrumentation-genai-langchain/src/opentelemetry/instrumentation/genai/langchain/operation_mapping.py
```diff
@@ -54,31 +55,67 @@ class OperationName:
...
     if metadata:
         name = metadata.get(_META_AGENT_NAME)
         if name:
-            return str(name)
+            metadata_name = str(name)
+            if not (
+                declared_agent_name
+                and ancestor_agent_names
+                and metadata_name.lower() in ancestor_agent_names
+            ):
```
lmolkova, line 114: Nested agents get a duplicate agent span when the outer agent sets `metadata["agent_name"]` - the inner agent's internal nodes still inherit the outer name and open another layer:

```
invoke_agent math_agent        <- root
  execute_tool delegate
    invoke_agent inner
      invoke_agent math_agent  <- spurious
```

The inherited-name check shouldn't be conditional on `declared_agent_name`.

The nested tests don't catch this because they only check names are unique and roots are roots, never the full span set. This one fails on the branch:

```python
def test_nested_agent_emits_no_extra_agent_spans(
    span_exporter, start_instrumentation
) -> None:
    inner = create_agent(
        FakeModel(responses=[AIMessage(content="inner done")]),
        [noop],
        name="inner",
    )

    @tool
    def delegate(config: RunnableConfig) -> str:
        """Delegate to the inner agent."""
        result = inner.invoke({"messages": [("user", "work")]}, config)
        return str(result["messages"][-1].content)

    create_agent(
        FakeModel(
            responses=[
                AIMessage(
                    content="",
                    tool_calls=[
                        {"name": "delegate", "args": {}, "id": "call-1"}
                    ],
                ),
                AIMessage(content="outer done"),
            ]
        ),
        [delegate],
        name="outer",
    ).with_config(metadata={"agent_name": "math_agent"}).invoke(
        {"messages": [("user", "start")]}
    )

    assert sorted(span.name for span in spans) == [
        "execute_tool delegate",
        "invoke_agent inner",
        "invoke_agent math_agent",
    ]
```

```
E   AssertionError: Left contains one more item: 'invoke_agent math_agent'
```

instrumentation/opentelemetry-instrumentation-genai-langchain/src/opentelemetry/instrumentation/genai/langchain/operation_mapping.py
```diff
@@ -54,31 +55,67 @@ class OperationName:
...
+    if not isinstance(metadata, Mapping):
+        return None
+    typed_metadata = cast("Mapping[str, Any]", metadata)
+    if (
+        typed_metadata.get(_META_LANGCHAIN_INTEGRATION)
+        != _LANGCHAIN_CREATE_AGENT
+    ):
+        return None
+    name = typed_metadata.get(_META_LANGCHAIN_AGENT_NAME)
+    return str(name) if name else LANGGRAPH_IDENTIFIER
```
lmolkova: "LangGraph" isn't an agent name - it's the framework's default graph name. `gen_ai.agent.name` is the name the application gave the agent, and semconv says to leave it unset and name the span plain `invoke_agent` when there isn't one.

The conformance scenario dropped `agent_name`, so it now asserts this placeholder and no longer covers a named agent.

instrumentation/opentelemetry-instrumentation-genai-langchain/src/opentelemetry/instrumentation/genai/langchain/__init__.py
```diff
@@ -80,12 +85,50 @@ def _instrument(self, **kwargs: Any):
...
+                return
+
+        # create_react_agent is deprecated but still ships its own builder that
+        # bakes in none of the create_agent markers, so tag what it returns.
+        for symbol in ("create_react_agent", "create_tool_calling_executor"):
+            try:
+                wrap_function_wrapper(
+                    "langgraph.prebuilt", symbol, wrap_create_react_agent
+                )
+            except (ImportError, AttributeError):
```
lmolkova: Sorry, I sent you down the wrong path here - this came from my prototype. Please don't take my suggestions and prototypes as final solutions, they're sketches of a direction.

Patching the module attribute only works if the app imports `create_react_agent` after `instrument()`. With the normal module-level import - what both examples in this package do - the agent shows up as `invoke_workflow` instead. Same code, different telemetry depending on import order.

Detecting these graphs where `Pregel.stream` already sees them would avoid the ordering problem.

instrumentation/opentelemetry-instrumentation-genai-langchain/src/opentelemetry/instrumentation/genai/langchain/__init__.py
```diff
@@ -80,12 +85,50 @@ def _instrument(self, **kwargs: Any):
...
+                pass
 
     def _uninstrument(self, **kwargs: Any):
         """
         Cleanup instrumentation (unwrap).
         """
         unwrap("langchain_core.callbacks.base.BaseCallbackManager", "__init__")
+        try:
+            import langgraph.prebuilt
+            import langgraph.pregel
```
lmolkova, line 109: Also from my prototype, sorry. If `langgraph.prebuilt` is missing (deprecated for removal in v2), `uninstrument()` leaves `Pregel.stream`/`astream` patched. Unwrap `Pregel` in its own try block first.

instrumentation/opentelemetry-instrumentation-genai-langchain/src/opentelemetry/instrumentation/genai/langchain/agent_context.py
```diff
@@ -0,0 +1,177 @@
...
+
+
+def _pop(entry: _PendingAgent) -> None:
+    """Withdraw ``entry``, tolerating a stack the caller's context no longer owns."""
+    pending = _pending.get()
+    if pending and pending[-1] is entry:
+        _pending.set(pending[:-1])
+
+
+def wrap_stream(
```
lmolkova, line 115: Please add tests that drive invocations directly through `.stream()` and `.astream()` (in addition to `.invoke()` / `.ainvoke()`) to verify generator wrapping and announcement push/pop behavior.

instrumentation/opentelemetry-instrumentation-genai-langchain/src/opentelemetry/instrumentation/genai/langchain/callback_handler.py
```diff
@@ -632,15 +656,30 @@ def on_retriever_error(
...
-                return entity
+                if nearest_agent is None:
+                    nearest_agent = entity
+                if entity.agent_name:
+                    ancestor_agent_names.add(entity.agent_name.lower())
             current = self._invocation_manager.get_parent_run_id(current)
-        return None
+        return nearest_agent, ancestor_agent_names
+
+    def _find_nearest_agent(
```
lmolkova: `_find_nearest_agent` has no production callers and is only kept for internal unit tests. Consider removing it and testing behavior through public APIs instead.

instrumentation/opentelemetry-instrumentation-genai-langchain/src/opentelemetry/instrumentation/genai/langchain/operation_mapping.py
AgentGymLeader: Fixed in 602463b: an inherited `agent_name` matching an ancestor is now always excluded from classification, regardless of `declared_agent_name`. Added your repro as a regression test (`test_nested_agent_emits_no_extra_agent_spans`) asserting the full span set.

instrumentation/opentelemetry-instrumentation-genai-langchain/src/opentelemetry/instrumentation/genai/langchain/operation_mapping.py
AgentGymLeader: Fixed in 602463b: unnamed agents now emit plain `invoke_agent` spans with `gen_ai.agent.name` unset. Added `test_unnamed_create_agent_has_no_placeholder_name`, which asserts the span name is `invoke_agent` rather than `invoke_agent LangGraph`. Good catch on the conformance scenario — the same commit restores a named-agent scenario alongside the unnamed one.

instrumentation/opentelemetry-instrumentation-genai-langchain/src/opentelemetry/instrumentation/genai/langchain/__init__.py
AgentGymLeader: Fixed in 602463b: dropped the builder patch; ReAct graphs are classified at runtime in `Pregel.stream`/`astream`, so a module-level import of `create_react_agent` before `instrument()` is classified as `invoke_agent` rather than `invoke_workflow`. Added `test_preimported_create_react_agent_is_classified_at_runtime`, and a counterproof (`test_plain_state_graph_with_agent_node_is_not_an_agent`) that a plain StateGraph with an `agent` node is not classified as an agent.

instrumentation/opentelemetry-instrumentation-genai-langchain/src/opentelemetry/instrumentation/genai/langchain/__init__.py
AgentGymLeader: Fixed in 602463b: `Pregel` unwrap runs in its own try block now, and `test_uninstrument_restores_pregel_when_prebuilt_is_missing` verifies `stream`/`astream` are restored when `langgraph.prebuilt` is missing.

lmolkova (approved):

instrumentation/opentelemetry-instrumentation-genai-langchain/src/opentelemetry/instrumentation/genai/langchain/agent_context.py
AgentGymLeader: Done in 6c7ad79: direct .stream()/.astream() root tests, plus sync/async edge cases (exception before first item, never iterated, close after first item) checking the pending stack ends empty.

instrumentation/opentelemetry-instrumentation-genai-langchain/src/opentelemetry/instrumentation/genai/langchain/callback_handler.py
AgentGymLeader: Removed in 6c7ad79. Coverage now goes through on_chain_start (named child under a non-agent parent; same-name grandchild dedup through an unnamed parent). Fixed the two comments that referenced it.

PR #397, opened 2026-08-16
fix(anthropic): preserve caller errors in streaming responses

### LINKED ISSUE #389: anthropic: caller exception inside a with_streaming_response block records a successful span
### What happens

An exception raised by the caller inside a `with_streaming_response` block finalizes the span as a **success** — no `error.type`, status `UNSET`:

```python
with client.messages.with_streaming_response.create(..., stream=True) as raw:
    for _ in raw.parse():
        raise ValueError("boom")
```

`messages.create(stream=True)` and `messages.stream(...)` both record `error.type` for the equivalent caller-side error, so the raw-response path is inconsistent with the rest of the package.

### Why

The SDK's `ResponseContextManager.__exit__` (and `AsyncResponseContextManager.__aexit__`) receives the caller's exception and discards it, calling `response.close()` with no `exc_info`:

```python
def __exit__(self, exc_type, exc, exc_tb) -> None:
    if self.__response is not None:
        self.__response.close()
```

The instrumentation wraps the *response*, not the context manager, so it only ever observes a plain close and cannot distinguish success from failure.

### Fix direction

Instrument `MessagesWithStreamingResponse.create` / `AsyncMessagesWithStreamingResponse.create` and wrap the returned `ResponseContextManager`, so its `__exit__` sees the real `exc_info` and can fail the invocation.

alexliluz:
Fixes #389

Anthropic's `with_streaming_response` context manager closes the raw response without forwarding caller exceptions, so an error raised while consuming a stream is recorded as a successful span. Wrap the dynamically-created sync and async streaming methods and pass the exception to the active raw-response stream wrapper before SDK cleanup.

Tests: targeted sync/async regression tests pass on the latest and oldest supported environments; pyright and ruff pass. The full suite still has five pre-existing timeout-classification failures in this environment.

AI disclosure: OpenAI Codex (GPT-5) assisted with repository inspection, implementation, and test execution; the changes and results were manually reviewed.


instrumentation/opentelemetry-instrumentation-genai-anthropic/src/opentelemetry/instrumentation/genai/anthropic/wrappers.py
```diff
@@ -96,6 +134,64 @@ def __getattr__(self, name: str):
         return getattr(self._response, name)
 
 
+class MessagesStreamingResponseContextManagerWrapper(Generic[ResponseT]):
```
lmolkova: This is a plain class, not a transparent proxy, so instrumentation changes the type the SDK returns (repo rule: "Don't change the returned type"). Fails today:

```python
from anthropic._response import ResponseContextManager

cm = client.messages.with_streaming_response.create(...)
assert isinstance(cm, ResponseContextManager)  # AssertionError
```

Suggest patching `ResponseContextManager.__exit__` / `AsyncResponseContextManager.__aexit__` at class level instead: keeps the SDK type intact, and drops both wrapper classes and the two `__init__` patch targets. Please add a test asserting the returned context manager's type is unchanged.

instrumentation/opentelemetry-instrumentation-genai-anthropic/src/opentelemetry/instrumentation/genai/anthropic/_raw_response.py
```diff
@@ -105,6 +105,15 @@ def __init__(
         self._self_stream_wrapper: Any = None
         self._install_hooks(raw_response)
 
+    def _fail(self, error: BaseException) -> None:
```
lmolkova, line 111: This calls the subclass hook directly, bypassing the base stream wrapper's `_self_finalized` guard. It only works because `MessagesStreamWrapper._fail` keeps a second, independent guard (`_self_message_telemetry_finalized`) — two guards that must stay in sync. Route through the base finalization instead:

```python
self._self_stream_wrapper._finalize_failure(error)
```

instrumentation/opentelemetry-instrumentation-genai-anthropic/tests/test_sync_messages.py
```diff
@@ -1773,14 +1773,6 @@ def boom(*args, **kwargs):
...
-    reason="Known gap, tracked in #389: the SDK's "
-    "ResponseContextManager.__exit__ discards the caller's exception before "
-    "closing the response, so the proxy never sees it and the span is "
-    "finalized as a success. Fixing it means instrumenting "
-    "MessagesWithStreamingResponse.create and wrapping the context manager "
-    "itself, which is a new patch target and out of scope here."
-)
 @pytest.mark.vcr()
 @pytest.mark.cassette("test_sync_messages_create_streaming_with_raw_response")
 def test_sync_messages_with_streaming_response_user_exception(
```
lmolkova, line 1848: Coverage is just the two un-skipped tests. Please add (sync + async):

- context manager type transparency (see the `wrappers.py` comment)
- `instrument()` / `uninstrument()` round-trip on `with_streaming_response`
- caller exception raised *after* the stream is fully drained, and *before* `parse()` — both must finalize once, with the right `error.type`
- non-streaming `with_raw_response` + caller exception

lmolkova (approved):

lmolkova:
@alexliluz can you please merge main and resolve conflicts?

PR #403, opened 2026-08-17
[opentelemetry-instrumentation-genai-smolagents] instrument agent runs

alexander-akhmetov:
**Description**

Adds invoke_agent instrumentation and the duration metric for smolagents `MultiStepAgent.run`, streaming and non-streaming. The span records the agent name and description, the model, tool and managed-agent definitions, the finish reason, and the task and final answer when content capture is enabled.

**Known gaps:**

- Agent steps do not get separate spans. See https://github.com/open-telemetry/semantic-conventions-genai/issues/81.
- `CodeAgent` with the default `executor_type="local"` writes code and runs it in a worker thread. OpenTelemetry context isn't copied into the thread, so a managed agent called from that code cannot see the parent span.

Part of #141

**Type of change**

**How has this been tested?**

```sh
uv run tox -e py310-test-instrumentation-genai-smolagents-oldest
uv run tox -e py314-test-instrumentation-genai-smolagents-latest
uv run tox -e py314-test-instrumentation-genai-smolagents-conformance
uv run tox -e precommit
uv run tox -e typecheck
uv run tox -e readme
```

**Checklist**

- [x] Unit tests added
- [x] Documentation updated


instrumentation/opentelemetry-instrumentation-genai-smolagents/src/opentelemetry/instrumentation/genai/smolagents/patch.py
```diff
@@ -388,11 +421,196 @@ def wrapper(
...
+    # The base wrapper finalizes on StopIteration and on an Exception, but not
+    # on an interrupt, which would leave the span open.
+    def __next__(self) -> _RunStreamChunk:
+        try:
+            return super().__next__()
+        except BaseException as error:
+            self._finish_once(error)
+            raise
+
+    def close(self) -> None:
```
lmolkova, line 495: A `run(stream=True)` generator that isn't drained never finishes the span, and its context stays attached - so every later span on that thread nests under the leaked `invoke_agent`. GC closes the underlying generator, but nothing calls the wrapper's `close()`.

`break`ing out of `for step in agent.run(task, stream=True):` is a normal way to use this API, so this is easy to hit. Both of these fail on this branch:

```python
def _first_step_only(agent: CodeAgent) -> None:
    for _ in agent.run("Test question", stream=True):
        break


def test_abandoned_streaming_run_finalizes_the_span(
    instrument_with_content, span_exporter
) -> None:
    agent = CodeAgent(tools=[], model=FakeCodeModel(), max_steps=3)
    _first_step_only(agent)
    gc.collect()

    assert (
        len(
            spans_by_operation(
                span_exporter.get_finished_spans(), "invoke_agent"
            )
        )
        == 1
    )


def test_abandoned_streaming_run_does_not_leak_context(
    instrument_with_content, span_exporter, tracer_provider
) -> None:
    agent = CodeAgent(tools=[], model=FakeCodeModel(), max_steps=3)
    _first_step_only(agent)
    gc.collect()

    with tracer_provider.get_tracer("test").start_as_current_span("unrelated"):
        pass

    (unrelated,) = [
        span
        for span in span_exporter.get_finished_spans()
        if span.name == "unrelated"
    ]
    assert unrelated.parent is None
```

```
E  AssertionError: assert 0 == 1
E   +  where 0 = len([])
E   +    where [] = spans_by_operation((), 'invoke_agent')

E  assert SpanContext(trace_id=0xba5e..., span_id=0x68a7..., ...) is None
E   +  where SpanContext(...) = <ReadableSpan 'unrelated'>.parent
```

The context leak is the worse half: everything the process traces afterwards is silently reparented under a span that never ends. `_ModelStreamWrapper` has the same hole, but a partially consumed run generator is much more reachable. Finalizing on `__del__` here (or in `SyncStreamWrapper`) would close it.

instrumentation/opentelemetry-instrumentation-genai-smolagents/src/opentelemetry/instrumentation/genai/smolagents/_messages.py
```diff
@@ -202,18 +208,21 @@ def to_output_message(output_message: ChatMessage) -> OutputMessage:
...
+
+    ``get_tool_json_schema`` also accepts managed agents despite its ``Tool``
+    annotation.
     """
     try:
-        schema = get_tool_json_schema(tool)
+        schema = get_tool_json_schema(cast("Tool", tool))
         parameters = schema["function"]["parameters"]
-    except Exception:  # pylint: disable=broad-except
+    except BaseException:  # pylint: disable=broad-except
```
lmolkova: This widened from `except Exception` to `except BaseException`, so a Ctrl-C while building tool schemas is swallowed and the tool is silently dropped instead. Every other guard in this package (including the new `_recording`) catches `Exception`. Was there a case that needed the wider catch?

(Copilot raised this and the thread got resolved, but the code still has it.)

instrumentation/opentelemetry-instrumentation-genai-smolagents/src/opentelemetry/instrumentation/genai/smolagents/patch.py
```diff
@@ -388,11 +421,196 @@ def wrapper(
...
+    output: object,
+    *,
+    capture_content: bool,
+) -> None:
+    # Reaching ``max_steps`` returns normally with ``AgentMaxStepsError`` on the
+    # final ``ActionStep``.
+    steps = agent.memory.steps
+    last_step = steps[-1] if steps else None
+    finish_reason = (
+        "length"
```
lmolkova: it's not technically length, but AgentMaxStepsError, wdyt about using "max_steps" as the constant?

instrumentation/opentelemetry-instrumentation-genai-smolagents/src/opentelemetry/instrumentation/genai/smolagents/patch.py
```diff
@@ -388,11 +421,196 @@ def wrapper(
...
+        if isinstance(last_step, ActionStep)
+        and isinstance(last_step.error, AgentMaxStepsError)
```
lmolkova: if last_step.error, but of a different type we should report "error"

instrumentation/opentelemetry-instrumentation-genai-smolagents/src/opentelemetry/instrumentation/genai/smolagents/patch.py
```diff
@@ -48,12 +61,42 @@
...
+)
+
+
+@contextmanager
+def _recording(what: str) -> Iterator[None]:
+    """Suppress extraction errors so telemetry cannot break the call."""
+    try:
+        yield
+    except Exception:  # pylint: disable=broad-except
+        _logger.debug("Failed to record %s", what, exc_info=True)
```
lmolkova: why is this needed? we should not throw from instrumentation code and should not try to catch exceptions from instrumentation code. None of the instrumentations do

instrumentation/opentelemetry-instrumentation-genai-smolagents/src/opentelemetry/instrumentation/genai/smolagents/patch.py
alexander-akhmetov: Fixed: https://github.com/open-telemetry/opentelemetry-python-genai/pull/403/changes/043c38539dd1af8e08ef2864f6dc880780ff87a9

instrumentation/opentelemetry-instrumentation-genai-smolagents/src/opentelemetry/instrumentation/genai/smolagents/patch.py
alexander-akhmetov: Renamed: https://github.com/open-telemetry/opentelemetry-python-genai/pull/403/changes/87cb5f70222231c3a71395e4b8932ee0d610f07a

instrumentation/opentelemetry-instrumentation-genai-smolagents/src/opentelemetry/instrumentation/genai/smolagents/patch.py
alexander-akhmetov: Updated: https://github.com/open-telemetry/opentelemetry-python-genai/pull/403/changes/c8e2feb63f65a37799e5ffc1350ba099a1c268f3

instrumentation/opentelemetry-instrumentation-genai-smolagents/src/opentelemetry/instrumentation/genai/smolagents/patch.py
alexander-akhmetov: Removed it. I added it to avoid telemetry conversion failures affecting the call, but I agree that this is too broad and shouldn't be needed

instrumentation/opentelemetry-instrumentation-genai-smolagents/src/opentelemetry/instrumentation/genai/smolagents/patch.py
```diff
@@ -388,11 +408,200 @@ def wrapper(
...
+        try:
+            super().close()
+        except BaseException as error:
+            self._finish_once(error)
+            raise
+
+    def _process_chunk(self, chunk: _RunStreamChunk) -> None:
+        if isinstance(chunk, FinalAnswerStep):
+            self._self_final_output = chunk.output
+            self._self_saw_final = True
```
lmolkova, line 505: `FinalAnswerStep` is the last item the generator yields, so stopping on it never reaches `_on_stream_end` and the span stays open until the wrapper is collected.

```python
def test_break_on_final_answer_step_finishes_the_span(
    instrument_with_content, span_exporter
) -> None:
    agent = CodeAgent(tools=[], model=FakeCodeModel(), max_steps=3)
    stream = agent.run("Test question", stream=True)
    for step in stream:
        if isinstance(step, FinalAnswerStep):
            break

    (span,) = spans_by_operation(
        span_exporter.get_finished_spans(), "invoke_agent"
    )
    assert attr(span, GenAI.GEN_AI_RESPONSE_FINISH_REASONS) == ("stop",)
```

```
E  ValueError: not enough values to unpack (expected 1, got 0)
```

Everything needed to finalize is already on hand when that chunk arrives, so `_process_chunk` can call `_finalize_success()` instead of waiting for a `StopIteration` that may never come. Callers who stop here currently get no output messages at all, even after GC.

The leak that remains once nothing marks the end is #491.

instrumentation/opentelemetry-instrumentation-genai-smolagents/src/opentelemetry/instrumentation/genai/smolagents/patch.py
```diff
@@ -389,12 +406,204 @@ def wrapper(
...
+            return
+        self._self_finished = True
+        _finish(self._self_agent_invocation, error)
+
+    # The base wrapper finalizes on StopIteration and on an Exception, but not
+    # on an interrupt, which would leave the span open.
+    def __next__(self) -> _RunStreamChunk:
+        try:
+            return super().__next__()
+        except BaseException as error:
```
lmolkova, line 491: `except BaseException as error:` in `__next__` catches `StopIteration` re-raised by `super().__next__()`. Exclude `StopIteration` so normal completion never enters the error path:

```python
    def __next__(self) -> _RunStreamChunk:
        try:
            return super().__next__()
        except StopIteration:
            raise
        except BaseException as error:
            self._finish_once(error)
            raise
```

Also add a test verifying that normally drained streams complete with `UNSET` status and no `error.type`:

```python
def test_streaming_run_completes_without_error(
    instrument_with_content, span_exporter
) -> None:
    agent = CodeAgent(tools=[], model=FakeCodeModel(), max_steps=3)
    list(agent.run("Test question", stream=True))

    (agent_span,) = spans_by_operation(
        span_exporter.get_finished_spans(), "invoke_agent"
    )
    assert agent_span.status.status_code == StatusCode.UNSET
    assert attr(agent_span, error_attributes.ERROR_TYPE) is None
```

alexander-akhmetov: Updated: https://github.com/open-telemetry/opentelemetry-python-genai/pull/403/changes/c472df074576f30c1e02368ab3f41fe30026fc23

instrumentation/opentelemetry-instrumentation-genai-smolagents/src/opentelemetry/instrumentation/genai/smolagents/patch.py
alexander-akhmetov: Updated: https://github.com/open-telemetry/opentelemetry-python-genai/pull/403/changes/38e0254a2fab3d5ac9678e60cbd18eec186df5d7

lmolkova (approved):
LGTM!

PR #411, opened 2026-08-19
ci: simplify test matrix and remove redundant jobs

lmolkova:
- **Python versions**: Test only 3.10 and 3.14 across all packages in tox.ini, dropping 3.11-3.13 and PyPy.
- **Conformance tests**: Standardized to run once on 3.14.
- **Lint matrix**: Removed per-package lint jobs; precommit in misc.yml already covers the entire workspace.

Cuts ~50 jobs per CI run without sacrificing coverage.


tox.ini
```diff
@@ -9,85 +9,66 @@ envlist =
...
     py310-test-instrumentation-genai-anthropic-oldest
-    # No pypy3: jiter (an anthropic dependency) ships no PyPy wheels, so it would need a Rust source build
     py314-test-instrumentation-genai-anthropic-conformance
-    lint-instrumentation-genai-anthropic
 
     ; instrumentation-genai-claude-agent-sdk
-    py3{10,11,12,13}-test-instrumentation-genai-claude-agent-sdk-latest
+    py3{10,14}-test-instrumentation-genai-claude-agent-sdk-latest
     py310-test-instrumentation-genai-claude-agent-sdk-oldest
-    # No pypy3: jiter (an anthropic dependency) ships no PyPy wheels, so it would need a Rust source build
```
aabmass, line 56: I guess people are probably just not using PyPy with these libs, so this seems OK.

From what we've seen in python repos though, things do subtly break in PyPy if we care about it.

.github/workflows/lint.yml
aabmass, line 1: Is ruff lint still being run somehow?

aabmass (approved):

.github/workflows/lint.yml
aabmass: > Lint matrix: Removed per-package lint jobs; precommit in misc.yml already covers the entire workspace.

👍

tox.ini
lmolkova: we only tested it with openai and util, so I imagine we already failed at supporting it. Let's bring it back across the board if anyone asks for it?

AGENTS.md
```diff
@@ -92,15 +92,15 @@ Copy the shape from an existing package — paths in `tox.ini` are repo-root-rel
...
 - **`tox.ini`**:
-  - `envlist`: add `py3{…}-test-instrumentation-genai-<lib>-{oldest,latest}`, the
-    `py3{…}-…-<lib>-conformance` entry, and `lint-instrumentation-genai-<lib>`.
+  - `envlist`: add `py3{10,14}-test-instrumentation-genai-<lib>-latest`,
+    `py310-test-instrumentation-genai-<lib>-oldest`, and the
+    `py314-test-instrumentation-genai-<lib>-conformance` entry.
   - `[testenv] deps`: add the factor-conditional test-requirements lines
     (`<lib>-{oldest,latest,conformance}: -r …/tests/requirements.<factor>.txt` plus
-    `{[testenv]test_deps}` / `{[testenv]pytest_deps}`). Requirements install here — **not**
+    `{[testenv]test_deps}` / `{[testenv]pytest_deps}`). Requirements install here - **not**
```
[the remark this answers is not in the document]
lmolkova: 😡😡😡😡😡😡😡😡😡😡😡😡😡

tox.ini
aabmass: Maybe we should call it out in the changelog? (I know it's not really a change but some users might be surprised)

PR #413, opened 2026-08-19
Migrate CI jobs to CNCF runners

aabmass:
## What does this change do?
Migrates CI workflow jobs from `ubuntu-latest` to CNCF self-hosted runners (`cncf-ubuntu-2-8-x86`). Also updates `scorecard.yml` to use `open-telemetry/shared-workflows` v0.8.0 with `use-cncf-hosted-runner: true`, and adds `.github/actionlint.yaml` for runner label recognition.

## Why?
https://github.com/open-telemetry/community/issues/3622

### DIFF
--- a/.github/actionlint.yaml
+++ b/.github/actionlint.yaml
@@ -1,0 +1,3 @@
+self-hosted-runner:
+  labels:
+    - cncf-ubuntu-2-8-x86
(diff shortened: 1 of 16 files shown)


trask (approved):
I'm a little nervous, but not sure things can get worse, so... 🤞

eternalcuriouslearner (approved):

aabmass:
Thanks. Can always revert

PR #417, opened 2026-08-20
langchain: preserve dict-form output messages

### LINKED ISSUE #388: langchain: dict-form messages raise in on_chain_start and silently drop the span
Passing chat messages as plain dicts, which LangChain accepts everywhere and which its own docs use, raises inside `on_chain_start`. LangChain catches callback exceptions and logs them, so the failure is silent: the workflow or agent span is simply never created, and only the child `chat` spans survive.

## Repro

```python
from langchain.agents import create_agent
from langchain_openai import ChatOpenAI

agent = create_agent(
    model=ChatOpenAI(model="gpt-4o-mini", max_tokens=100),
    tools=[],
    system_prompt="You are a helpful assistant.",
    name="weather_assistant",
)
agent.invoke({"messages": [{"role": "user", "content": "hi"}]})
```

Run under `opentelemetry-instrument`.

**Actual:**

```
WARNING langchain_core.callbacks.manager: Error in
OpenTelemetryLangChainCallbackHandler.on_chain_start callback:
AttributeError("'dict' object has no attribute 'content'")
```

and no chain-level span at all. Replacing the dict with
`HumanMessage(content="hi")` produces the span, so the message form is the only
difference.

**Expected:** the same telemetry either way. A supported input form should not
silently remove a span.

## Why it happens

`on_chain_start` calls `make_input_message(inputs)` (`utils.py`). When
`inputs["messages"]` exists it is passed straight to `to_input_messages`, which
is annotated `Iterable[BaseMessage]` and reads `.content` off each entry. The
cast is unchecked, so a dict entry raises.

`make_input_message` already has a fallback for state without a `messages` key.
The same tolerance is needed per entry: convert dicts with
`langchain_core.messages.convert_to_messages`, or skip entries that are not
`BaseMessage` rather than raising.

`make_output_message` casts identically, so `on_chain_end` is exposed to the
same failure for outputs.

## Context

Found while adding LangChain conformance scenarios in
https://github.com/open-telemetry/semantic-conventions-conformance/pull/33.
Related to #387, which is about the operation the span is classified as; this
one is about the span not existing.

alexliluz:
# Description

LangChain accepts role/content dicts as messages. The input path already normalizes them, but the output path only kept `AIMessage` objects, so workflow and agent spans silently lost dict-form assistant output.

Normalize output messages with LangChain's `convert_to_messages` before building telemetry, and cover both the converter and `on_chain_end` path.

Fixes #388

## Type of change

# How has this been tested?

- [x] `uv run tox -e py312-test-instrumentation-genai-langchain-latest` (237 passed)
- [x] `uv run tox -e py312-test-instrumentation-genai-langchain-oldest` (236 passed, 1 skipped)
- [x] `uv run tox -e typecheck`
- [x] `uv run tox -e precommit`
- [x] `uv run tox -e py314-test-instrumentation-genai-langchain-conformance` (5 existing scenario skips)

# Checklist

- [x] Unit tests added

AI-assisted implementation; I reproduced the missing output on current `main`, reviewed the diff, and ran the checks above locally.


instrumentation/opentelemetry-instrumentation-genai-langchain/tests/test_callback_handler.py
```diff
@@ -1043,6 +1052,31 @@ def test_workflow_output_messages_set_on_chain_end(self):
...
+                ]
+            },
+            run_id=run_id,
+        )
+
+        assigned = workflow_inv.output_messages
+        assert len(assigned) == 1
+        assert assigned[0].role == "assistant"
+        assert assigned[0].parts[0].content == "The final answer is 42"
+
```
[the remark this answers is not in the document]
alexliluz: Good point. Workflows and agents currently share the same `on_chain_end` → `make_last_output_message` path, but explicit coverage for both invocation classifications will guard against future divergence. I added an agent-level dict-output regression test. The latest suite passes 237 tests; the oldest suite passes 236 tests with one existing skip.

instrumentation/opentelemetry-instrumentation-genai-langchain/src/opentelemetry/instrumentation/genai/langchain/utils.py
```diff
@@ -206,19 +206,29 @@ def to_input_messages(
...
 ) -> list[OutputMessage]:
-    """Convert LangChain ``AIMessage`` instances into ``OutputMessage`` s.
+    """Convert LangChain assistant messages into ``OutputMessage`` s.
 
-    Non-``AIMessage`` entries are skipped: only assistant turns are recorded
-    as ``gen_ai.output.messages``. Tool execution results belong on the
-    *input* side of the next inference call, not the output side of the
-    previous one.
+    LangChain-supported message representations, such as role/content dicts,
+    are normalized first. Non-``AIMessage`` entries are skipped: only
```
rads-1996, line 216: The docstring needs to be fixed, the non-"ai messages" entries are skipped should be removed.

instrumentation/opentelemetry-instrumentation-genai-langchain/.changelog/417.fixed
```diff
@@ -0,0 +1 @@
+preserve dict-form assistant messages in workflow and agent output telemetry
```
rads-1996, line 1: nit, new line

instrumentation/opentelemetry-instrumentation-genai-langchain/src/opentelemetry/instrumentation/genai/langchain/utils.py
```diff
@@ -206,19 +206,29 @@ def to_input_messages(
 
 
 def to_output_messages(
-    messages: Iterable[BaseMessage],
+    messages: Iterable[Any],
```
eternalcuriouslearner, line 209: nit(blocking): Can we avoid using Any here?

instrumentation/opentelemetry-instrumentation-genai-langchain/src/opentelemetry/instrumentation/genai/langchain/utils.py
```diff
@@ -206,19 +206,29 @@ def to_input_messages(
...
+    assistant turns are recorded as ``gen_ai.output.messages``. Tool execution
+    results belong on the *input* side of the next inference call, not the
+    output side of the previous one.
     """
+    materialized_messages = list(messages)
+    try:
+        normalized_messages: Iterable[BaseMessage] = convert_to_messages(
+            materialized_messages
+        )
+    except Exception:  # pylint: disable=broad-except
```
eternalcuriouslearner, line 226: nit(blocking): Do we have a test for this? If not can you please add it?

instrumentation/opentelemetry-instrumentation-genai-langchain/src/opentelemetry/instrumentation/genai/langchain/utils.py
```diff
@@ -206,19 +206,29 @@ def to_input_messages(
...
+        normalized_messages = [
+            m for m in materialized_messages if isinstance(m, BaseMessage)
+        ]
```
eternalcuriouslearner, line 229: question(blocking): Can we extract this logic into a function like say

```python
def _normalize_messages(
    messages: Iterable[MessageLikeRepresentation],
) -> list[BaseMessage]:
    materialized = list(messages)

    try:
        return convert_to_messages(materialized)
    except Exception:  # pylint: disable=broad-except
        normalized: list[BaseMessage] = []

        for message in materialized:
            try:
                normalized.extend(convert_to_messages([message]))
            except Exception:  # pylint: disable=broad-except
                continue

        return normalized
```

and use this logic for both `to_output_messages` and `to_input_messages` functions?

eternalcuriouslearner (changes requested):
LGTM!! requested few changes to address few loopholes.

PR #432, opened 2026-08-20
ci: batch post-release updates into single PR for bulk releases

lmolkova:
Consolidates post-release updates (README generation and dev version bumps) into a single PR when running bulk releases, while keeping dedicated flows for individual package patches.

Prevents duplicate PRs per package and fixes the version bump step in bulk release which failed previously due to missing `uv`.

### DIFF
--- a/.github/workflows/release-all.yml
+++ b/.github/workflows/release-all.yml
@@ -101,7 +101,7 @@ jobs:
           packages-dir: dist/
           skip-existing: true
 
-  finalize:
+  github-release:
     needs: [enumerate, publish]
     permissions:
       contents: write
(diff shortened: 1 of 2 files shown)


.github/workflows/release-all.yml
```diff
@@ -128,6 +129,15 @@ jobs:
...
           ref: main
 
+      - name: Install uv
+        uses: astral-sh/setup-uv@20cfd1bf945f4377ade1205e4dbc17946fc9a30d # v10.0.1
+
+      - name: Regenerate instrumentations README
+        run: python ./scripts/generate_instrumentation_readme.py
+
+      - name: Bump repository version to next dev version
+        run: uv run ./scripts/version_utils.py bump --dev
```
trask: The README is generated before the version bump, so this commit pairs a table built from the pre-bump `.dev` versions with the bumped `version.py` files. `misc / generate` regenerates the table on the resulting PR and fails. The same swap is needed in `_release-finalize.yml`, where the regeneration also precedes the gated bump.

```suggestion
      - name: Bump repository version to next dev version
        run: uv run ./scripts/version_utils.py bump --dev

      - name: Regenerate instrumentations README
        run: python ./scripts/generate_instrumentation_readme.py
```

trask (approved):
Lightly filtered AI-generated feedback

aabmass (approved):

PR #439, opened 2026-08-21
[`opentelemetry-instrumentation-genai-dspy`] Add initial setup / boilerplate

DylanRussell:
# Description

Initial boilerplate / setup for the `dspy` instrumentation.

`dspy` is a bit of a tricky case. IMO the only spans this instrumentation should write (that are good matches for sem convs) are `execute_tool` and `inference` spans. 

Open inference instrumentation wraps the various `dspy` primitives (`Module`, `Predict`, `Retrieve`, `Adapter`) and create spans (which aren't in our sem convs..) for each one. What these primitives do doesn't align with any existing sem conv. Wrapping them in spans is interesting and makes some sense, as it describes what the framework is doing internally.. 

There is 1 built in Module `ReAct` (https://dspy.ai/diving-deeper/react/) which triggers what they call an "agentic loop".. so probably that one could be an invoke agent span..

Also `dspy` is yet another python library that wraps inference calls (`adk`, `agno`, and lots of other libraries do this too), so we have the same issue of double instrumentation when the model is also instrumented. The way I think we should resolve this is for the `dspy` instrumentation to check if the corresponding `model` instrumentation is installed and active. If it is, don't instrument the model call.

We can eventually add an env var that allows the user to choose which span they want (or both), but I assume if they have the model level instrumentation enabled they prefer to have the span from that.

## Type of change

# How has this been tested?

Unit tests

# Checklist

- [x] Unit tests added
- [x] Documentation updated

### DIFF
--- a/.github/workflows/changelog.yml
+++ b/.github/workflows/changelog.yml
@@ -38,7 +38,7 @@ jobs:
 
       - name: Ensure no direct changes to CHANGELOG.md
         run: |
-          if [[ $(git diff --name-only FETCH_HEAD -- '**/CHANGELOG.md') ]]
+          if [[ $(git diff --diff-filter=MD --name-only FETCH_HEAD -- '**/CHANGELOG.md') ]]
           then
             echo "CHANGELOG.md files should not be directly modified."
             echo "Please add a changelog fragment file to the affected package's .changelog/ directory instead."
--- a/README.md
+++ b/README.md
@@ -28,6 +28,7 @@ All instrumentations use [opentelemetry-util-genai](./util/opentelemetry-util-ge
 | --------------- | ----------------- | ------- | ------ |
 | [opentelemetry-instrumentation-genai-claude-agent-sdk](./instrumentation/opentelemetry-instrumentation-genai-claude-agent-sdk) | claude-agent-sdk >= 0.1.14 | 1.2b0.dev | skeleton |
 | [opentelemetry-instrumentation-genai-crewai](./instrumentation/opentelemetry-instrumentation-genai-crewai) | crewai >= 1.10.1 | 1.2b0.dev | skeleton |
+| [opentelemetry-instrumentation-genai-dspy](./instrumentation/opentelemetry-instrumentation-genai-dspy) | dspy >= 2.5.0 | 1.2b0.dev | skeleton |
 | [opentelemetry-instrumentation-genai-llama-index](./instrumentation/opentelemetry-instrumentation-genai-llama-index) | llama-index-core >= 0.14.19 | 1.2b0.dev | skeleton |
 | [opentelemetry-instrumentation-genai-weaviate-client](./instrumentation/opentelemetry-instrumentation-genai-weaviate-client) | weaviate-client >= 3.0.0, <5.0.0 | 1.2b0.dev | skeleton |
 <!-- end -->
--- a/instrumentation/opentelemetry-instrumentation-genai-dspy/.changelog/.gitignore
+++ b/instrumentation/opentelemetry-instrumentation-genai-dspy/.changelog/.gitignore
@@ -1,0 +1,1 @@
+!.gitignore
--- a/instrumentation/opentelemetry-instrumentation-genai-dspy/.changelog/439.added
+++ b/instrumentation/opentelemetry-instrumentation-genai-dspy/.changelog/439.added
@@ -1,0 +1,1 @@
+Add initial DSPy instrumentation package setup.
--- a/instrumentation/opentelemetry-instrumentation-genai-dspy/CHANGELOG.md
+++ b/instrumentation/opentelemetry-instrumentation-genai-dspy/CHANGELOG.md
@@ -1,0 +1,16 @@
+# Changelog
+
+All notable changes to this project will be documented in this file.
+
+The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
+and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
+
+<!--
+Do *NOT* add changelog entries here!
+
+This changelog is managed by towncrier and is compiled at release time.
+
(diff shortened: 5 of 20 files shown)


aabmass:
> There is 1 built in Module `ReAct` (https://dspy.ai/diving-deeper/react/) which triggers what they call an "agentic loop".. so probably that one could be an invoke agent span..

That sounds like a good fit. Earlier versions of LangGraph also had ReAct as the only canned agent loop you could use.

> Also `dspy` is yet another python library that wraps inference calls (`adk`, `agno`, and lots of other libraries do this too), so we have the same issue of double instrumentation when the model is also instrumented. The way I think we should resolve this is for the `dspy` instrumentation to check if the corresponding `model` instrumentation is installed and active. If it is, don't instrument the model call.

Were you able to check dspy with popular providers (say OpenAI, Anthropic, Gemini) uses the client libraries we already instrumented under the hood?

.github/workflows/changelog.yml
```diff
@@ -38,7 +38,7 @@ jobs:
 
       - name: Ensure no direct changes to CHANGELOG.md
         run: |
-          if [[ $(git diff --name-only FETCH_HEAD -- '**/CHANGELOG.md') ]]
+          if [[ $(git diff --diff-filter=MD --name-only FETCH_HEAD -- '**/CHANGELOG.md') ]]
```
aabmass, line 41: 👍

aabmass (approved):

DylanRussell:
Ah interesting it actually just uses `litellm` for this.. which apparently has native otel instrumentation.. So maybe this instrumentation should just emit `execute_tool` and `invoke_agent` spans and recommend the native `litellm` instrumentation for inference spans...

`agno` actually uses the SDKs, but I think optionally allows `litellm` or possibly it did but recently disabled that option due to the fact that it was recently compromised on pypy: https://github.com/agno-agi/agno/blob/7b8e5308e9c88e30153dfb0619c7b362bed272a2/libs/agno/pyproject.toml#L252

eternalcuriouslearner:
> > There is 1 built in Module `ReAct` (https://dspy.ai/diving-deeper/react/) which triggers what they call an "agentic loop".. so probably that one could be an invoke agent span..
> 
> 
> 
> That sounds like a good fit. Earlier versions of LangGraph also had ReAct as the only canned agent loop you could use.
> 
> 
> 
> > Also `dspy` is yet another python library that wraps inference calls (`adk`, `agno`, and lots of other libraries do this too), so we have the same issue of double instrumentation when the model is also instrumented. The way I think we should resolve this is for the `dspy` instrumentation to check if the corresponding `model` instrumentation is installed and active. If it is, don't instrument the model call.
> 
> 
> 
> Were you able to check dspy with popular providers (say OpenAI, Anthropic, Gemini) uses the client libraries we already instrumented under the hood?

Seconding this—I’m trying to instrument LlamaIndex’s ReActAgent as an invoke_agent operation. CrewAI, LlamaIndex, and DSPy appear to use similar agent architectures, such as ReAct.

DylanRussell:
I think if we are going to the trouble of instrumenting this package we should go ahead and instrument the model call, it's not much work and we can also support disabling it if the user wants to use `litellm`'s model span instead..

I tried out `litellm`'s gen ai otel instrumentation, it's OK.. They don't support the completion hook stuff, and there are a bunch of minor issues with what they produce.. Here's the list of attribute keys from their span

`[
  "gen_ai.cost.discount_amount",
  "gen_ai.cost.discount_percent",
  "gen_ai.cost.input_cost",
  "gen_ai.cost.margin_fixed_amount",
  "gen_ai.cost.margin_percent",
  "gen_ai.cost.margin_total_amount",
  "gen_ai.cost.original_cost",
  "gen_ai.cost.output_cost",
  "gen_ai.cost.reasoning_cost",
  "gen_ai.cost.tool_usage_cost",
  "gen_ai.cost.total_cost",
  "gen_ai.input.messages",
  "gen_ai.operation.name",
  "gen_ai.output.messages",
  "gen_ai.request.model",
  "gen_ai.response.finish_reasons",
  "gen_ai.response.id",
  "gen_ai.response.model",
  "gen_ai.system",
  "gen_ai.usage.input_tokens",
  "gen_ai.usage.output_tokens",
  "gen_ai.usage.total_tokens",
  "hidden_params",
  "litellm.call_id",
  "litellm.provider.model",
  "llm.is_streaming",
  "llm.request.type",
  "metadata.applied_guardrails",
  "metadata.cold_storage_object_key",
  "metadata.mcp_tool_call_metadata",
  "metadata.prompt_management_metadata",
  "metadata.requester_custom_headers",
  "metadata.requester_ip_address",
  "metadata.requester_metadata",
  "metadata.routing_decision",
  "metadata.spend_logs_metadata",
  "metadata.team_alias",
  "metadata.team_id",
  "metadata.usage_object",
  "metadata.user_agent",
  "metadata.user_api_key_alias",
  "metadata.user_api_key_auth_metadata",
  "metadata.user_api_key_budget_reset_at",
  "metadata.user_api_key_end_user_id",
  "metadata.user_api_key_hash",
  "metadata.user_api_key_max_budget",
  "metadata.user_api_key_org_alias",
  "metadata.user_api_key_org_id",
  "metadata.user_api_key_project_alias",
  "metadata.user_api_key_project_id",
  "metadata.user_api_key_request_route",
  "metadata.user_api_key_spend",
  "metadata.user_api_key_team_alias",
  "metadata.user_api_key_team_id",
  "metadata.user_api_key_team_max_budget",
  "metadata.user_api_key_team_spend",
  "metadata.user_api_key_user_email",
  "metadata.user_api_key_user_id",
  "metadata.user_api_key_user_max_budget",
  "metadata.user_api_key_user_spend",
  "metadata.vector_store_request_metadata"
]`

aabmass:
> I think if we are going to the trouble of instrumenting this package we should go ahead and instrument the model call, it's not much work and we can also support disabling it if the user wants to use `litellm`'s model span instead..

SGTM, maybe drop that in the issue. I'm merging this one

PR #443, opened 2026-08-21
[`opentelemetry-instrumentation-genai-portkey`] Add initial boilerplate / setup

DylanRussell:
# Description

This PR adds initial boilerplate / setup stuff. This is another open inference library port..

Portkey wraps around the Open AI SDK's `client.chat.completions.with_raw_response.create` which is  not instrumented by our instrumentation as well as exposing an API of it's own (a prompt library API called via httpx) to provide the user with a gateway that routes requests to any model provider API..

This instrumentation will just emit inference spans..

## Type of change

# How has this been tested?

Unit tests

# Checklist

- [x] Unit tests added
- [x] Documentation updated


lmolkova:
@DylanRussell do you want to try stacked PRs ? They don't work on forks, but you should be able to push to origin with your github alias as prefix and then they start working.

I'm a bit worried that we created a lot of empty skeletons, but we'renot populating them with instrumentations and I'd prefer us to merge real instrumentations. Stacked PRs allow to review individual stages in isolation, but merge them together

eternalcuriouslearner:
Question(non-blocking): Should we emit inference spans? Can we just emit agent, workflow and tool spans?

DylanRussell:
This library just wraps around inference calls (see my description of it in the first comment). I don't think any other spans make sense to write.

Sure, let me try out stacked PRs..

DylanRussell:
Hmm I see portkey wraps around the open AI SDK. But it calls an API which the open AI SDK instrumentor does not instrument `self.openai_client.with_raw_response.chat.completions.create(...)`

DylanRussell:
ok switching over to stacked PRs.. lets continue the discussion on https://github.com/open-telemetry/opentelemetry-python-genai/pull/464

PR #450, opened 2026-08-21
`opentelemetry-util-genai`: Align message-part class names with semconv (*Part suffix) (Util changes)

rads-1996:
# Description

Complies with the [semantic conventions for non-normative models.](https://github.com/open-telemetry/semantic-conventions-genai/blob/main/docs/gen-ai/non-normative/models.py) (util changes)

## Type of change

# How has this been tested?

configuration.

- [x] All test suites have been run

# Checklist

- [x] Unit tests added
- [ ] Documentation updated


util/opentelemetry-util-genai/src/opentelemetry/util/genai/types.py
```diff
@@ -202,6 +274,15 @@ class Uri:
...
 
 
+@dataclass()
+class Uri(UriPart):
+    """Represents an external referenced file sent to the model by URI
+
+    .. deprecated:: 1.2b0
+        Use ``UriPart`` instead.
+    """
+
```
aabmass: All these deprecated notices and extra classes looks like a lot of tech debt to take on for an experimental package.

Is there a strong reason to align with the non-normative conventions? We could even update the non-normative conventions to match the names in this util. WDYT?

rads-1996: @aabmass The initial discussion for this change started in another PR -

image

Following Liudmila's remark, I went and checked that for some data classes we were following the non-normative semantic conventions and for some we weren't. 

Even if we move with the suggestion of modifying the semantic conventions, we will still have to provide backward compatibility for the classes which already included the *Part suffix such as `GenericPart, CompactionPart`. Unless we don't mind some classes having the *Part suffix and others not.

lmolkova: > We could even update the non-normative conventions to match the names in this util. WDYT?

1. The JSON schemas in semconv are normative
2.  I think `Blob` and `Url` are not great names - they are too broad and confusing. Having Part suffix make them readable.
3. We have an inconsistent mix of parts, some of them have Part suffix and others don't

Having few deprecated things that will be removed in around the util stabilization time does not look like  a huge debt to me.

util/opentelemetry-util-genai/src/opentelemetry/util/genai/types.py
```diff
@@ -67,7 +67,16 @@ class ToolCallRequest:
 
 
 @dataclass()
-class ToolCallResponse:
+class ToolCallRequest(ToolCallRequestPart):
+    """Represents a tool call requested by the model (message part only).
+
+    .. deprecated:: 1.2b0
+        Use ``ToolCallRequestPart`` instead.
+    """
```
lmolkova: Subclassing isn't a drop-in rename - a dataclass subclass is not equal to its parent, and `isinstance` doesn't hold in the parent -> child direction. Against this branch:

```python
Text(content="a") == TextPart(content="a")            # False
isinstance(TextPart(content="a"), Text)                 # False
isinstance(ToolCallRequestPart(...), ToolCallRequest)   # False
```

So a `CompletionHook` written against the old names silently stops matching once instrumentations emit `*Part`:

```python
for p in message.parts:
    if isinstance(p, Text):        # False for TextPart -> content is no longer redacted
        p.content = "[redacted]"
```

In this repo, langchain `test_callback_handler.py` (`isinstance(part, ToolCallRequest)`) and smolagents `test_models.py` (`isinstance(parts[1], Blob)`) break as soon as the instrumentation PR lands.

A plain alias keeps identity, equality and `isinstance` intact. Sphinx autodoc picks up the string literal after a module-level assignment as that attribute's docstring, so the deprecation note still renders:

```suggestion
ToolCallRequest = ToolCallRequestPart
""".. deprecated:: 1.2b0
    Alias of :class:`ToolCallRequestPart`, kept for backwards compatibility.
"""
```

Same for the other eight deprecated names (`ToolCallResponse`, `ServerToolCall`, `ServerToolCallResponse`, `Text`, `Reasoning`, `Blob`, `File`, `Uri`).

util/opentelemetry-util-genai/tests/test_utils.py
```diff
@@ -136,6 +147,21 @@ def _normalize_to_dict(value: Any) -> dict[str, Any]:
...
 
 
+class TestDeprecatedMessagePartNames(unittest.TestCase):
+    def test_names_subclass_part_classes(self):
+        self.assertTrue(issubclass(Text, TextPart))
+        self.assertTrue(issubclass(Reasoning, ReasoningPart))
+        self.assertTrue(issubclass(Blob, BlobPart))
+        self.assertTrue(issubclass(File, FilePart))
+        self.assertTrue(issubclass(Uri, UriPart))
+
```
lmolkova: `issubclass` passes for both the subclass and the alias shape, so it can't pin the behavior that matters. Assert identity:

```suggestion
    def test_deprecated_names_are_aliases(self):
        self.assertIs(Text, TextPart)
        self.assertIs(Reasoning, ReasoningPart)
        self.assertIs(Blob, BlobPart)
        self.assertIs(File, FilePart)
        self.assertIs(Uri, UriPart)

    def test_parts_compare_equal_across_names(self):
        self.assertEqual(Text(content="hello"), TextPart(content="hello"))
        self.assertIsInstance(TextPart(content="hello"), Text)
```

util/opentelemetry-util-genai/tests/test_toolcall.py
```diff
@@ -155,6 +160,31 @@ def test_compactionpart_is_message_part():
     assert isinstance(msg.parts[0], CompactionPart)
 
 
+def test_deprecated_tool_call_names_subclass_part_classes():
+    """The pre-*Part tool call names subclass their replacements."""
+    assert issubclass(ToolCallRequest, ToolCallRequestPart)
+    assert issubclass(ToolCallResponse, ToolCallResponsePart)
+    assert issubclass(ServerToolCall, ServerToolCallPart)
+    assert issubclass(ServerToolCallResponse, ServerToolCallResponsePart)
```
lmolkova:
```suggestion
def test_deprecated_tool_call_names_are_aliases():
    """The pre-*Part tool call names are aliases of their replacements."""
    assert ToolCallRequest is ToolCallRequestPart
    assert ToolCallResponse is ToolCallResponsePart
    assert ServerToolCall is ServerToolCallPart
    assert ServerToolCallResponse is ServerToolCallResponsePart
```

util/opentelemetry-util-genai/src/opentelemetry/util/genai/types.py
rads-1996: Addressed

util/opentelemetry-util-genai/tests/test_utils.py
rads-1996: Addressed

util/opentelemetry-util-genai/tests/test_toolcall.py
rads-1996: Addressed

util/opentelemetry-util-genai/.changelog/450.deprecated
```diff
@@ -0,0 +1 @@
+Deprecate the message part classes `Text`, `Reasoning`, `Blob`, `File`, `Uri`, `ToolCallRequest`, `ToolCallResponse`, `ServerToolCall` and `ServerToolCallResponse`; they remain available as deprecated subclasses of their *Part replacements.
```
lmolkova: "deprecated subclasses" is left over from the earlier implementation - they are aliases now. The distinction matters to users: `Text is TextPart`, so `isinstance` and equality hold in both directions.

```suggestion
Deprecate the message part classes `Text`, `Reasoning`, `Blob`, `File`, `Uri`, `ToolCallRequest`, `ToolCallResponse`, `ServerToolCall` and `ServerToolCallResponse`; they remain available as aliases of their *Part replacements.
```

lmolkova (approved):

PR #466, opened 2026-08-24
[`opentelemetry-instrumentation-genai-portkey`] Add inference spans for sync / async and streaming / non-streaming

DylanRussell:
# Description

This PR adds inference spans for sync / async and streaming / non-streaming variants by monkey patching the 4 API methods that the portkey SDK exposes to users to make inference calls.

Portkey has 2 SDK methods that wrap around the Open AI SDK's `client.chat.completions.with_raw_response.create` which is not (yet?) instrumented by our Open AI SDK instrumentation, and has 2 SDK methods that call it's own API directly via httpx.

## Type of change

# How has this been tested?

Unit tests

# Checklist

- [x] Unit tests added
- [x] Documentation updated


instrumentation/opentelemetry-instrumentation-genai-portkey/src/opentelemetry/instrumentation/genai/portkey/utils.py
```diff
@@ -0,0 +1,352 @@
...
+
+def is_streaming(kwargs: dict[str, Any]) -> bool:
+    """Check whether the request is configured for streaming."""
+    stream = kwargs.get("stream")
+    if not value_is_set(stream):
+        return False
+    return bool(stream)
+
+
+def get_provider(client_instance: Any) -> str:
```
lmolkova: could we use exact type instead of Any?

instrumentation/opentelemetry-instrumentation-genai-portkey/src/opentelemetry/instrumentation/genai/portkey/utils.py
```diff
@@ -0,0 +1,352 @@
...
+    """Derive the provider name from the Portkey client instance."""
+    provider = getattr(client_instance, "provider", None)
+    if provider and isinstance(provider, str):
```
lmolkova: would it return the same set as we document in semconv?

instrumentation/opentelemetry-instrumentation-genai-portkey/src/opentelemetry/instrumentation/genai/portkey/utils.py
```diff
@@ -0,0 +1,352 @@
...
+        return provider.lower()
+    return "portkey"
+
+
+def get_server_address_and_port(
+    client_instance: Any,
```
lmolkova: is there a type we can use?

instrumentation/opentelemetry-instrumentation-genai-portkey/src/opentelemetry/instrumentation/genai/portkey/utils.py
lmolkova: could we use exact type instead of Any?

instrumentation/opentelemetry-instrumentation-genai-portkey/src/opentelemetry/instrumentation/genai/portkey/utils.py
lmolkova, line 95: would it return the same set as we document in semconv?

instrumentation/opentelemetry-instrumentation-genai-portkey/src/opentelemetry/instrumentation/genai/portkey/utils.py
lmolkova: is there a type we can use?

instrumentation/opentelemetry-instrumentation-genai-portkey/tests/conformance/tool_calling.py
```diff
@@ -0,0 +1,73 @@
...
+                                "type": "object",
+                                "properties": {
+                                    "city": {"type": "string"},
+                                },
+                                "required": ["city"],
+                            },
+                        },
+                    }
+                ]
+                client.chat.completions.create(
```
lmolkova: let's do the second call submitting tool responses so we can cover corresponding parts too

instrumentation/opentelemetry-instrumentation-genai-portkey/.changelog/466.added
```diff
@@ -0,0 +1 @@
+Add inference and streaming span instrumentation for Portkey
```
lmolkova, line 1: please add it to relesable packages in workflows as well

instrumentation/opentelemetry-instrumentation-genai-portkey/src/opentelemetry/instrumentation/genai/portkey/patch.py
```diff
@@ -79,58 +88,106 @@ def unpatch_portkey() -> None:
...
         args: tuple[Any, ...],
         kwargs: dict[str, Any],
     ) -> Any:
-        # TODO: Implement full inference invocation telemetry mapping
-        return wrapped(*args, **kwargs)
+        invocation = create_inference_invocation(
+            handler, instance, kwargs, capture_content
+        )
+        if is_streaming(kwargs):
+            return PortkeyStreamWrapper(
```
lmolkova: A streaming request that fails before the stream exists leaks the span. `handler.inference()` has already started it, and `wrapped(*args, **kwargs)` on line 103 runs outside `with invocation`, so nothing calls `stop()` or `fail()` - no span, no error type, no duration metric. Same in the other three wrappers (129, 157, 184).

Failing test:

```python
def test_sync_streaming_request_failure_records_span(
    tracer_provider, logger_provider, meter_provider, span_exporter
):
    with instrument(PortkeyInstrumentor(), tracer_provider=tracer_provider,
                    logger_provider=logger_provider, meter_provider=meter_provider,
                    content_capture="SPAN_ONLY"):
        p = Portkey(api_key="test_pk", provider="openai")
        p.chat.completions.openai_client = MagicMock()
        p.chat.completions.openai_client.chat.completions.create.side_effect = RuntimeError("boom")
        p.chat.completions._post = MagicMock(side_effect=RuntimeError("boom"))

        with pytest.raises(RuntimeError, match="boom"):
            p.chat.completions.create(
                messages=[{"role": "user", "content": "Hi"}],
                model="gpt-4o", stream=True,
            )

        assert len(span_exporter.get_finished_spans()) == 1  # got 0
```

The openai instrumentation's shape works: `try: result = wrapped(...) ... except Exception as error: invocation.fail(error); raise`. The four bodies here differ only in sync/async and `is_prompt`, so folding them into one factory would fix all four at once.

instrumentation/opentelemetry-instrumentation-genai-portkey/src/opentelemetry/instrumentation/genai/portkey/utils.py
```diff
@@ -0,0 +1,352 @@
...
+
+def _apply_request_parameters(
+    invocation: InferenceInvocation,
+    kwargs: dict[str, Any],
+    capture_content: bool,
+) -> None:
+    if (temp := get_value(kwargs.get("temperature"))) is not None:
+        invocation.temperature = float(temp)
+    if (
+        top_p := get_value(kwargs.get("top_p") or kwargs.get("p"))
```
lmolkova: nit: `kwargs.get("top_p") or kwargs.get("p")` drops `top_p=0.0`. Same on line 222 for `max_tokens`. Check `value_is_set` per key instead of `or`.

instrumentation/opentelemetry-instrumentation-genai-portkey/src/opentelemetry/instrumentation/genai/portkey/patch.py
DylanRussell: Ok I refactored it like you said. I feel like there are some failure cases we would probably not want to write a span for, like if the user passes a param the method doesn't accept, a TypeError would be raised.. But there are a lot of cases we do want to create a span for so I think this is better on the whole

instrumentation/opentelemetry-instrumentation-genai-portkey/.changelog/466.added
DylanRussell: Ack, done

instrumentation/opentelemetry-instrumentation-genai-portkey/src/opentelemetry/instrumentation/genai/portkey/utils.py
DylanRussell: Yes ! done

instrumentation/opentelemetry-instrumentation-genai-portkey/src/opentelemetry/instrumentation/genai/portkey/utils.py
DylanRussell: No.. Created a Map to do the conversion to the enum..

instrumentation/opentelemetry-instrumentation-genai-portkey/src/opentelemetry/instrumentation/genai/portkey/utils.py
DylanRussell: Yes

instrumentation/opentelemetry-instrumentation-genai-portkey/src/opentelemetry/instrumentation/genai/portkey/utils.py
DylanRussell: Done

instrumentation/opentelemetry-instrumentation-genai-portkey/tests/conformance/tool_calling.py
DylanRussell: Ack, done

DylanRussell:
I ran the conformance tests against the real APIs and got a passing run. Also noticed that portkey is using a vendored copy of open AI. So even if it were using methods we monkey patched (which it doesn't seem to be), it's using a vendored copy, so it makes sense to do the monkey patching against the portkey SDK..

lmolkova (approved):

PR #474, opened 2026-08-25
[`opentelemetry-instrumentation-genai-langchain`] Propagate gen_ai.conversation.id across the LangChain invocation tree

### LINKED ISSUE #475: langchain: gen_ai.conversation.id is set only on invoke_agent spans
### Describe your environment

OS: Ubuntu 22.04
Python version: Python 3.12
GenAI library (e.g. anthropic, openai) and version: langchain 1.3.16, langchain-openai 1.4.3

### What happened?

The LangChain instrumentation reads a conversation identifier from the
`thread_id`, `session_id`, or `conversation_id` metadata keys, but only applies
it to `invoke_agent` spans. The `chat` and `invoke_workflow` spans in the same
trace have no `gen_ai.conversation.id`, even though semconv defines the
attribute as conditionally required on all three.
`WorkflowInvocation` in `opentelemetry-util-genai` has no `conversation_id`
field at all, so an `invoke_workflow` span cannot carry one even when the
application supplies it.
Separately, `AgentInvocation` applies the attribute from
`_get_request_attributes`, which runs when the span finishes. The attribute is
absent for the whole life of the span, so a `SpanProcessor.on_start` cannot see
it and a child span cannot read it from its parent.

### Steps to Reproduce

`examples/agent/main.py` in the langchain package already reproduces this — it
sets `session_id` in the agent's metadata:
```python
agent = create_react_agent(llm, tools=[multiply, add]).with_config(
    {
        "metadata": {
            "agent_name": "coordinator",
            "session_id": session_id,
        },
    }
)
```
Run it against any exporter and inspect the resulting spans, or swap the OTLP
exporters for an `InMemorySpanExporter` and print `gen_ai.conversation.id` for
each span.

### Expected Result

Every span in the trace that semconv defines the attribute for carries it — the
`invoke_agent` span, the `chat` spans for the model calls it makes, and the
`invoke_workflow` span if the run is classified as a workflow.

### Actual Result

Only the `invoke_agent` span carries `gen_ai.conversation.id`. The `chat` spans
and any `invoke_workflow` span have no such attribute.

### Additional context

_No response_

Yes

### Tip

sfc-gh-zeningchen:
# Description

`gen_ai.conversation.id` only ever reached `invoke_agent` spans. Every `chat`
and `invoke_workflow` span was missing it, and `WorkflowInvocation` had no
`conversation_id` field at all, so a workflow root could not carry one even in
principle.

It was also applied in `AgentInvocation._get_request_attributes`, which runs
when the span finishes. The attribute was therefore absent for the entire life
of the span: no `SpanProcessor.on_start` could see it, and a child span could
not read it from its parent.

semconv defines `gen_ai.conversation.id` on three spans — inference,
invoke_agent, and invoke_workflow. All three now accept a `conversation_id`
and set it in the span's creation attributes. `WorkflowInvocation` gains the
field.

On the LangChain side, the callback handler resolves the id once per run from
the `thread_id`, `session_id`, or `conversation_id` metadata keys, and
`_InvocationManager` carries it down the invocation tree. Tool, retrieval, and
unclassified chain runs record it for their descendants without putting it on
their own spans, since semconv does not define the attribute for `execute_tool`
or `retrieval`. A model call nested under a tool still gets it.

Fixes #475 

## Type of change

# How has this been tested?

- [x] `uv run tox -e py314-test-util-genai`
- [x] `uv run tox -e py314-test-instrumentation-genai-langchain-latest`
- [x] `uv run tox -e typecheck`
- [x] `uv run pre-commit run ruff --all-files`
- [x] Live LangChain application, 12 scenarios covering agents, multi-tool
      loops, workflows, retrieval, streaming, and error paths. All 31 eligible
      spans (21 `chat`, 8 `invoke_agent`, 2 `invoke_workflow`) carried
      `gen_ai.conversation.id`; the 8 `execute_tool` and 1 `retrieval` spans
      did not, as intended.

New tests: conversation id on local agent, remote agent, and workflow spans in
`util-genai`, and its absence when none is supplied.
`langchain/tests/test_conversation_id.py` builds multi-span trees and asserts
inheritance, including through parents that do not carry the attribute
themselves. Resolution and precedence are covered in `test_callback_handler.py`,
`test_invocation_manager.py`, and `test_tools.py`.

# Checklist

- [x] Unit tests added
- [ ] Documentation updated


util/opentelemetry-util-genai/src/opentelemetry/util/genai/handler.py
```diff
@@ -314,6 +314,7 @@ def inference(
         server_address: str | None = None,
         server_port: int | None = None,
         operation_name: str | None = None,
+        conversation_id: str | None = None,
```
lmolkova: conversation id is not defined on inference spans - please suggest adding it in https://github.com/open-telemetry/semantic-conventions-genai/blob/main/docs/gen-ai/gen-ai-spans.md#inference before proceeding with this change. It should also appear on the corresponding event.

Once the semconv change lands, do not pass it as a start-time parameter to `handler.inference()`. Only sampling-relevant params go at start time; add it as a settable property on `InferenceInvocation`.

util/opentelemetry-util-genai/src/opentelemetry/util/genai/_inference_invocation.py
```diff
@@ -138,6 +140,7 @@ def _get_start_attributes(self) -> dict[str, AttributeValue]:
             (GenAI.GEN_AI_PROVIDER_NAME, self._provider),
             (server_attributes.SERVER_ADDRESS, self._server_address),
             (server_attributes.SERVER_PORT, self._server_port),
+            (GenAI.GEN_AI_CONVERSATION_ID, self.conversation_id),
```
lmolkova: Because `_get_metric_attributes()` calls `_get_start_attributes()`, this adds `gen_ai.conversation.id` to duration and token metrics. It cannot go on metrics as it has high cardinality.

sfc-gh-zeningchen: thanks for catching this! I've updated this part to avoid sending this with metrics.

util/opentelemetry-util-genai/src/opentelemetry/util/genai/handler.py
sfc-gh-zeningchen: Thanks! Just to confirm, it seems like `gen_ai.conversation.id` is already part of the inference span's [semconv](https://github.com/open-telemetry/semantic-conventions-genai/blob/main/docs/gen-ai/gen-ai-spans.md#inference) and the corresponding [event](https://github.com/open-telemetry/semantic-conventions-genai/blob/main/docs/gen-ai/gen-ai-events.md). Am I looking at the right place?
I've updated the PR so that `conversation_id` is no longer part of start attribute, also it was already in `_get_attributes` so that `_maybe_create_event` will pick it up as well. Thanks for catching this!

instrumentation/opentelemetry-instrumentation-genai-langchain/tests/requirements.oldest.txt
```diff
@@ -22,11 +22,11 @@
 
 
 
+# Drop this once opentelemetry-util-genai 1.2b0 is published.
```
rads-1996: Could you rebase your PR. A recent PR merge already added this editable dependency

sfc-gh-zeningchen: Sure! Just rebased

instrumentation/opentelemetry-instrumentation-genai-langchain/src/opentelemetry/instrumentation/genai/langchain/callback_handler.py
```diff
@@ -632,6 +645,19 @@ def on_retriever_error(
...
+    ) -> str | None:
+        if metadata:
+            for key in CONVERSATION_ID_METADATA_KEYS:
+                conversation_id = metadata.get(key)
+                if conversation_id:
+                    return str(conversation_id)
+
+        # The application attaches the id to a single point in the chain, so a
+        # run without one of its own tries to inherit from its nearest ancestor.
+        return self._invocation_manager.get_conversation_id(parent_run_id)
```
lmolkova: The parent lookup, and with it the whole `_InvocationState.conversation_id` plumbing, looks redundant: LangChain merges a run's config metadata into every descendant run, so a nested callback already receives `thread_id`/`session_id`/`conversation_id` in its own `metadata`.

Checked with langchain-core 1.6.1 / langgraph 1.x: an id set once via `with_config({"metadata": ...})` or via `configurable.thread_id` reaches the metadata of every chain, chat and tool callback below it, including a model call made inside a tool body that never forwards the config:

```
tool ask   {'thread_id': 't1'}   parent=False
chat       {'thread_id': 't1'}   parent=True
```

If there is a case where metadata does not propagate, a test that fails without the inheritance would pin it down. Otherwise resolving from `metadata` alone drops the extra `add_invocation_state` parameter, the `_InvocationState` field and `get_conversation_id`.

instrumentation/opentelemetry-instrumentation-genai-langchain/src/opentelemetry/instrumentation/genai/langchain/invocation_manager.py
```diff
@@ -39,12 +41,24 @@ def add_invocation_state(
             parent_invocation_state = self._invocations[parent_run_id]
             parent_invocation_state.children.append(run_id)
 
+        invocation_state.conversation_id = (
```
lmolkova: nit: dead branch. Every caller passes a value that `_resolve_conversation_id` already resolved from the parent, so the `else` only runs from tests. Two places implementing inheritance can drift - keep one.

lmolkova (approved):

instrumentation/opentelemetry-instrumentation-genai-langchain/src/opentelemetry/instrumentation/genai/langchain/callback_handler.py
sfc-gh-zeningchen: Thanks for calling it out, I checked and it turns out what you said is correct! 
I've removed the parent lookup, also removed the field from `_InvocationState`, now `conversation.id` comes from metadata only. Sorry for the confusion.

instrumentation/opentelemetry-instrumentation-genai-langchain/src/opentelemetry/instrumentation/genai/langchain/invocation_manager.py
sfc-gh-zeningchen: Thanks! This one went away with the same change that removes `conversation_id` from `_InvocationState`

util/opentelemetry-util-genai/src/opentelemetry/util/genai/handler.py
lmolkova: sorry, you're absolutely right!

PR #482, opened 2026-08-25
[`opentelemetry-instrumentation-genai-langchain`] Record streaming telemetry for the LangChain instrumentation

### LINKED ISSUE #481: langchain: streamed calls are not marked as streaming, thus no streaming timing signals emitted
### Describe your environment

OS: Ubuntu 22.04
Python version: Python 3.12
GenAI library (e.g. anthropic, openai) and version: langchain 1.3.16, langchain-openai 1.4.3

### What happened?

`opentelemetry-util-genai` 1.1b0 added streaming telemetry in #269 — the
`gen_ai.request.stream` and `gen_ai.response.time_to_first_chunk` span
attributes, and the `gen_ai.client.operation.time_to_first_chunk` and
`gen_ai.client.operation.time_per_output_chunk` histograms. None of it reaches
the LangChain path.
Those values are set by the shared stream wrappers, which need an SDK stream to
wrap. The LangChain instrumentation is callback-based: it opens an
`InferenceInvocation` in `on_chat_model_start`, finishes it in `on_llm_end`, and
never sees a stream. `on_llm_new_token` is not implemented, so there is no
per-chunk hook either.
A streamed call is therefore indistinguishable from a synchronous one, and
time-to-first-token cannot be separated from total duration.

### Steps to Reproduce

Stream a chat call with the instrumentation enabled:
```python
LangChainInstrumentor().instrument()
list(ChatOpenAI(model="gpt-4o-mini", stream_usage=True).stream("Say hello."))
```
Then inspect the finished span's attributes and the emitted metric names.

### Expected Result

The `chat` span carries `gen_ai.request.stream` = `true` and a plausible
`gen_ai.response.time_to_first_chunk`, and both streaming histograms record
points. Non-streamed calls are unchanged.

### Actual Result

Both span attributes are absent and neither histogram is emitted.

### Additional context

_No response_

Yes

### Tip

sfc-gh-zeningchen:
# Description

Streamed LangChain calls never set `gen_ai.request.stream` or
`gen_ai.response.time_to_first_chunk`, and never recorded the
`time_to_first_chunk` / `time_per_output_chunk` histograms added in #269,
because the shared stream wrappers are the only producers of those values and a
callback-based instrumentation has no stream to wrap. This adds
`GenAIInvocation.record_stream_chunk()` to `opentelemetry-util-genai` as the
entry point for reporting chunk arrival without a stream, and implements
`on_llm_new_token` to call it once per output chunk. The capability lives in the
util rather than the instrumentation, per the direction on #8.

Fixes #481 

## Type of change

# How has this been tested?

New `tests/test_streaming.py` in the langchain package covers `.stream()`,
`.astream()`, a non-streamed `.invoke()` asserting no stream attributes or
metric points appear, and a parametrized case separating LangChain's
end-of-stream marker from a genuinely content-free chunk. These patch the
model's `_stream` / `_astream` rather than replaying a cassette, because the
assertions are about chunk timing, which a cassette does not reproduce. Two
tests in `test_handler_metrics.py` cover `record_stream_chunk()` directly.

- [x] `util-genai` suite: 332 passed
- [x] `langchain` suite: 239 passed
- [x] Both matrix legs: verified at the declared floor (`langchain==0.3.21`,
      `langchain-core` 0.3.86, where `chunk_position` does not exist) and at
      latest (`langchain==1.3.16`, `langchain-core` 1.6.0)
- [x] `tox -e precommit` (ruff, ruff-format) and `pyright` clean

# Checklist

- [x] Unit tests added
- [ ] Documentation updated


instrumentation/opentelemetry-instrumentation-genai-langchain/tests/test_streaming.py
```diff
@@ -0,0 +1,209 @@
+# Copyright The OpenTelemetry Authors
```
lmolkova, line 1: Please add a streaming conformance scenario under `tests/conformance/inference_streaming.py` (mirroring `anthropic`/`openai` packages) to validate `gen_ai.request.stream=true` and streaming metrics against the semantic conventions.

instrumentation/opentelemetry-instrumentation-genai-langchain/.changelog/482.fixed
```diff
@@ -0,0 +1 @@
+Mark streamed calls with `gen_ai.request.stream` and record `gen_ai.response.time_to_first_chunk` plus the streaming timing metrics, which were never emitted for LangChain.
```
lmolkova, line 1: Please rename changelog fragment files to use the PR number (482) instead of 461 (`482.fixed`, `482.changed`, `util/.../482.added`).

lmolkova (approved):
Thank you! Please add conformance tests with expected streaming metrics and it should be good to go.

lmolkova (approved):

PR #485, opened 2026-08-26
openai-agents: mark tool and agent span errors

ordemri:
The agents library records failures on Span.error, but on_span_end always called invocation.stop(). A failed tool call was therefore indistinguishable from a successful one by status and error.type: the span ended UNSET with no error.type, so a backend could not filter or count failures.

SpanError is a mapping with no exception type, while util-genai derives error.type from an exception class. AgentsException is the library's own base type and is present across the package's supported agents-library range, so error.type stays bounded and the change works on both ends of the declared opentelemetry-util-genai range.

Only failures the SDK records on Span.error are marked. A tool that reports failure by returning data, without raising, leaves Span.error unset and remains unmarked.

Assisted-by: Claude Opus 5

# Description

Please include a summary of the change and which issue, if any, is fixed.
Include the motivation and context for the change.

Fixes # (issue)

## Type of change

# How has this been tested?

configuration.

- [ ] Test A

# Checklist

- [ ] Unit tests added
- [ ] Documentation updated


instrumentation/opentelemetry-instrumentation-genai-openai-agents/src/opentelemetry/instrumentation/genai/openai_agents/processor.py
```diff
@@ -130,6 +131,12 @@ def on_span_end(self, span: Span[Any]) -> None:
                 invocation.tool_result = (
                     output if isinstance(output, str) else str(output)
                 )
+        # SpanError is a mapping, not a raised exception, so it never
+        # reaches util-genai's exception path on its own.
+        span_error = getattr(span, "error", None)
+        if span_error is not None:
```
lmolkova, line 140: Construct `Error` from `opentelemetry.util.genai.types` using `ErrorTypeValues.OTHER.value` and `str(span_error)`:

```suggestion
        span_error = getattr(span, "error", None)
        if span_error is not None:
            invocation.fail(
                Error(
                    type=ErrorTypeValues.OTHER.value,
                    message=str(span_error),
                )
            )
            return
```

See commit [fed2c6a](https://github.com/lmolkova/opentelemetry-python-genai/commit/fed2c6a6f1d2c695ca4086659779df3f52aa3178) fas a sample.

Rationale: we don't need to invent exception types when there was no exception, we report what openai agents provide.

Ideally we'd want something useful and descriptive as error.type (actual exception type name, etc), but all OpenAI gives is SpanError dict with message and arbitrary context. `SpanError` us not quite useful, so let's use fallback OTHER type. If there is some heuristic allowing to have more granular errors in the error type - let's use them.

instrumentation/opentelemetry-instrumentation-genai-openai-agents/tests/test_processor.py
```diff
@@ -219,6 +220,60 @@ def test_no_content_captured_when_capture_env_unset(
     assert tool_span.attributes["gen_ai.tool.name"] == "get_weather"
 
 
+def test_tool_span_error_sets_error_status_and_type(
```
lmolkova, line 223: Please add an integration test exercising `Runner.run` with a failing tool and a VCR cassette:

```python
@agents.function_tool
def _failing_tool() -> str:
    raise ValueError("database connection failed")


@pytest.mark.vcr
@pytest.mark.asyncio
async def test_runner_failing_tool_records_span_error(
    tracer_provider: TracerProvider,
    logger_provider: Any,
    meter_provider: Any,
    span_exporter: InMemorySpanExporter,
) -> None:
    with instrument(
        OpenAIAgentsInstrumentor(),
        tracer_provider=tracer_provider,
    ):
        agent = agents.Agent(
            name="test_agent",
            model="gpt-4o-mini",
            tools=[_failing_tool],
        )
        await agents.Runner.run(agent, "run tool")

    spans = {s.name: s for s in span_exporter.get_finished_spans()}
    assert "execute_tool _failing_tool" in spans
    tool_span = spans["execute_tool _failing_tool"]
    assert tool_span.status.status_code is StatusCode.ERROR
    assert "database connection failed" in (tool_span.status.description or "")
    assert tool_span.attributes is not None
    assert tool_span.attributes["error.type"] == "_OTHER"
```

See commit [fed2c6a](https://github.com/lmolkova/opentelemetry-python-genai/commit/fed2c6a6f1d2c695ca4086659779df3f52aa3178) for the full change including the VCR cassette.

lmolkova (commented):
thanks for the contribution!

instrumentation/opentelemetry-instrumentation-genai-openai-agents/src/opentelemetry/instrumentation/genai/openai_agents/processor.py
ordemri: Applied — I took your patch from fed2c6a directly, so this is your code (37a8abb).

Agreed on not inventing a type. One note in case it's useful: `str(span_error)` turns out to be load-bearing. With `span_error["message"]` alone the status description came out as `Error running tool (non-fatal)` — the actual failure text lives in `data`, so serializing the whole mapping is what preserves it.

On a more granular `error.type`: the only thing to key off is the SDK's ~29 hardcoded `SpanError(...)` message literals, and those carry no stability guarantee — a mapping would quietly degrade to `_OTHER` whenever one is reworded, which is the failure mode hardest to notice. `_OTHER` looks like the right call here.

instrumentation/opentelemetry-instrumentation-genai-openai-agents/tests/test_processor.py
ordemri: Added in 37a8abb — I took fed2c6a wholesale: your cassette, your test in `test_instrumentor.py`, and the instrumentor change that builds its own `TelemetryHandler` instead of taking the `get_telemetry_handler` singleton. That last one turned out to be the root cause of a test-isolation problem I had worked around in conftest, so thanks — the workaround is gone.

Two small things:

- The reordering in that patch moved `pytest_plugins` above the vcr import, which left `pytest.register_assert_rewrite` running after the module was already imported and raised `PytestAssertRewriteWarning` on every run. I restored the original order.
- The link in your comment points at `fed2c6a6f1d2c695...`, which doesn't resolve; the short `fed2c6a` does. Looks like an amend after you posted it, noting in case anyone else goes looking.

CI on this branch is still pending workflow approval, if you're able to kick it off.

lmolkova (approved):
Please fix failing tests!

ordemri:
Fixed in 30888cd. Two independent things, both only visible in the `openai_agents-oldest` env:

1. Reporting `error.type` as a string needs util-genai 1.1b0 — that is where `Error.type` became the attribute value instead of the exception class (#282). The package still declared `>= 1.0b0`, and on 1.0b0 `_apply_error_attributes` does `error.type.__qualname__`, so all three error tests raised `AttributeError: 'str' object has no attribute '__qualname__'`. Bumped the floor to 1.1b0.

2. The cassette's `usage` blocks had no `input_tokens_details` / `output_tokens_details`. openai 1.109.1, which is what the oldest env resolves to, models those as non-null, so response parsing failed validation before the tool ever ran. Newer openai tolerates their absence, which is why only the oldest env went red. Added both.

I reproduced the exact CI result locally (3 failed, 17 passed) against openai-agents 0.3.3 / openai 1.109.1 / util-genai 1.0b0, then confirmed 20 passed on both the oldest resolution and latest.

PR #488, opened 2026-08-27
[`opentelemetry-instrumentation-genai-bedrock`] Add inference spans for `Converse` and `ConverseStream` methods

DylanRussell:
# Description

Adds logic to create an `InferenceInvocation` and parse the request/response for attributes and set them on the `invocation` for the `Converse` and `ConverseStream` methods in `boto3`.

Remaining gaps:
- Similar logic needs to be added for the InvokeModel & InvokeModelWithResponseStream methods
- We need to monkey patch `aioboto3` to get the async equivalents.

## Type of change

# How has this been tested?

Unit tests..

# Checklist

- [x] Unit tests added
- [x] Documentation updated


instrumentation/opentelemetry-instrumentation-genai-bedrock/tests/conformance/tool_calling.py
```diff
@@ -0,0 +1,98 @@
...
+
+
+class ToolCallingScenario(Scenario):
+    expected_spans = {"chat": 1}
+    expected_metrics = (
+        "gen_ai.client.operation.duration",
+        "gen_ai.client.token.usage",
+    )
+    expected_violations = (
+        ExpectedViolation(
```
lmolkova: can we populate it?

instrumentation/opentelemetry-instrumentation-genai-bedrock/tests/conformance/tool_calling.py
```diff
@@ -0,0 +1,98 @@
...
+            advice_id="genai_expected_attribute_missing",
+            message_substring="gen_ai.response.id",
```
lmolkova: can we populate it?

lmolkova: can we populate it?

instrumentation/opentelemetry-instrumentation-genai-bedrock/src/opentelemetry/instrumentation/genai/bedrock/extractors.py
```diff
@@ -0,0 +1,236 @@
...
+
+
+def extract_converse_response(
+    response: dict[str, Any],
+    invocation: InferenceInvocation,
+    model_id: str | None = None,
+    *,
+    capture_content: bool = True,
+) -> None:
+    invocation.response_model_name = model_id
```
lmolkova: `gen_ai.response.id` is available as `ResponseMetadata.RequestId`; the conformance scenarios declare it missing instead.

And `response_model_name` is just the request `modelId` - with an inference profile that puts an ARN in `gen_ai.response.model`.

instrumentation/opentelemetry-instrumentation-genai-bedrock/src/opentelemetry/instrumentation/genai/bedrock/extractors.py
```diff
@@ -0,0 +1,236 @@
...
+        )
+
+    tool_result = block.get("toolResult")
+    if _is_dict(tool_result):
+        return ToolCallResponse(
+            id=tool_result.get("toolUseId"),
+            response=tool_result.get("content"),
+        )
+
+    return None
```
lmolkova, line 141: `reasoningContent` and the other non-text blocks are dropped, should we follow up on the other parts in a separate PR?

instrumentation/opentelemetry-instrumentation-genai-bedrock/src/opentelemetry/instrumentation/genai/bedrock/extractors.py
lmolkova: `gen_ai.response.id` is available as `ResponseMetadata.RequestId`; the conformance scenarios declare it missing instead.

`response_model_name` should be dropped - Bedrock doesn't return a response model. If weaver complains in conformance, please add that one as  expected violation - it's useful input for semconv to document the case where a provider has no response model.

lmolkova (approved):

instrumentation/opentelemetry-instrumentation-genai-bedrock/src/opentelemetry/instrumentation/genai/bedrock/extractors.py
DylanRussell: removed `response_model_name`.

I don't think `requestId` should be used for that. When you send an HTTP request to any AWS service, AWS's API gateway assigns a unique request tracking ID and returns it in the HTTP response header x-amzn-RequestId -- where as i think response id is like a model completion / continuation id (something generated by the model)

instrumentation/opentelemetry-instrumentation-genai-bedrock/src/opentelemetry/instrumentation/genai/bedrock/extractors.py
DylanRussell: I added it in here.. WDYT of the various supported block types here (https://docs.aws.amazon.com/bedrock/latest/APIReference/API_runtime_ReasoningContentBlock.html) -- i left out audio / video blocks, but added parsing for document / image types.. not sure what the best approach is

DylanRussell:
I recorded all the VCR tests against the real API and got passing results.. updated all the .yaml files with the real req/resp exchange..

instrumentation/opentelemetry-instrumentation-genai-bedrock/src/opentelemetry/instrumentation/genai/bedrock/extractors.py
lmolkova: we can add them as `GenericPart` and create semconv issue to add others. 

BTW, would you mind updating Blob (deprecated now) -> BlobPart and such?

DylanRussell: Opened https://github.com/open-telemetry/semantic-conventions-genai/issues/488 and used GenericPart for them. Switched to `XPart` types too..

DylanRussell: On second thought I wonder if it's worth adding these as GenericParts with just the `type` set, is that useful at all ? I don't think we will ever add raw audio / video to the telemetry, it's way too big.

Even a single image can be too big.. Google's telemetry API has a cap at 64KB for an attribute value for example.. Maybe we should be checking the size of `BlobPart`s (like mages) and leaving them out if they exceed 50KB or something like that..

PR #494, opened 2026-08-30
feat: Add LlamaIndex agent and tool instrumentation

eternalcuriouslearner:
Adds `invoke_agent` spans for LlamaIndex `FunctionAgent` and `ReActAgent` runs and `execute_tool` spans for tool calls, while deduplicating nested `FunctionTool.call`/`acall` callbacks. Provider inference remains delegated to the provider SDK instrumentation.

Includes VCR-backed tests, provider composition, conformance scenarios, examples, and documentation.

Known gap: `AgentWorkflow` is intentionally handled by the next PR in this stack.


instrumentation/opentelemetry-instrumentation-genai-llama-index/src/opentelemetry/instrumentation/genai/llama_index/__init__.py
```diff
@@ -32,8 +32,14 @@
 from __future__ import annotations
 
 from collections.abc import Collection
-from typing import Any
+from typing import Any, Protocol, cast
 
+from llama_index.core.instrumentation.span_handlers import BaseSpanHandler
+from llama_index_instrumentation import get_dispatcher
```
lmolkova: Please import `get_dispatcher` from `llama_index.core.instrumentation` instead of `llama_index_instrumentation` directly (here and in the test suite). `llama_index_instrumentation` is an internal transitive dependency rather than the declared `llama-index-core` package, so importing it directly bypasses the public API and risks breaking if upstream restructures or vendors internal packages.

lmolkova (approved):

PR #505, opened 2026-08-31
[`opentelemetry-instrumentation-genai-langchain`] Record gen_ai.response.model on streamed spans

### LINKED ISSUE #503: langchain: streamed spans are missing gen_ai.response.model
### Describe your environment

OS: Ubuntu 22.04
Python version: Python 3.12
GenAI library (e.g. anthropic, openai) and version: langchain-openai 1.6.0

### What happened?

`gen_ai.response.model` is absent from the `chat` span of a streamed call. 

### Steps to Reproduce

Stream a chat call with the instrumentation enabled:
```python
LangChainInstrumentor().instrument()
list(ChatOpenAI(model="gpt-4o-mini", stream_usage=True).stream("Write a one-sentence greeting."))
```
Then inspect the finished chat span's attributes.

### Expected Result

The streamed span carries gen_ai.response.model, as the non-streamed span does.

### Actual Result

Neither gen_ai.response.model nor gen_ai.response.id is present.

### Additional context

_No response_

Yes

### Tip

sfc-gh-zeningchen:
# Description

`gen_ai.response.model` and `gen_ai.response.id` were absent from every
streamed `chat` span and present on every non-streamed one.

`on_llm_end` read both from `LLMResult.llm_output`. LangChain never populates
that for a streamed call: it assembles the final `LLMResult` from the merged
chunks alone. The values are still on the message's `response_metadata`, which
LangChain fills with the union of the generation's `generation_info` and
whatever the provider wrote onto the message, so reading that covers
langchain-openai (which reports the model through `generation_info`) and
langchain-anthropic (which writes it onto the message at `message_start`).

The precedence chain, generation then `llm_output` then the Responses API
served-model header, moves into one helper so it lives in one place instead of
being spread across the callback.

`gen_ai.response.id` stays absent for streamed Chat Completions and Anthropic
calls. Both providers send a response id on every chunk, but their LangChain
adapters do not surface it on the streaming path. Only the OpenAI Responses API
records one, and that path is covered here. `message.id` is not a substitute
because LangChain fills it with its own run id.

Fixes #503

## Type of change

# How has this been tested?

- [x] `TestOnLlmEndStreamedResponse` in `tests/test_callback_handler.py` covers
  the model and id read from a streamed generation, `llm_output` and the
  served-model header keeping precedence, both attributes staying unset when no
  source reports them, and `message.id` never being used as the response id.
- [x] `test_chat_openai_streamed_response_model` in `tests/test_llm_call.py`
  drives a real `.stream()`. The model is reported only in `generation_info`
  there, and LangChain is what copies it into `response_metadata`, so a
  hand-built `LLMResult` would skip that step and prove nothing.

# Checklist

- [x] Unit tests added


instrumentation/opentelemetry-instrumentation-genai-langchain/.changelog/505.fixed
```diff
@@ -0,0 +1 @@
+Record gen_ai.response.model and gen_ai.response.id on streamed spans
```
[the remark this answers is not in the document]
sfc-gh-zeningchen: fixed

instrumentation/opentelemetry-instrumentation-genai-langchain/src/opentelemetry/instrumentation/genai/langchain/utils.py
```diff
@@ -383,6 +383,64 @@ def serialize(obj: Any) -> str | None:
...
+    response_metadata: Mapping[str, Any] = (
+        getattr(message, "response_metadata", None) or {}
+    )
+
+    response_model = _first_string(
+        response_metadata.get("model_name"),
+        response_metadata.get("model"),
+    )
+    response_id = _first_string(response_metadata.get("id"))
+    return response_model, response_id
```
lmolkova, line 554: should we check in generation_info too and test it?

```python
def test_model_from_generation_info(self):
    run_id = _run_id()
    handler, _, llm_inv = _make_handler_with_llm_invocation(run_id)
    gen = ChatGenerationChunk(
        message=AIMessageChunk(content="hello"),
        generation_info={"model_name": "claude-sonnet-4-5"},
    )
    handler.on_llm_end(
        response=LLMResult(generations=[[gen]]), run_id=run_id
    )
    # Fails: llm_inv.response_model_name is None because response_fields_from_generation
    # only inspects message.response_metadata and ignores generation_info.
    assert llm_inv.response_model_name == "claude-sonnet-4-5"
```

instrumentation/opentelemetry-instrumentation-genai-langchain/tests/test_llm_call.py
```diff
@@ -991,3 +994,50 @@ def test_chat_anthropic_claude_sonnet_cache_token_details(
     assert (
         span.attributes.get(gen_ai_attributes.GEN_AI_USAGE_INPUT_TOKENS) == 22
     )
+
+
+def test_chat_openai_streamed_response_model(
```
lmolkova, line 1309: Please add test coverage for async streaming (`.astream()`).

lmolkova (approved):

instrumentation/opentelemetry-instrumentation-genai-langchain/src/opentelemetry/instrumentation/genai/langchain/utils.py
sfc-gh-zeningchen: Thanks! It was initially dropped because LangChain merges `generation_info` into `response_metadata` before `on_llm_end` sees it. But thinking it reading `generation_info` too can help avoid depending on that internal. Added it as a fallback with this test.

instrumentation/opentelemetry-instrumentation-genai-langchain/tests/test_llm_call.py
sfc-gh-zeningchen: thanks, added `test_chat_openai_astreamed_response_model`

PR #507, opened 2026-08-31
Instrument LangGraph execution state changes

Utkarsh-Sinha0:
Adds a LangGraph producer for execution state-change telemetry using the schema proposed in open-telemetry/semantic-conventions-genai#483.

Behavior:
- non-empty LangGraph node deltas emit `gen_ai.execution.state.changed` in the active workflow span context
- the event includes only `gen_ai.execution.state.changed_key.count`
- state values, changed-key names, execution identifiers, and scheduler step metadata are never emitted
- callback state retains only the LangGraph-node marker, not application metadata

Validation:
- LangChain instrumentation suite: 323 passed
- shared utility suite: 363 passed
- README rendering: passed
- Ruff checks: passed

Merge dependency:
- open-telemetry/semantic-conventions-genai#483 must land first
- this implementation must then consume the generated semantic-convention constants instead of provisional string literals


meshailabs:
Building the suspend/resume half alongside this: `gen_ai.agent.paused` / `checkpointed` / `resumed` from open-telemetry/semantic-conventions-genai#445, on the same callback plumbing. `paused` reads `__interrupt__` from `on_chain_end`; `checkpointed` and `resumed_from` come from wrapping the checkpointer. Not touching `state.changed`. If this lands first I will rebase onto it so the LangGraph metadata handling stays in one place.

PR #512, opened 2026-08-31
[`opentelemetry-instrumentation-genai-langchain`] Set `gen_ai.agent.name` on child `execute_tool` spans

sfc-gh-zeningchen:
# Description

`gen_ai.agent.name` is listed as **Conditionally Required — "When applicable"** on the `execute_tool` span in [semconv](https://opentelemetry.io/docs/specs/semconv/gen-ai/gen-ai-spans/#execute-tool-span), but the LangChain callback never set it, so tool spans running under an `invoke_agent` could not be joined back to their agent. Add an `agent_name` kwarg on `TelemetryHandler.tool()`; the LangChain callback resolves it via the existing `_find_nearest_agent`. Metric attribute sets are unchanged (the shared `gen_ai.client.operation.duration` metric does not list `gen_ai.agent.name` today, and util-genai does not emit `gen_ai.execute_tool.duration`).

## Type of change

# How has this been tested?

New unit tests in `tests/test_tools.py` (langchain) cover single agent, nested agents (nearest wins), and no-agent baseline. Tests in `test_toolcall.py` and `test_handler_metrics.py` (util-genai) assert the attribute lands on the span but not on `gen_ai.client.operation.duration`.

- [x] `util-genai` suite: 334 passed
- [x] `langchain` suite: 237 passed
- [x] `tox -e precommit` (ruff, ruff-format) and `pyright` clean

# Checklist

- [x] Unit tests added


util/opentelemetry-util-genai/src/opentelemetry/util/genai/_tool_invocation.py
```diff
@@ -61,6 +63,7 @@ def __init__(
         tool_call_id: str | None = None,
         tool_type: str | None = None,
         tool_description: str | None = None,
+        agent_name: str | None = None,
```
lmolkova, line 66: this is not one of the sampling-relevant attributes according to semconv https://github.com/open-telemetry/semantic-conventions-genai/blob/67dff024110be5bd9f318006e733f4078e0f4c97/docs/gen-ai/gen-ai-spans.md?plain=1#L1190-L1195, so it should not be set at the start time.

I can see why it should be one of them, mind sending a PR to semantic-conventions-genai to fix it?

util/opentelemetry-util-genai/tests/test_handler_metrics.py
```diff
@@ -618,6 +550,27 @@ def test_fail_tool_records_duration_with_error(self) -> None:
         self.assertAlmostEqual(duration_point.sum, 1.5, places=3)
         self.assertNotIn("gen_ai.client.token.usage", metrics)
 
+    def test_stop_tool_does_not_leak_agent_name_to_duration_metric(
```
lmolkova: why so?

it's an attribute on duration metric https://github.com/open-telemetry/semantic-conventions-genai/blob/main/docs/gen-ai/gen-ai-metrics.md#metric-gen_aiexecute_toolduration


Or maybe we didn't update this repo yet to report relatively new gen_ai.execute_tool.duration metric

lmolkova (approved):

rads-1996 (commented):
I am thinking if it would be a good idea to also add the agent.id attribute. I can take it up in  a separate PR, if it can't be part of this one.

lmolkova:
> I am thinking if it would be a good idea to also add the agent.id attribute. I can take it up in a separate PR, if it can't be part of this one.

agent.id is not part of internal agent span, it's intended as identifier of a hosted remote agent on client spans.

util/opentelemetry-util-genai/src/opentelemetry/util/genai/_tool_invocation.py
sfc-gh-zeningchen: sure, I just updated the code to remove it from sampling-relevant attributes. I'll send a PR to semconv and follow-up on this later

util/opentelemetry-util-genai/tests/test_handler_metrics.py
sfc-gh-zeningchen: sorry, the name is a little bit confusing, this was about `gen_ai.client.operation.duration` instead of `gen_ai.execute_tool.duration`. Updated the test name to make it clear.
happy to follow up with the emission of `gen_ai.execute_tool.duration` later!

PR #520, opened 2026-09-01
util-genai: record cancellation as a failure, not a success

ordemri:
## Description

`GenAIInvocation.__exit__` called `fail()` only when the in-flight exception derived from
`Exception`, so `asyncio.CancelledError` took the `stop()` branch: a cancelled operation was exported
with `StatusCode.UNSET`, no `error.type`, and a duration metric with no error dimension —
indistinguishable from one that completed. This guards on `BaseException` instead, matching the guard
the stream wrappers in `stream.py` already use.

`fail()` already accepts a `BaseException`, so nothing else changed. `__exit__` still returns `None`
— suppressing `CancelledError` would break asyncio cancellation for every caller — and two tests
cover that, one of them cancelling a real task mid-`await`.

### Known gaps

- The wider guard also catches `KeyboardInterrupt`, `SystemExit` and `GeneratorExit`. `GeneratorExit`
  is not reachable from any shipped instrumentation today — no `with <invocation>:` block in this
  repository contains a `yield` — but it is covered by a test. Happy to narrow this; it is one line.
- This fixes the context-manager path only. Packages that finalize in their own `except Exception:`
  handler are **worse off than unfixed, not equal**: on cancellation neither branch runs and the span
  is never ended, so nothing is exported at all. Happy to follow up separately.

## Type of change

## How has this been tested?

New tests in `util/opentelemetry-util-genai/tests/test_invocation_base_exception.py`, written before
the fix and confirmed red against unmodified `main`: **6 failed, 342 passed** — five with
`AssertionError: <StatusCode.UNSET: 0> != <StatusCode.ERROR: 2>`, one with `KeyError: 'error.type'`.
The file's other tests passed before and after, so it discriminates rather than failing wholesale.

- [x] `tox -e py310-test-util-genai` — 348 passed
- [x] `tox -e py314-test-util-genai` — 348 passed
- [x] `tox -e py310-test-instrumentation-genai-agno-oldest` — 20 passed
- [x] `tox -e py310-test-instrumentation-genai-agno-latest` — 20 passed
- [x] `tox -e py314-test-instrumentation-genai-agno-latest` — 20 passed
- [x] `google-genai` and `smolagents` oldest / latest — passed
- [ ] Conformance environments **not verified**: they skip locally because the `weaver` binary is not
      installed, so no assertion ran. Relying on CI for those.

## Checklist

- [x] Changelog updated
- [x] Unit tests added
- [ ] Documentation updated — not applicable

### DIFF
--- a/util/opentelemetry-util-genai/.changelog/520.fixed
+++ b/util/opentelemetry-util-genai/.changelog/520.fixed
@@ -1,0 +1,1 @@
+Record cancellation as a failure: the invocation context manager finalized any non-`Exception` `BaseException` (`asyncio.CancelledError`, `KeyboardInterrupt`, `SystemExit`, `GeneratorExit`) through the success path, so cancelled operations were exported with an unset span status and no `error.type`.
--- a/util/opentelemetry-util-genai/src/opentelemetry/util/genai/_invocation.py
+++ b/util/opentelemetry-util-genai/src/opentelemetry/util/genai/_invocation.py
@@ -227,7 +227,7 @@ class GenAIInvocation(AbstractContextManager["GenAIInvocation"]):
         exc_value: BaseException | None,
         traceback: TracebackType | None,
     ) -> None:
-        if exc_value is not None and isinstance(exc_value, Exception):
+        if exc_value is not None:
             self.fail(exc_value)
         else:
             self.stop()
--- a/util/opentelemetry-util-genai/tests/test_invocation_base_exception.py
+++ b/util/opentelemetry-util-genai/tests/test_invocation_base_exception.py
@@ -1,0 +1,192 @@
+# Copyright The OpenTelemetry Authors
+# SPDX-License-Identifier: Apache-2.0
+
+"""Tests for how ``GenAIInvocation.__exit__`` finalizes on ``BaseException``.
+
+``asyncio.CancelledError``, ``KeyboardInterrupt``, ``SystemExit`` and
+``GeneratorExit`` derive from ``BaseException`` but not from ``Exception``. An
+``__exit__`` that only recognizes ``Exception`` finalizes them through the
+success path, so a cancelled operation is exported as a successful one.
+"""
+
+from __future__ import annotations
+
+import asyncio
+from unittest import TestCase
+
+import pytest
+
+from opentelemetry.sdk.trace import TracerProvider
+from opentelemetry.sdk.trace.export import SimpleSpanProcessor
+from opentelemetry.sdk.trace.export.in_memory_span_exporter import (
+    InMemorySpanExporter,
+)
+from opentelemetry.trace.status import StatusCode
+from opentelemetry.util.genai.handler import TelemetryHandler
+
+# Bounds every await in the cancellation scenarios below. The scenarios park on
+# ``asyncio.sleep(60)`` -- effectively forever -- so that a cancellation that
+# failed to propagate would otherwise stall the run instead of failing it.
+_TIMEOUT = 5
+
+
+class _InvocationExitTestBase(TestCase):
+    def setUp(self) -> None:
+        self.span_exporter = InMemorySpanExporter()
+        self.tracer_provider = TracerProvider()
+        self.tracer_provider.add_span_processor(
+            SimpleSpanProcessor(self.span_exporter)
+        )
+        self.handler = TelemetryHandler(tracer_provider=self.tracer_provider)
+
+    def _only_span(self):
+        spans = self.span_exporter.get_finished_spans()
+        self.assertEqual(len(spans), 1)
+        return spans[0]
+
+
+class InvocationExitBaseExceptionTest(_InvocationExitTestBase):
+    """A ``BaseException`` leaving the block must be recorded as a failure."""
+
+    def test_cancelled_error_sets_error_status(self) -> None:
+        with pytest.raises(asyncio.CancelledError):
+            with self.handler.workflow("wf"):
+                raise asyncio.CancelledError()
+
+        self.assertEqual(
+            self._only_span().status.status_code, StatusCode.ERROR
+        )
+
+    def test_cancelled_error_sets_error_type(self) -> None:
+        with pytest.raises(asyncio.CancelledError):
+            with self.handler.workflow("wf"):
+                raise asyncio.CancelledError()
+
+        self.assertEqual(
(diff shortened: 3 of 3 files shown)


util/opentelemetry-util-genai/src/opentelemetry/util/genai/_invocation.py
```diff
@@ -227,7 +227,15 @@ def __exit__(
...
     ) -> None:
-        if exc_value is not None and isinstance(exc_value, Exception):
+        # Guard on BaseException, not Exception: asyncio.CancelledError,
+        # KeyboardInterrupt, SystemExit and GeneratorExit do not derive from
+        # Exception, and treating them as success reports a cancelled or
+        # interrupted operation as a completed one. Matches the guard already
+        # used by the stream wrappers in ``stream.py``.
+        #
+        # Returning None (never a truthy value) is load-bearing: suppressing
+        # CancelledError here would break asyncio cancellation for every caller.
```
lmolkova:
```suggestion
```

I think this code is self-explanatory

util/opentelemetry-util-genai/tests/test_invocation_base_exception.py
```diff
@@ -0,0 +1,187 @@
+# Copyright The OpenTelemetry Authors
```
lmolkova, line 1: can you please write this test against the actual langgraph public API. I think it's an existing pattern, but this is very synthetic, does not show how it'd work in real life and we're trying to stop using it.

lmolkova (approved):
Thank you!

util/opentelemetry-util-genai/src/opentelemetry/util/genai/_invocation.py
ordemri: Agreed, applied. Comment block removed.

util/opentelemetry-util-genai/tests/test_invocation_base_exception.py
```diff
@@ -0,0 +1,187 @@
...
+            async def work() -> None:
+                with self.handler.workflow("wf"):
+                    started.set()
+                    await asyncio.sleep(60)
+
+            task = asyncio.ensure_future(work())
+            await started.wait()
+            task.cancel()
+            with pytest.raises(asyncio.CancelledError):
+                await task
```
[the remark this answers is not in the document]
ordemri: Good catch, applied. Both scenarios now bound the startup wait and the awaited
task with `asyncio.wait_for` on a shared 5s constant, at both sites.

The assertions are unchanged. I checked that `wait_for` is transparent here: when
the task is cancelled externally it still propagates `CancelledError` and
`task.cancelled()` is still true, on 3.10 through 3.14. I also checked the guard
can actually fire. With `task.cancel()` removed, the two tests fail in about 10s
total instead of stalling for two minutes.

util/opentelemetry-util-genai/tests/test_invocation_base_exception.py
```diff
@@ -0,0 +1,187 @@
...
+
+    def test_cancelled_error_sets_error_type(self) -> None:
+        with pytest.raises(asyncio.CancelledError):
+            with self.handler.workflow("wf"):
+                raise asyncio.CancelledError()
+
+        self.assertEqual(
+            self._only_span().attributes["error.type"],
+            "asyncio.exceptions.CancelledError",
+        )
```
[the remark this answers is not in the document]
ordemri: Declining this one.

Deriving the expected value with `fq_exception_type`, the same helper the code
under test uses to produce `error.type`, would make the assertion tautological. It
would pass even if `fq_exception_type` returned the wrong string, which is
precisely what the test exists to pin. The literal is the oracle.

On brittleness: I ran the string myself on CPython 3.10.21, 3.11.16, 3.12.2,
3.13.15 and 3.14.7, which cover every version in the package's classifiers. It is
`asyncio.exceptions.CancelledError` on all five. If it ever does change, this
assertion failing is the correct outcome, since `error.type` is an exported
attribute value and a change there is a change users would see.

util/opentelemetry-util-genai/tests/test_invocation_base_exception.py
ordemri: Happy to. I went to write it and hit something I wanted to check with you first.

langgraph is instrumented through the langchain package, and that instrumentation
is callback-based: `on_chain_end` calls `invocation.stop()` and `on_chain_error`
calls `invocation.fail(error)` directly. It never uses the invocation as a context
manager, so a langgraph-driven test doesn't reach `GenAIInvocation.__exit__`, and
it passes the same with and without this fix. `on_chain_error` is already typed to
take a `BaseException` and hands it straight to `fail()`.

The place a real-API cancellation test would land in `__exit__` is agno:
`_tool_call_aexecute` awaits inside `with _start_tool_invocation(...)`. Happy to
open that as a separate PR against agno.

Want me to add the langgraph test anyway, or go the agno route?

util/opentelemetry-util-genai/.changelog/520.fixed
```diff
@@ -0,0 +1 @@
+Record cancellation as a failure: the invocation context manager finalized any non-`Exception` `BaseException` (`asyncio.CancelledError`, `KeyboardInterrupt`, `SystemExit`, `GeneratorExit`) through the success path, so cancelled operations were exported with an unset span status and no `error.type`.
```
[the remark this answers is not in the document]
ordemri: Already fixed. The fragment is `.changelog/520.fixed` on the current head; this
review ran against the first push.

util/opentelemetry-util-genai/tests/test_invocation_base_exception.py
lmolkova: thanks for the research! 

Let's add langgraph test anyway - it'd assert the behavior is correct regardless of this PR and would be awesome to have a similar fix in the agno. If you believe there is a good candidate for common code to put in util-genai to make it easier (or maybe an additional guidance we can add to agents.md and review instructions), please share!

lzchen (approved):

util/opentelemetry-util-genai/tests/test_invocation_base_exception.py
lmolkova: let's follow up in #630

PR #522, opened 2026-09-01
fix(genai-openai): do not drain generator content; capture inline base64 file_data

### LINKED ISSUE #521: genai-openai: multimodal content-part arrays are dropped from captured messages (including the text part)
**Describe your environment**

- `opentelemetry-instrumentation-genai-openai` 1.1b0 (also reproduced on current `main`)
- `opentelemetry-util-genai` 1.1b0
- openai SDK 2.x, Python 3.12
- Content capture opted in: `OTEL_INSTRUMENTATION_GENAI_CAPTURE_MESSAGE_CONTENT=SPAN_ONLY`

**What happened?**

When a chat completion request uses the OpenAI content-part array form for multimodal messages, e.g.

```python
messages = [
    {"role": "user", "content": [
        {"type": "text", "text": "What is in this image?"},
        {"type": "image_url", "image_url": {"url": "https://example.com/cat.png"}},
    ]},
]
```

the captured `gen_ai.input.messages` records the user message with **empty `parts`** — the text part is silently dropped along with the image part:

```json
[{"role":"user","parts":[]}]
```

**What did you expect to see?**

At minimum the `text` part should be preserved. Ideally non-text parts are represented too (there is already a `GenericPart` type in `opentelemetry-util-genai` documented for exactly this: provider-specific part types that don't map to the standard semconv part types).

**Root cause**

`_prepare_input_messages` in `instrumentation/opentelemetry-instrumentation-genai-openai/src/opentelemetry/instrumentation/genai/openai/utils.py` only appends a part when `_is_text_part(content)` is true, and `_is_text_part` accepts only `str` or an iterable of `str`:

```python
def _is_text_part(content: Any) -> bool:
    return isinstance(content, str) or (
        isinstance(content, Iterable)
        and all(isinstance(part, str) for part in content)
    )
```

A content-part array (list of dicts / SDK part models) fails this check, and there is no fallback branch, so the message ends up with no parts at all.

**Additional context**

This is a regression in observability coverage for anyone migrating from third-party OpenAI instrumentations, which serialized the full content-part array. I have a fix ready (map `text` parts to `TextPart`, preserve other part types as `GenericPart`) and will open a PR referencing this issue.

FeelKyoun:
# Description

Rebased on `main` now that #358 has landed, and reduced to the delta it did not cover. Two fixes remain: `_content_to_parts` walked any `Iterable`, so generator `content` was drained by the instrumentation before the wrapped call and the request went out empty; and `_document_to_part` accepted only a `data:` URL for `file_data`, while the SDK documents the field as "The base64 encoded file data" — so the documented shape was dropped and only the undocumented one captured. `filename` now supplies the media type for both the `file_id` reference and the inline blob.

Everything else from the original PR is upstream already (#540, #358) and has been dropped.

## Type of change

# How has this been tested?

Five unit tests in `test_content_parts_unit.py`; three of them fail against `main` as it stands, and the other two are guard tests that pass today only because the part is dropped entirely (malformed base64 `file_data`, and the `data:` URL media type winning over `filename`).

`precommit` and `typecheck` clean; `-latest` 335 passed, `-v2` 339 passed, `-oldest` 157 passed, `-conformance` 12 passed (weaver 0.26.1).

# Checklist

- [x] Unit tests added
- [ ] Documentation updated (n/a)

### DIFF
--- a/instrumentation/opentelemetry-instrumentation-genai-openai/src/opentelemetry/instrumentation/genai/openai/utils.py
+++ b/instrumentation/opentelemetry-instrumentation-genai-openai/src/opentelemetry/instrumentation/genai/openai/utils.py
@@ -200,12 +200,6 @@ def _is_text_part(content: Any) -> bool:
     )
 
 
-_AUDIO_MIME_TYPES: Mapping[str, str] = {
-    "mp3": "audio/mpeg",
-    "wav": "audio/wav",
-}
-
-
 def _as_plain_value(item: Any) -> Any:
     """Return a JSON-safe representation of a content part.
 
@@ -242,24 +236,6 @@ def _image_url_part(item: Any) -> MessagePart | None:
     return image_from_url(url)
 
 
-def _input_audio_part(item: Any) -> MessagePart | None:
-    """Map an ``input_audio`` content part to an audio ``BlobPart``."""
-    input_audio = get_property_value(item, "input_audio")
-    data = get_property_value(input_audio, "data")
-    if not isinstance(data, str) or not data:
-        return None
-    if data.startswith("data:"):
-        return image_from_url(data, modality="audio")
-    content = decode_base64(data)
-    if content is None:
-        return None
-    audio_format = get_property_value(input_audio, "format")
-    mime_type = None
-    if isinstance(audio_format, str):
-        mime_type = _AUDIO_MIME_TYPES.get(audio_format.lower())
-    return BlobPart(mime_type=mime_type, modality="audio", content=content)
-
-
 def _file_part(item: Any) -> MessagePart | None:
     """Map a ``file`` content part to a ``FilePart`` (``file_id``) or a
     document ``BlobPart`` (inline ``file_data``, base64 or data URL)."""
@@ -286,7 +262,8 @@ def _file_part(item: Any) -> MessagePart | None:
 
 def _convert_content_part(item: Any) -> MessagePart | None:
     """Map one OpenAI content part to a semconv message part; typed parts
-    with no semconv mapping become ``GenericPart`` rather than being dropped."""
+    with no semconv mapping become ``GenericPart`` rather than being dropped.
+    ``input_audio`` parts are the exception and are never captured."""
     if isinstance(item, str):
         return TextPart(content=item)
     item_type = get_property_value(item, "type")
@@ -298,7 +275,9 @@ def _convert_content_part(item: Any) -> MessagePart | None:
     if item_type == "image_url":
         return _image_url_part(item)
     if item_type == "input_audio":
-        return _input_audio_part(item)
+        # Inline audio is intentionally not captured: a single clip can be
+        # megabytes, and it would be base64-inlined into the span attribute.
+        return None
     if item_type == "file":
         return _file_part(item)
     return GenericPart(type=item_type, value=_as_plain_value(item))
--- a/instrumentation/opentelemetry-instrumentation-genai-openai/src/opentelemetry/instrumentation/genai/openai/utils.py
+++ b/instrumentation/opentelemetry-instrumentation-genai-openai/src/opentelemetry/instrumentation/genai/openai/utils.py
@@ -24,7 +24,6 @@ from opentelemetry.util.genai.invocation import (
     InferenceInvocation,
 )
 from opentelemetry.util.genai.types import (
-    BlobPart,
     FilePart,
     FunctionToolDefinition,
     GenericPart,
@@ -36,7 +35,7 @@ from opentelemetry.util.genai.types import (
     ToolCallResponsePart,
     ToolDefinition,
 )
-from opentelemetry.util.genai.utils import decode_base64, image_from_url
+from opentelemetry.util.genai.utils import image_from_url
 
 _logger = logging.getLogger(__name__)
 
@@ -237,33 +236,27 @@ def _image_url_part(item: Any) -> MessagePart | None:
 
 
 def _file_part(item: Any) -> MessagePart | None:
-    """Map a ``file`` content part to a ``FilePart`` (``file_id``) or a
-    document ``BlobPart`` (inline ``file_data``, base64 or data URL)."""
+    """Map a ``file`` content part with a ``file_id`` to a ``FilePart``.
+
+    Inline ``file_data`` is intentionally not captured: a single document can
+    be megabytes, and it would be base64-inlined into the span attribute.
+    """
     file_ref = get_property_value(item, "file")
+    file_id = get_property_value(file_ref, "file_id")
+    if not isinstance(file_id, str) or not file_id:
+        return None
     filename = get_property_value(file_ref, "filename")
     mime_type = None
     if isinstance(filename, str) and filename:
         mime_type = mimetypes.guess_type(filename)[0]
-    file_id = get_property_value(file_ref, "file_id")
-    if isinstance(file_id, str) and file_id:
-        return FilePart(
-            mime_type=mime_type, modality="document", file_id=file_id
-        )
-    file_data = get_property_value(file_ref, "file_data")
-    if not isinstance(file_data, str) or not file_data:
-        return None
-    if file_data.startswith("data:"):
-        return image_from_url(file_data, modality="document")
-    content = decode_base64(file_data)
-    if content is None:
-        return None
-    return BlobPart(mime_type=mime_type, modality="document", content=content)
+    return FilePart(mime_type=mime_type, modality="document", file_id=file_id)
 
 
 def _convert_content_part(item: Any) -> MessagePart | None:
     """Map one OpenAI content part to a semconv message part; typed parts
     with no semconv mapping become ``GenericPart`` rather than being dropped.
-    ``input_audio`` parts are the exception and are never captured."""
+    Inline media payloads (``input_audio``, ``file_data``) are the exception
+    and are never captured."""
     if isinstance(item, str):
         return TextPart(content=item)
     item_type = get_property_value(item, "type")
--- a/instrumentation/opentelemetry-instrumentation-genai-openai/src/opentelemetry/instrumentation/genai/openai/utils.py
+++ b/instrumentation/opentelemetry-instrumentation-genai-openai/src/opentelemetry/instrumentation/genai/openai/utils.py
@@ -4,7 +4,9 @@
 from __future__ import annotations
 
 import json
-from collections.abc import Callable, Iterable, Mapping
+import logging
+import mimetypes
+from collections.abc import Iterable, Mapping, Sequence
 from typing import Any
 from urllib.parse import urlparse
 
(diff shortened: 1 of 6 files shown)


instrumentation/opentelemetry-instrumentation-genai-openai/src/opentelemetry/instrumentation/genai/openai/utils.py
```diff
@@ -206,6 +242,8 @@ def _prepare_input_messages(messages) -> list[InputMessage]:
                 chat_message.parts += extract_tool_calls_new(tool_calls)
             if _is_text_part(content):
                 chat_message.parts.append(TextPart(content=str(content)))
+            elif content is not None and isinstance(content, Iterable):
+                chat_message.parts += _extract_content_parts(content)
```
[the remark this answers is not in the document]
FeelKyoun: Good catch — addressed in the latest commit. `content` is now routed through a single `_content_to_parts` helper that materializes non-string, non-mapping iterables exactly once before branching, so single-pass iterables are no longer partially consumed, and mappings keep the previous `_is_text_part` behavior (string-keyed mappings captured as `str(mapping)`) instead of being iterated as content-part arrays. Added unit tests for both cases (generator content and mapping content).

instrumentation/opentelemetry-instrumentation-genai-openai/src/opentelemetry/instrumentation/genai/openai/utils.py
```diff
@@ -191,6 +192,68 @@ def _is_text_part(content: Any) -> bool:
...
+            parts.append(TextPart(content=item))
+            continue
+        item_type = get_property_value(item, "type")
+        if item_type == "text":
+            parts.append(
+                TextPart(content=str(get_property_value(item, "text") or ""))
+            )
+        else:
+            parts.append(
+                GenericPart(type=str(item_type), value=_as_plain_value(item))
```
lmolkova: `image_url` parts should map to semconv standard media parts (`UriPart` or `BlobPart`) rather than `GenericPart`.

`opentelemetry-util-genai` provides the `image_from_url` helper in `opentelemetry.util.genai.utils` for this exact purpose (it handles external URLs as `UriPart` and base64 data URLs as `BlobPart`). Please reuse `image_from_url(url)` here. `GenericPart` should only be used for provider-specific parts that have no standard semconv representation.

instrumentation/opentelemetry-instrumentation-genai-openai/src/opentelemetry/instrumentation/genai/openai/utils.py
```diff
@@ -191,6 +192,68 @@ def _is_text_part(content: Any) -> bool:
...
+    iterables are materialized exactly once — ``content`` may be a
+    single-pass iterable, and type-checking then re-iterating would silently
+    drop the items already consumed — then map to a ``TextPart`` when
+    string-only, or through ``_extract_content_parts`` otherwise.
+    """
+    if isinstance(content, str):
+        return [TextPart(content=content)]
+    if content is None:
+        return []
+    if isinstance(content, Mapping):
```
lmolkova: Stringifying mappings as `TextPart(content=str(content))` turns arbitrary dicts into malformed text parts like `TextPart(content="{'unexpected': 'shape'}")`.

Similarly, `all(isinstance(item, str) for item in items): return [TextPart(content=str(items))]` produces stringified Python list syntax (`TextPart(content="['a', 'b']")`). Iterables should route through `_extract_content_parts` so string items become individual `TextPart` instances.

lmolkova (changes requested):
Thank for the contribution! Please use specialized parts for images, URLs and others. 

Also please consider adding conformance test for new modalities - this is where we validate value schema.

instrumentation/opentelemetry-instrumentation-genai-openai/src/opentelemetry/instrumentation/genai/openai/utils.py
FeelKyoun: Done in d7e6b9a — `image_url` parts now go through `image_from_url(url)`: an external URL becomes a `UriPart` and a `data:` URL becomes a `BlobPart`.

While at it I mapped the other standard OpenAI parts too, so `GenericPart` is only used for typed parts that have no semconv representation:
- `input_audio` → audio `BlobPart` (mime type derived from `format`)
- `file` → `FilePart` for `file_id`, or a document `BlobPart` for inline `file_data`

instrumentation/opentelemetry-instrumentation-genai-openai/src/opentelemetry/instrumentation/genai/openai/utils.py
FeelKyoun: Agreed — that was a bad carry-over from the old `_is_text_part` behaviour. Both stringifications are gone in d7e6b9a:
- a bare mapping is treated as a single content part, so `{"type": "text", "text": ...}` becomes a `TextPart` and an untyped dict is dropped instead of becoming a bogus text part;
- string items in a list each become their own `TextPart`.

Everything now routes through one `_convert_content_part` and the iterable is consumed exactly once.

FeelKyoun:
Thanks for the review, @lmolkova — addressed in d7e6b9a:

- **Standard media parts instead of `GenericPart`** for `image_url` (`UriPart` / `BlobPart` via `image_from_url`), `input_audio` (audio `BlobPart`) and `file` (`FilePart` / document `BlobPart`); `GenericPart` is only used for typed parts with no semconv mapping. No more `str(dict)` / `str(list)` text parts (details in the threads).
- **Conformance test:** added `MultimodalScenario` to the openai conformance suite — a chat turn with text + an external image URL + an inline base64 PNG. The cassette is recorded against the real `gpt-4o-mini` API and the scenario validates with weaver live-check, asserting that `text`, `uri` (image) and `blob` (image) parts all land on the input message.

Verification: the full openai conformance suite (9 scenarios) passes locally in replay mode with weaver 0.25.1, the unit suite passes (260 passed), and pre-commit is clean.

`input_audio` and `file` are covered by unit tests but not by the conformance scenario, since they need an audio-capable model / an uploaded file. Happy to add those if you'd like them exercised there as well.

instrumentation/opentelemetry-instrumentation-genai-openai/src/opentelemetry/instrumentation/genai/openai/utils.py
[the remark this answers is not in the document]
FeelKyoun: Correction to my earlier reply here: "materialize once" was the wrong fix. The SDK accepts any iterable for `content` and materializes it itself, so consuming a generator inside the instrumentation (which runs before `wrapped()`) drained the caller's input and the SDK sent `content: []`. As of 7d339a8 only `Sequence` content is walked; other iterables are left untouched and simply not captured, and the unit test now asserts the generator is still unconsumed afterwards.

FeelKyoun:
Follow-up in 7d339a8 after a self-review pass on `utils.py` (no change to the part mapping you asked for, just hardening):

- **No request mutation.** Only `Sequence` content is walked as a content-part array. A generator (which the SDK accepts and materializes itself) was being drained before `wrapped()` ran, so the SDK sent empty content — that predates this PR for the first item, but my earlier "materialize once" change made it lose everything. Generator content is now left untouched and not captured (per the "No new side effects" rule in AGENTS.md).
- **Never raises into the caller.** Each part is converted under a guard and skipped on failure, and `GenericPart.value` is made JSON-safe (pydantic `model_dump(mode="json")` + `json` round-trip with `default=str`), so neither `create_chat_invocation` (which runs outside the `try` in `patch.py`) nor span export at `invocation.stop()` can fail on odd payloads.
- **`file_data` is base64 per the SDK docs**, not only a data URL: plain base64 is decoded into a document `BlobPart` (mime type from `filename`, also used for `FilePart.mime_type`); data URLs still work.
- `input_audio.data` given as a data URL is decoded instead of dropped; unmapped audio formats get `mime_type=None` rather than a synthesized `audio/<format>`.
- `get_property_value` accepts any `Mapping`; non-string `text` is dropped rather than stringified; the dispatch table is now an if-chain like the sibling instrumentations.

Unit tests: 24/24 (new cases for each of the above); full openai suite 270 passed; conformance suite (9 scenarios incl. `MultimodalScenario`) passes in replay mode; pre-commit clean.

instrumentation/opentelemetry-instrumentation-genai-openai/src/opentelemetry/instrumentation/genai/openai/utils.py
```diff
@@ -191,6 +200,135 @@ def _is_text_part(content: Any) -> bool:
...
+        else get_property_value(image_url, "url")
+    )
+    if not isinstance(url, str) or not url:
+        return None
+    return image_from_url(url)
+
+
+def _input_audio_part(item: Any) -> MessagePart | None:
+    """Map an ``input_audio`` content part to an audio ``BlobPart``."""
+    input_audio = get_property_value(item, "input_audio")
```
DylanRussell: I don't think we should put audio into telemetry.. it's going to be potentially huge right ?

FeelKyoun: Agreed. There is no size guard anywhere in util-genai and a single clip would be base64-inlined into `gen_ai.input.messages` (a span that large can fail the whole OTLP export batch), so inline audio is no longer captured as of 4f9f343: `input_audio` parts are explicitly skipped, rather than falling through to a `GenericPart` that would have embedded the base64 in `value`, while the text/image parts around them are still captured.

The same argument applies to inline `file_data` documents, so 4b0748a gives them the same treatment: `file` parts are captured only when they carry a `file_id` (as a `FilePart`), and inline `file_data` is skipped. For what it's worth, the anthropic instrumentation is the one place that inlines document bytes today; google-genai records documents as a `UriPart` reference only. @lmolkova, if "and others" in your review was meant to include inline audio and documents as a `BlobPart`, just say so and I'll restore those mappings.

The remaining inline case is data-URL images, which come from util-genai's `image_from_url` (as @lmolkova asked for above). I left that as-is since it's shared util behaviour rather than something this PR decides; if those shouldn't be inlined either, that seems like a util-genai-level call.

lmolkova:
Thanks for working on this! PR #358 is in final review and covers most of this area (audio, files, and refusals) on top of the multimodal changes merged in #540.

Let's wait for #358 to land first, and then rebase to address any remaining delta (such as MIME type inference from filename or generator handling).

FeelKyoun:
Thanks — #358 landed, so this is rebased on `main` and cut down to what it didn't cover. Everything else (audio, files, refusals, the image mapping) is upstream now and has been dropped.

Two things left:

- **Generator handling.** `_content_to_parts` walks any `Iterable`, and a message's `content` is typed `Union[str, Iterable[ChatCompletionContentPartParam]]`. Capturing runs before the wrapped call, so generator content was drained by the instrumentation and the request went out empty. Only a materialized `Sequence` is walked now.
- **`file_data` / `filename`.** `_document_to_part` accepts only a `data:` URL, but the SDK documents `file_data` as "The base64 encoded file data" — so the documented shape is dropped and only the undocumented one is captured. Plain base64 now decodes into a document `BlobPart`, and `filename` supplies the media type for both the `file_id` reference and the inline blob (a `data:` URL still takes it from the URL header).

lmolkova (approved):
thank you!

PR #528, opened 2026-09-01
util-genai & langchain: nest async/graph child spans under their caller

### LINKED ISSUE #513: langchain: spans are not nested — every step becomes its own root span/trace (async), and attach/detach raises "was created in a different Context"
**Component:** `opentelemetry-instrumentation-genai-langchain` (`1.2b0.dev`) + `opentelemetry-util-genai`
**Type:** bug

## Summary

When the LangChain callback handler is driven by an **async** LangGraph run (`graph.ainvoke`, and/or sync tools/models invoked under an async graph via a thread-pool), the emitted spans are **not nested**. Each GenAI step — `invoke_workflow`, every `chat`, every `execute_tool` — is exported as a **separate root span with its own trace id** (`parentId = null`). A single user interaction fragments into N one-span traces instead of one connected trace.

At runtime OpenTelemetry also logs:

```
Failed to detach context
ValueError: <Token var=<ContextVar name='current_context' ...>> was created in a different Context
```

## Root cause

Parent/child linkage relies **entirely on the ambient OTel context** via `attach`/`detach`, and the LangChain `parent_run_id` — which the handler already receives and stores — is **never** used to set the span parent.

`util/opentelemetry-util-genai/.../_invocation.py`:

```python
def _start(self, attributes=None):
    self.span = self._tracer.start_span(      # no context= -> parents from CURRENT contextvar
        name=self._span_name, kind=self._span_kind, attributes=attributes,
    )
    self._span_context = set_span_in_context(self.span)
    self._context_token = attach(self._span_context)   # attach on THIS callback
...
def _finish(self, error=None):
    ...
    detach(context_token)   # detach on a LATER, different callback
    self.span.end()
```

`start_span()` is called with **no `context=`**, so it takes whatever is in the ambient `contextvars` at that instant as the parent. `attach()` happens in `on_*_start` and the matching `detach()` happens in `on_*_end` — a *different* callback invocation.

`instrumentation/.../callback_handler.py` receives `parent_run_id` on every callback but only feeds it to bookkeeping (`add_invocation_state`, `_find_nearest_agent`) — it is never turned into a parent span context:

```python
self._invocation_manager.add_invocation_state(run_id, parent_run_id, llm_invocation)
```

Because the start/finish `attach`/`detach` pair is split across separate callbacks, and LangGraph runs nodes as their own asyncio tasks (and dispatches sync models/tools to thread-pool workers), the contextvar set by a parent's `attach()` is **not visible** in the child callback's execution context. So:

1. the child's `start_span` sees an empty context → new **root span → new trace**, and
2. the later `detach()` runs in a different `Context` than the `attach()` → `ValueError: ... created in a different Context` → "Failed to detach context".

`attach`/`detach` are LIFO, single-call-stack primitives; using them across separate, possibly-concurrent callbacks that hop asyncio tasks / threads is not sound. (Compare traceloop's `opentelemetry-instrumentation-langchain`, which keeps its own `run_id → span` map and explicitly threads the parent via `start_span(context=...)`; under the identical async graph it produces one correctly nested trace.)

## Steps to reproduce

Minimal supervisor graph with two react-agent workers, instrumented with `LangChainInstrumentor`, run with `graph.ainvoke`:

```python
LangChainInstrumentor().instrument(tracer_provider=provider)

# ... build a StateGraph supervisor + weather/finance create_react_agent nodes ...
graph = builder.compile()

await graph.ainvoke({"messages": [{"role": "user",
    "content": "What's the weather in Paris, and convert 100 USD to EUR?"}]})
provider.shutdown()
```

(Full ~140-line repro available; it's a standard supervisor + `create_react_agent` weather/finance graph.)

## Observed

- **async** (`graph.ainvoke`): `Failed to detach context` logged; exported spans are all `parentId=null`, each in its own trace id. One interaction → many single-span traces.
- Span scope: `opentelemetry.util.genai.handler`.

## Expected

One trace per top-level invocation, with `chat` / `execute_tool` / agent spans nested under the workflow/agent span — regardless of sync vs. async execution.

## Suggested fix

Don't depend on ambient `attach`/`detach` for parent linkage across callbacks. Instead, look up the parent invocation by `parent_run_id`, and pass its span context explicitly:

```python
parent = self._invocation_manager.get_invocation(parent_run_id)
parent_ctx = parent._span_context if parent else None
span = tracer.start_span(name=..., kind=..., context=parent_ctx, attributes=...)
```

This makes nesting independent of which asyncio task / thread the callback fires on, and removes the cross-Context `detach` error. (Also see #491, which is the same `attach`/`detach` context-lifetime fragility surfacing via streaming.)

## Environment

- `opentelemetry-instrumentation-genai-langchain` 1.2b0.dev
- `opentelemetry-util-genai` (matching)
- langgraph V1.x (`create_react_agent`), langchain-openai, Python 3.14

sidsri14:
Fixes #513 (narrow slice).

## What

LangChain async/graph traces currently emit each step as its own root
trace because child callbacks run in LangChain's own task loop where the
ambient OpenTelemetry context doesn't carry the parent span. This makes
child spans (tool, chat, agent, workflow) start with an empty parent and
become their own root traces.

This PR:

- Adds an optional `parent_context` parameter to `GenAIInvocation` and the
  langchain-used handler factories (`workflow`, `invoke_local_agent`,
  `inference`, `tool`). When provided, the invocation's span parents
  explicitly to it instead of relying on the ambient context.
- In the langchain callback handler, resolves the nearest live ancestor
  invocation's span from `parent_run_id` (walking up the run hierarchy past
  intermediate chain nodes that emit no telemetry) and passes it to the
  factories. Child spans now nest under their caller and share its trace id.

The change is fully backward-compatible: when no `parent_context` is given
(or no ancestor invocation is found), behavior is unchanged.

## Known gaps

- Scoped to LangChain's four span-emitting handlers. Other Instrumentation
  libraries and invocation/factory types are untouched.
- The cross-Context `attach`/`detach` error OTel logs internally on async
  paths is not eliminated by this PR (it is logged, not raised); parenting
  is now correct regardless.
- The `agent`/`workflow` span in LangGraph's `create_agent` scenario is
  still classified as a workflow rather than an agent in some configurations;
  this PR fixes nesting, not the agent/workflow classification.

### DIFF
--- a/instrumentation/opentelemetry-instrumentation-genai-langchain/src/opentelemetry/instrumentation/genai/langchain/callback_handler.py
+++ b/instrumentation/opentelemetry-instrumentation-genai-langchain/src/opentelemetry/instrumentation/genai/langchain/callback_handler.py
@@ -97,13 +97,17 @@ class OpenTelemetryLangChainCallbackHandler(BaseCallbackHandler):
         emits no telemetry. Walk up the run hierarchy to the nearest live
         invocation and parent to its span so child spans nest under their
         caller instead of becoming their own root trace.
+
+        A parent is usable as long as its span context is valid even when the
+        span itself is not recorded (e.g. downstream sampling); parenting to it
+        keeps the trace tree's shape and depth correct.
         """
         current = parent_run_id
         visited: set[UUID] = set()
         while current is not None and current not in visited:
             visited.add(current)
             parent = self._invocation_manager.get_invocation(current)
-            if parent is not None and parent.span.is_recording():
+            if parent is not None and parent.span.get_span_context().is_valid:
                 return set_span_in_context(parent.span)
             current = self._invocation_manager.get_parent_run_id(current)
         return None
@@ -673,7 +677,9 @@ class OpenTelemetryLangChainCallbackHandler(BaseCallbackHandler):
         provider = meta.get("ls_vector_store_provider") or None
         request_model = meta.get("ls_embedding_model") or None
         retrieval = self._telemetry_handler.retrieval(
-            provider=provider, request_model=request_model
+            provider=provider,
+            request_model=request_model,
+            parent_context=self._parent_context(parent_run_id),
         )
         retrieval.query_text = query
         self._invocation_manager.add_invocation_state(
--- a/instrumentation/opentelemetry-instrumentation-genai-langchain/src/opentelemetry/instrumentation/genai/langchain/callback_handler.py
+++ b/instrumentation/opentelemetry-instrumentation-genai-langchain/src/opentelemetry/instrumentation/genai/langchain/callback_handler.py
@@ -17,6 +17,7 @@ from langchain_core.outputs import (
     LLMResult,
 )
 
+from opentelemetry.context import Context
 from opentelemetry.instrumentation.genai.langchain.invocation_manager import (
     _InvocationManager,
 )
@@ -38,6 +39,7 @@ from opentelemetry.instrumentation.genai.langchain.utils import (
     response_fields_from_generation,
     to_input_messages,
 )
+from opentelemetry.trace import set_span_in_context
 from opentelemetry.util.genai.handler import TelemetryHandler
 from opentelemetry.util.genai.invocation import (
     AgentInvocation,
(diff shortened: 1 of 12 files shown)


instrumentation/opentelemetry-instrumentation-genai-langchain/src/opentelemetry/instrumentation/genai/langchain/callback_handler.py
```diff
@@ -83,6 +66,26 @@ def __init__(self, telemetry_handler: TelemetryHandler) -> None:
...
+        invocation and parent to its span so child spans nest under their
+        caller instead of becoming their own root trace.
+        """
+        current = parent_run_id
+        visited: set[UUID] = set()
+        while current is not None and current not in visited:
+            visited.add(current)
+            parent = self._invocation_manager.get_invocation(current)
+            if parent is not None and parent.span.is_recording():
+                return set_span_in_context(parent.span)
```
rads-1996, line 111: The `is_recording()` gate discards valid non-recording parents and can produce orphan sampled children and incorrect span depth. Maybe use `get_span_context().is_valid`.

instrumentation/opentelemetry-instrumentation-genai-langchain/src/opentelemetry/instrumentation/genai/langchain/callback_handler.py
rads-1996, line 679: Retrieval invocation path is missing `parent_context`

Issue #348, opened 2026-08-03
openai instrumentation should have parity with langchain openai wrapper


lmolkova:
We should make sure features being added to langchain openai wrapper are added to openai instrumentation too

https://github.com/open-telemetry/opentelemetry-python-genai/pull/305
https://github.com/open-telemetry/opentelemetry-python-genai/pull/296

rads-1996:
@lmolkova Could you please assign this issue to me. Thanks!

rads-1996:
Anthropic should have feature parity with - https://github.com/open-telemetry/opentelemetry-python-genai/pull/296

PR #540, opened 2026-09-02
Capture OpenAI multimodal image content (image_url and Responses API input_image) as BlobPart, UriPart, or FilePart message parts

rads-1996:
# Description

Fixes # (https://github.com/open-telemetry/opentelemetry-python-genai/issues/348)

## Type of change

# How has this been tested?

configuration.

- [x]  uv run tox -e py312-test-instrumentation-genai-langchain -- -q
- [x]  uv run tox -e py312-test-instrumentation-genai-langchain-conformance -- -q
- [x]  uv run --python 3.12 tox -e lint-instrumentation-genai-langchain

# Checklist

- [x] Unit tests added
- [ ] Documentation updated

### DIFF
--- a/instrumentation/opentelemetry-instrumentation-genai-openai/src/opentelemetry/instrumentation/genai/openai/response_extractors.py
+++ b/instrumentation/opentelemetry-instrumentation-genai-openai/src/opentelemetry/instrumentation/genai/openai/response_extractors.py
@@ -17,6 +17,7 @@ from opentelemetry.semconv._incubating.attributes import (
 
 from ._raw_response import ParsableResponse
 from .utils import (
+    _content_to_parts,
     _openai_response_format_to_output_type,
     get_property_value,
     get_served_model,
@@ -240,25 +241,7 @@ def get_input_messages(
         if not isinstance(role, str):
             continue
 
-        name = _get_field(item, "name")
-        name_str = str(name) if name is not None else None
-
-        content = _get_field(item, "content")
-        if isinstance(content, str):
-            messages.append(
-                InputMessage(
-                    role=role,
-                    parts=[TextPart(content=content)],
-                    name=name_str,
-                )
-            )
-            continue
-
-        parts = []
-        for part in _get_sequence(content):
-            text = _get_field(part, "text")
-            if isinstance(text, str):
-                parts.append(TextPart(content=text))
+        parts = _content_to_parts(_get_field(item, "content"))
         if parts:
             messages.append(
                 InputMessage(role=role, parts=parts, name=name_str)
--- a/instrumentation/opentelemetry-instrumentation-genai-openai/src/opentelemetry/instrumentation/genai/openai/response_extractors.py
+++ b/instrumentation/opentelemetry-instrumentation-genai-openai/src/opentelemetry/instrumentation/genai/openai/response_extractors.py
@@ -241,6 +241,8 @@ def get_input_messages(
         if not isinstance(role, str):
             continue
 
+        name = _get_field(item, "name")
+        name_str = str(name) if name is not None else None
         parts = _content_to_parts(_get_field(item, "content"))
         if parts:
             messages.append(
--- a/instrumentation/opentelemetry-instrumentation-genai-openai/src/opentelemetry/instrumentation/genai/openai/utils.py
+++ b/instrumentation/opentelemetry-instrumentation-genai-openai/src/opentelemetry/instrumentation/genai/openai/utils.py
@@ -59,7 +59,7 @@ def get_served_model(headers: Mapping[str, str] | None) -> str | None:
 
 
 def get_property_value(obj, property_name):
-    if isinstance(obj, dict):
+    if isinstance(obj, Mapping):
         return obj.get(property_name, None)
 
     return getattr(obj, property_name, None)
@@ -220,7 +220,7 @@ def _content_to_parts(content: Any) -> list[MessagePart]:
             continue
 
         image_url = get_property_value(item, "image_url")
-        if isinstance(image_url, Mapping):
+        if not isinstance(image_url, str):
             image_url = get_property_value(image_url, "url")
         if isinstance(image_url, str) and image_url:
             image_part = image_from_url(image_url)
--- a/instrumentation/opentelemetry-instrumentation-genai-openai/src/opentelemetry/instrumentation/genai/openai/utils.py
+++ b/instrumentation/opentelemetry-instrumentation-genai-openai/src/opentelemetry/instrumentation/genai/openai/utils.py
@@ -251,24 +251,28 @@ def _prepare_input_messages(messages) -> list[InputMessage]:
             name=str(name) if name is not None else None,
         )
         chat_messages.append(chat_message)
+        parts: list[MessagePart] = []
 
         content = get_property_value(message, "content")
 
         if role == Role.ASSISTANT.value:
             tool_calls = get_property_value(message, "tool_calls")
             if tool_calls:
-                chat_message.parts += extract_tool_calls_new(tool_calls)
-            chat_message.parts += _content_to_parts(content)
+                parts += extract_tool_calls_new(tool_calls)
+            parts += _content_to_parts(content)
 
         elif role == Role.TOOL.value:
             tool_call_id = get_property_value(message, "tool_call_id")
-            chat_message.parts.append(
+            parts.append(
                 ToolCallResponsePart(id=tool_call_id, response=content)
             )
 
         else:
             # system, developer, user, fallback
-            chat_message.parts += _content_to_parts(content)
+            parts += _content_to_parts(content)
+
+        if parts:
+            chat_messages.append(InputMessage(role=str(role), parts=parts))
     return chat_messages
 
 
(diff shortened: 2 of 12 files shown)


instrumentation/opentelemetry-instrumentation-genai-openai/src/opentelemetry/instrumentation/genai/openai/utils.py
```diff
@@ -191,6 +194,50 @@ def _is_text_part(content: Any) -> bool:
     )
 
 
+def _content_to_parts(content: Any) -> list[MessagePart]:
+    if isinstance(content, str):
+        return [TextPart(content=content)]
+    if not isinstance(content, Iterable) or isinstance(content, Mapping):
```
lmolkova:
```suggestion
    if not isinstance(content, Iterable) or isinstance(content, (Mapping, bytes, bytearray)):
```
`bytes` and `bytearray` are iterable in Python (yielding `int`). Should they be excluded like `Mapping`?

instrumentation/opentelemetry-instrumentation-genai-openai/src/opentelemetry/instrumentation/genai/openai/utils.py
```diff
@@ -204,8 +251,7 @@ def _prepare_input_messages(messages) -> list[InputMessage]:
             tool_calls = get_property_value(message, "tool_calls")
             if tool_calls:
                 chat_message.parts += extract_tool_calls_new(tool_calls)
-            if _is_text_part(content):
-                chat_message.parts.append(TextPart(content=str(content)))
+            chat_message.parts += _content_to_parts(content)
```
lmolkova: In `response_extractors.py`, messages with empty `parts` are dropped (`if parts: messages.append(...)`). Should `_prepare_input_messages` also avoid appending messages when `parts` is empty?

instrumentation/opentelemetry-instrumentation-genai-openai/src/opentelemetry/instrumentation/genai/openai/response_extractors.py
```diff
@@ -203,18 +204,7 @@ def get_input_messages(
...
-                InputMessage(role=role, parts=[TextPart(content=content)])
-            )
-            continue
-
-        parts = []
-        for part in _get_sequence(content):
-            text = _get_field(part, "text")
-            if isinstance(text, str):
-                parts.append(TextPart(content=text))
+        parts = _content_to_parts(_get_field(item, "content"))
```
lmolkova: `_get_field` in this module duplicates `get_property_value` from `utils.py`. Can we reuse `get_property_value` here?

instrumentation/opentelemetry-instrumentation-genai-openai/src/opentelemetry/instrumentation/genai/openai/utils.py
```diff
@@ -205,8 +252,7 @@ def _prepare_input_messages(messages) -> list[InputMessage]:
             tool_calls = get_property_value(message, "tool_calls")
             if tool_calls:
                 chat_message.parts += extract_tool_calls_new(tool_calls)
```
lmolkova: In `response_extractors.py`, messages with empty `parts` are dropped (`if parts: messages.append(...)`). Should `_prepare_input_messages` also avoid appending messages when `parts` is empty?

instrumentation/opentelemetry-instrumentation-genai-openai/src/opentelemetry/instrumentation/genai/openai/utils.py
```diff
@@ -192,6 +195,50 @@ def _is_text_part(content: Any) -> bool:
     )
 
 
+def _content_to_parts(content: Any) -> list[MessagePart]:
+    if isinstance(content, str):
+        return [TextPart(content=content)]
+    if not isinstance(content, Iterable) or isinstance(content, Mapping):
```
lmolkova, line 202: byte array would also match iterable. 

Is there a known type(s) of content we can use instead of `Any`?

lmolkova (approved):
Would this work for streaming? Do we have any test coverage for streaming / async multi-modal content?

instrumentation/opentelemetry-instrumentation-genai-openai/src/opentelemetry/instrumentation/genai/openai/utils.py
rads-1996: Yes, that makes sense. I will add changes for it.

instrumentation/opentelemetry-instrumentation-genai-openai/src/opentelemetry/instrumentation/genai/openai/utils.py
rads-1996: It looks like we could use the `object` type although that does not directly prevent byte array to be rejected but makes the type check stricter than Any. With it we would still add a run time guard - 

```
if not isinstance(content, Iterable) or isinstance(
    content, (Mapping, bytes, bytearray)
):
    return []
```

aabmass (approved):

instrumentation/opentelemetry-instrumentation-genai-openai/src/opentelemetry/instrumentation/genai/openai/utils.py
aabmass: Can we look at the types that are actually passed from the OpenAI code being instrumented and narrow it down to those?

rads-1996: @aabmass, @lmolkova `_content_to_parts` receives raw caller-supplied values before OpenAI SDK validation, including request dictionaries, reusable Pydantic response objects, attribute-shaped objects, and invalid values it defensively ignores. Therefore, there isn't a conclusive list of types we can narrow it down to. Should we move ahead with the object approach and then defensively discard certain types such as bytes?

instrumentation/opentelemetry-instrumentation-genai-openai/src/opentelemetry/instrumentation/genai/openai/utils.py
rads-1996: Addressed.

rads-1996:
> Would this work for streaming? Do we have any test coverage for streaming / async multi-modal content?

Yes, this would work for streaming. I had missed adding the test coverage for it earlier. Added it now. Thanks for pointing it out.

instrumentation/opentelemetry-instrumentation-genai-openai/src/opentelemetry/instrumentation/genai/openai/utils.py
```diff
@@ -193,32 +196,75 @@ def _is_text_part(content: Any) -> bool:
...
+
+    parts: list[MessagePart] = []
+    for item in content:
+        if isinstance(item, str):
+            parts.append(TextPart(content=item))
+            continue
+
+        part_type = get_property_value(item, "type")
+        text = get_property_value(item, "text")
+        if part_type in ("text", "input_text") or (
```
lmolkova: Regression: the type allowlist drops Responses API `output_text` parts, so a prior assistant turn fed back into `input` disappears from `gen_ai.input.messages`. The old code took any part with a `text` field.

Fails on this branch, passes on main:

```python
def test_extract_input_messages_keeps_assistant_output_text(loaded_module):
    messages = loaded_module.get_input_messages(
        [
            {"role": "user", "content": [{"type": "input_text", "text": "Hi"}]},
            {
                "role": "assistant",
                "content": [
                    {
                        "type": "output_text",
                        "text": "Hello!",
                        "annotations": [],
                    }
                ],
            },
        ]
    )

    assert [(msg.role, msg.parts) for msg in messages] == [
        ("user", [TextPart(content="Hi")]),
        ("assistant", [TextPart(content="Hello!")]),
    ]
```

`ResponsesConversationScenario` uses `previous_response_id`, so nothing else covers multi-turn Responses input.

instrumentation/opentelemetry-instrumentation-genai-openai/tests/test_chat_completions.py
```diff
@@ -1152,6 +1154,28 @@ def test_chat_completion_streaming(
         )
 
 
+def test_chat_completion_streaming_captures_multimodal_input(
```
lmolkova, line 1204: Sync non-streaming multimodal is uncovered - async has `test_async_chat_completion_captures_multimodal_input`, sync only gets the streaming variant.

instrumentation/opentelemetry-instrumentation-genai-openai/src/opentelemetry/instrumentation/genai/openai/utils.py
lmolkova: with `not isinstance(content, Iterable)` I think it's fine - we won't create anything meaningful in the code below, but would handle it gracefully.

lmolkova (commented):
Thanks for the update! It seems there is a regression my AI bot found, can you please check ?

instrumentation/opentelemetry-instrumentation-genai-openai/src/opentelemetry/instrumentation/genai/openai/utils.py
rads-1996: Addressed.

instrumentation/opentelemetry-instrumentation-genai-openai/tests/test_chat_completions.py
rads-1996: Addressed.

lmolkova (approved):

PR #506, opened 2026-08-31
[`opentelemetry-instrumentation-genai-langchain`] Resolve message roles by class instead of BaseMessage.type

sfc-gh-zeningchen:
# Description

A message produced by streaming reported its LangChain class name as the role:
`gen_ai.output.messages` carried `"role": "AIMessageChunk"`.

Roles were mapped from `BaseMessage.type`. Every streaming `*Chunk` class
subclasses its non-chunk counterpart while overriding `type` with its own class
name, so the two stay distinguishable when deserialized. The class hierarchy
identifies the role and `type` does not, so this matches on the class.

The effect reaches beyond streamed spans. `convert_to_messages` passes an
`AIMessageChunk` through unchanged, so an application that keeps an accumulated
streamed reply in its history also emitted `"role": "AIMessageChunk"` in
`gen_ai.input.messages` of later non-streamed calls, and on agent and workflow
spans.

`role` drives the role badge and the System/User/Assistant/Tool filter in
consumers, so an affected row shows a raw Python class name and matches no
filter value.

The spec role values move to `opentelemetry-util-genai` as a `Role` enum, next
to `FinishReason`. semconv constrains roles in the message JSON schemas, which
are not code-generated into `opentelemetry-semconv`, so instrumentations have
been spelling the literals out; google-genai already keeps a private copy of
this enum. `InputMessage.role` and `OutputMessage.role` widen to `str | Role`,
mirroring `finish_reason`, since the schemas also admit provider-specific roles.

Fixes #504

## Type of change

- [x] Bug fix (non-breaking change which fixes an issue)

# How has this been tested?

- [x] A parametrized test in `tests/test_callback_handler.py` covers all ten
  LangChain message classes, each chunk class alongside its non-chunk parent.
- [x] `test_streamed_output_message_role_is_assistant` covers the role on a
  streamed `on_llm_end`.
- [x] In `opentelemetry-util-genai`, `TestRole` checks the values against the
  message schemas and that a member set on a message serializes as `"assistant"`
  rather than `"Role.ASSISTANT"`.

# Checklist

- [x] Followed the style guidelines of this project
- [x] Changelog updated if the change requires an entry
- [x] Unit tests added

Issue #504, opened 2026-08-31
langchain: message roles report the LangChain class name instead of a spec role


sfc-gh-zeningchen:
### Describe your environment

OS: Ubuntu 22.04
Python version: Python 3.12
Package version: opentelemetry-instrumentation-genai-langchain 1.1b1, opentelemetry-util-genai 1.1b0 (also reproduces on main)
GenAI library (e.g. anthropic, openai) and version: langchain-openai 1.6.0

### What happened?

Messages produced by streaming report their LangChain class name as the role, e.g. 
A streamed call emits `"role": "AIMessageChunk"` in `gen_ai.output.messages`
instead of `"assistant"`.

### Steps to Reproduce

```
llm = ChatOpenAI(model="gpt-4o-mini")
reply = sum(llm.stream("capital of France?"))
llm.invoke([HumanMessage("capital of France?"), reply, HumanMessage("and Italy?")])
```

### Expected Result

Roles are the values the GenAI message schemas define: system, user, assistant, tool.

### Actual Result

```
[{"role":"AIMessageChunk","parts":[{"content":"Welcome aboard!","type":"text"}],"finish_reason":"unknown"}]
```

### Would you like to implement a fix?

Yes

### Tip

<sub>[React](https://github.blog/news-insights/product-news/add-reactions-to-pull-requests-issues-and-comments/) with 👍 to help prioritize this issue. Please use comments to provide useful context, avoiding `+1` or `me too`, to help us triage it. Learn more [here](https://opentelemetry.io/community/end-user/issue-participation/).</sub>

PR #575, opened 2026-09-02
Adopt shared `Role` enum across instrumentations

### LINKED ISSUE #534: Adopt the shared Role enum across instrumentations
### What problem do you want to solve?

#506 adds a `Role` enum to `opentelemetry-util-genai` with the four values the
GenAI message schemas define. Once it lands, instrumentations can stop
spelling the literals out. Today each one has its own approach:
| Package | How the role is produced |
| --- | --- |
| langchain | Shared `Role`, as of #506 |
| google-genai | Its own private `Role` enum, same four values |
| smolagents | A local `_ROLE_MAP` dict of literals |
| openai | Provider passthrough, plus bare literals |
| anthropic | Provider passthrough (`role=message.role`) |
| agno, qwen-agent | Bare literals |

The schemas accept any string for `role`, so an off-spec value passes weaver
without complaint and only surfaces downstream, where `role` drives the role
badge and the System/User/Assistant/Tool filter. That is how
`"role": "AIMessageChunk"` reached langchain spans in #504.

### Describe the solution you'd like

Use `opentelemetry.util.genai.types.Role` wherever an instrumentation emits a
role it controls, and delete google-genai's private copy.

### Describe alternatives you've considered

Leave the literals in place. They are correct today, but nothing catches a
value drifting outside the set, which is what happened in #504.

### Additional Context

Blocked on #506, which adds the enum.

### Would you like to implement it?

Yes

### Tip

sfc-gh-zeningchen:
# Description

The GenAI message schemas constrain `role` to `system`, `user`, `assistant`, and `tool`, but they are not code-generated into `opentelemetry-semconv`, so each instrumentation spelled the values out its own way: a private `Role` enum in google-genai, a local `_ROLE_MAP` in smolagents, bare literals elsewhere. The schemas accept any string, so an off-spec value passes weaver and only surfaces downstream, where `role` drives the role badge and the System/User/Assistant/Tool filter — that is how `"role": "AIMessageChunk"` reached langchain spans in #504.

[#506](https://github.com/open-telemetry/opentelemetry-python-genai/pull/506) added a shared `Role` enum to `opentelemetry-util-genai` and adopted it in langchain. This PR adopts it in google-genai (deleting its private copy), smolagents, openai, agno, qwen-agent, bedrock, portkey, and llama-index; anthropic is passthrough-only and is unchanged. Only roles an instrumentation picks itself change — provider-supplied roles still pass through, and the enum values are identical to the literals they replace, so there is no behaviour change.

Portkey's `opentelemetry-util-genai` floor moves to `1.2b0.dev`, where `Role` was added, matching every other package that depends on it; its `oldest` tox env gains the same `-e util/opentelemetry-util-genai` workaround the others already carry.

Fixes #534.

## Type of change

- [x] Maintenance (non-breaking refactor, no behaviour change)

# How has this been tested?

No new tests: the change is a literal-for-enum substitution, and the existing suites already assert the emitted role values, so they are the regression check.

- [x] `latest` suites pass for google-genai, smolagents, agno, qwen-agent, portkey, openai, and llama-index
- [x] bedrock's `latest` suite fails identically on `main` and on this branch (15 pre-existing VCR failures), so it is unaffected
- [x] `tox -e ruff` clean
- [x] `tox -e changelog-preview` renders the portkey entry

# Checklist

### DIFF
--- a/instrumentation/opentelemetry-instrumentation-genai-qwen-agent/src/opentelemetry/instrumentation/genai/qwen_agent/utils.py
+++ b/instrumentation/opentelemetry-instrumentation-genai-qwen-agent/src/opentelemetry/instrumentation/genai/qwen_agent/utils.py
@@ -99,7 +99,7 @@ def find_tool_call_id(
         function_id = (
             _field_value(extra, "function_id") if extra is not None else None
         )
-        if _field_value(msg, "role") in ("function", "tool"):
+        if _field_value(msg, "role") in ("function", Role.TOOL.value):
             if function_id:
                 responded.add(str(function_id))
             continue
@@ -164,7 +164,7 @@ def convert_to_input_messages(
 
             # qwen-agent uses role="function" internally, but the DashScope
             # API converts it to role="tool"; handle both.
-            if role in ("function", "tool") and content:
+            if role in ("function", Role.TOOL.value) and content:
                 parts.append(
                     ToolCallResponsePart(
                         id=_tool_call_response_id(msg),
@@ -247,11 +247,11 @@ def convert_to_final_output_messages(
 
     for msg in reversed(messages):
         try:
-            role = _field_value(msg, "role") or "assistant"
+            role = _field_value(msg, "role") or Role.ASSISTANT.value
             function_call = _field_value(msg, "function_call")
             content = _field_value(msg, "content") or ""
 
-            if role in ("function", "tool") or function_call:
+            if role in ("function", Role.TOOL.value) or function_call:
                 continue
 
             if _extract_content_text(content):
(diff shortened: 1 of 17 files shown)


instrumentation/opentelemetry-instrumentation-genai-qwen-agent/src/opentelemetry/instrumentation/genai/qwen_agent/utils.py
```diff
@@ -217,7 +218,7 @@ def convert_to_output_messages(
 
             output_messages.append(
                 OutputMessage(
```
lmolkova, line 220: nit: one bare literal is left in this file - `convert_to_final_output_messages` (line 250) still defaults `role` to `"assistant"`. It only feeds the local filter and is never emitted, but it is the last one left in the repo, so converting it keeps the sweep complete.

lmolkova (approved):
thank you!

instrumentation/opentelemetry-instrumentation-genai-qwen-agent/src/opentelemetry/instrumentation/genai/qwen_agent/utils.py
sfc-gh-zeningchen: thanks for catching that, switched to use enum there as well

lzchen (approved):

PR #588, opened 2026-09-03
[`opentelemetry-instrumentation-genai-openai-agents`] Record `gen_ai.tool.call.arguments` on execute_tool spans

sfc-gh-zeningchen:
# Description

`execute_tool` spans never carried `gen_ai.tool.call.arguments`, even with
content capture enabled. The processor read `FunctionSpanData.input` in
`on_span_start`, but the agents library assigns it inside the
`with function_span(...)` block, so the read always saw `None`. This reads it in
`on_span_end` instead, next to the existing `output` read, and deserializes it
to the object semconv specifies.

`input` stays `None` when a run sets `RunConfig.trace_include_sensitive_data=False`,
so the attribute is absent for those runs. A string that does not parse to a
JSON object is recorded as-is.

## Type of change

# How has this been tested?

New `test_runner_records_tool_call_arguments` drives `Runner.run` against a
cassette and asserts the attribute on the exported span. New parametrized
`test_tool_arguments_value_shapes` covers the argument string shapes providers
produce, plus the `None` and empty cases. Three existing tests pre-set
`FunctionSpanData.input` at construction, which the real SDK never does and
which is why they passed while the attribute was missing; they now assign it
after `on_span_start`.

- [x] `openai-agents` suite: 30 passed at the declared floor
      (`openai-agents==0.3.3`) and at latest (`openai-agents==0.22.0`)
- [x] `py314-test-instrumentation-genai-openai_agents-conformance`: 1 passed

# Checklist

- [x] Unit tests added
- [ ] Documentation updated

### DIFF
--- a/instrumentation/opentelemetry-instrumentation-genai-openai-agents/src/opentelemetry/instrumentation/genai/openai_agents/processor.py
+++ b/instrumentation/opentelemetry-instrumentation-genai-openai-agents/src/opentelemetry/instrumentation/genai/openai_agents/processor.py
@@ -135,8 +135,10 @@ class GenAITracingProcessor(TracingProcessor):
         invocation = self._invocations.pop(span, None)
         if invocation is None:
             return
-        if isinstance(invocation, ToolInvocation) and isinstance(
-            span.span_data, FunctionSpanData
+        if (
+            isinstance(invocation, ToolInvocation)
+            and isinstance(span.span_data, FunctionSpanData)
+            and invocation.should_capture_content_on_span
         ):
             arguments = span.span_data.input
             if arguments:
--- a/instrumentation/opentelemetry-instrumentation-genai-openai-agents/src/opentelemetry/instrumentation/genai/openai_agents/processor.py
+++ b/instrumentation/opentelemetry-instrumentation-genai-openai-agents/src/opentelemetry/instrumentation/genai/openai_agents/processor.py
@@ -32,7 +32,9 @@ the openai SDK directly and produces those.
 
 from __future__ import annotations
 
+import json
 import weakref
+from collections.abc import Mapping
 from typing import Any
 
 from agents.tracing import Span, Trace, TracingProcessor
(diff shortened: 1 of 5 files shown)


instrumentation/opentelemetry-instrumentation-genai-openai-agents/src/opentelemetry/instrumentation/genai/openai_agents/processor.py
```diff
@@ -126,9 +135,14 @@ def on_span_end(self, span: Span[Any]) -> None:
...
             return
-        if isinstance(invocation, ToolInvocation) and isinstance(
-            span.span_data, FunctionSpanData
+        if (
+            isinstance(invocation, ToolInvocation)
+            and isinstance(span.span_data, FunctionSpanData)
+            and invocation.should_capture_content_on_span
         ):
+            arguments = span.span_data.input
+            if arguments:
```
lmolkova:
```suggestion
            if arguments and arguments.strip():
```
question: should whitespace-only strings be ignored like empty strings?

instrumentation/opentelemetry-instrumentation-genai-openai-agents/src/opentelemetry/instrumentation/genai/openai_agents/processor.py
```diff
@@ -126,9 +135,14 @@ def on_span_end(self, span: Span[Any]) -> None:
         invocation = self._invocations.pop(span, None)
         if invocation is None:
             return
-        if isinstance(invocation, ToolInvocation) and isinstance(
-            span.span_data, FunctionSpanData
+        if (
+            isinstance(invocation, ToolInvocation)
+            and isinstance(span.span_data, FunctionSpanData)
+            and invocation.should_capture_content_on_span
```
lmolkova, line 141: JFYI - we should probably switch to more generic check, but let's get this one in and I'll update https://github.com/open-telemetry/opentelemetry-python-genai/pull/593

lmolkova (approved):

PR #591, opened 2026-09-03
[`opentelemetry-instrumentation-genai-dspy`] Support copy and deepcopy on wrapped DSPy methods

DylanRussell:
Implements copy and deepcopy support on wrapped DSPy functions and bound methods.

DSPy optimizers and teleprompters (such as `dspy.compile`, `BootstrapFewShot`, and `MIPRO`) rely heavily on `copy.copy` and `copy.deepcopy` to clone module trees and bound method callables during prompt optimization. Without custom wrapper copy handlers, copying wrapped DSPy callables raises `NotImplementedError: object proxy must define __deepcopy__()`.

## Type of change

## Checklist
- [x] Unit tests added
- [x] Documentation updated

### DIFF
--- a/instrumentation/opentelemetry-instrumentation-genai-dspy/src/opentelemetry/instrumentation/genai/dspy/patch.py
+++ b/instrumentation/opentelemetry-instrumentation-genai-dspy/src/opentelemetry/instrumentation/genai/dspy/patch.py
@@ -49,7 +49,15 @@ _REACT_V2_MODULE = "dspy.predict.react_v2"
 _REACT_V2_CLASS = "ReActV2"
 
 
-class _CopyableBoundFunctionWrapper(BoundFunctionWrapper[Any, Any]):
+if TYPE_CHECKING:
+    _BoundFunctionWrapper = BoundFunctionWrapper[Any, Any]
+    _FunctionWrapper = FunctionWrapper[Any, Any]
+else:
+    _BoundFunctionWrapper = BoundFunctionWrapper
+    _FunctionWrapper = FunctionWrapper
+
+
+class _CopyableBoundFunctionWrapper(_BoundFunctionWrapper):
     """BoundFunctionWrapper that supports copy and deepcopy."""
 
     def __init__(
@@ -113,7 +121,7 @@ class _CopyableBoundFunctionWrapper(BoundFunctionWrapper[Any, Any]):
         )
 
 
-class _CopyableFunctionWrapper(FunctionWrapper[Any, Any]):
+class _CopyableFunctionWrapper(_FunctionWrapper):
     """FunctionWrapper that supports copy and deepcopy."""
 
     __bound_function_wrapper__ = _CopyableBoundFunctionWrapper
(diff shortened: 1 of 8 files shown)


instrumentation/opentelemetry-instrumentation-genai-dspy/src/opentelemetry/instrumentation/genai/dspy/patch.py
```diff
@@ -40,32 +47,146 @@
...
+            wrapped,
+            instance,
+            wrapper,
+            enabled,
+            binding,
+            parent,
+            owner,
+            *args,
+        )
+        super().__init__(*init_args, **kwargs)
```
lmolkova, line 85: The `owner` argument requires wrapt >= 1.17, but the package declares no wrapt bound and gets `wrapt >= 1.0.0` transitively from `opentelemetry-instrumentation`. On 1.14/1.15/1.16 every copy raises:

```
TypeError: FunctionWrapperBase() takes at most 6 arguments (7 given)
```

add `wrapt >= 1.17` to `dependencies` (portkey already declares wrapt). The `oldest` env won't catch this - wrapt is transitive, so `lowest-direct` still resolves the newest.

instrumentation/opentelemetry-instrumentation-genai-dspy/tests/test_instrumentor.py
```diff
@@ -35,3 +41,81 @@ def test_instrument_with_global_providers() -> None:
...
+    try:
+        react_v2_mod = import_module("dspy.predict.react_v2")
+        react_v2_cls = getattr(react_v2_mod, "ReActV2", None)
+        if react_v2_cls is not None:
+            react_v2 = react_v2_cls("question -> answer", tools=[tool])
+            assert copy.copy(react_v2) is not None
+            assert copy.deepcopy(react_v2) is not None
+            assert copy.copy(react_v2.forward) is not None
+            assert copy.deepcopy(react_v2.forward) is not None
+    except (ImportError, AttributeError):
```
lmolkova: The tests only copy a bound method directly, which is not what DSPy does. The case that actually breaks today is a module tree holding an instrumented bound method - e.g. the agent-as-tool pattern:

```python
sub = dspy.ReAct("question -> answer", tools=[...])
tool = dspy.Tool(func=sub.forward, name="subagent", desc="d")
outer = dspy.ReAct("question -> answer", tools=[tool])
copy.deepcopy(outer)   # NotImplementedError without this PR
```

Worth covering that instead of / in addition to the direct-copy cases, and asserting the copied module still emits spans when called (right now the ReAct copies are only checked for `is not None`).

Also the `try/except (ImportError, AttributeError): pass` around ReActV2 means that block silently asserts nothing on versions where it does exist and something else goes wrong - a `pytest.importorskip` would fail loudly instead.

lmolkova (approved):

instrumentation/opentelemetry-instrumentation-genai-dspy/tests/test_instrumentor.py
DylanRussell: Ack added a test like that and a test around reactv2 missing

instrumentation/opentelemetry-instrumentation-genai-dspy/src/opentelemetry/instrumentation/genai/dspy/patch.py
DylanRussell: Ack upgraded to 1.17

PR #593, opened 2026-09-03
[opentelemetry-util-genai] Add should_capture_content to GenAIInvocation and deprecate should_capture_content_on_spans

lmolkova:
## What does this change do?

Adds a `should_capture_content` boolean property to `GenAIInvocation` initialized by `TelemetryHandler`, and switches invocations and tool wrappers to it. Deprecates `should_capture_content_on_spans()` in `utils.py` and `should_capture_content_on_span` on `ToolInvocation`.

## Why?

Reading environment variables on every call is not awesome on the hot path. Env vars are harder to patch for tests and harder to migrate to declarative config in the future, so limiting number of times we read env vars is future-proof  and makes things easier.

In addition, checking only spans does not make sense because we probably want to have events for each operation eventually.


eternalcuriouslearner (approved):
I am
Assuming we need not change anything in other instrumentations. Please correct me if I am wrong.

lmolkova:
> I am Assuming we need not change anything in other instrumentations. Please correct me if I am wrong.

no, you're right, it's exhaustive

PR #594, opened 2026-09-03
[`opentelemetry-instrumentation-genai-dspy]` Instrument `dspy.Retrieve` to emit a retrieval span

DylanRussell:
Instruments `dspy.Retrieve.forward` to emit`retrieval` spans with query text, `top_k`, and retrieved document attributes.

### DIFF
--- a/instrumentation/opentelemetry-instrumentation-genai-dspy/src/opentelemetry/instrumentation/genai/dspy/patch.py
+++ b/instrumentation/opentelemetry-instrumentation-genai-dspy/src/opentelemetry/instrumentation/genai/dspy/patch.py
@@ -12,6 +12,7 @@ from copy import copy, deepcopy
 from importlib import import_module
 from typing import TYPE_CHECKING, Any, cast
 
+import dspy
 from wrapt import (
     BoundFunctionWrapper,
     FunctionWrapper,
@@ -167,7 +168,7 @@ def _wrap_function(
 
 def patch_dspy(handler: TelemetryHandler) -> None:
     """Apply patches to DSPy Tool, ReAct, and Retrieve classes."""
-    import dspy
+    import dspy.predict.react
 
     tool_module = dspy.Tool.__module__
     tool_name = dspy.Tool.__name__
@@ -226,7 +227,6 @@ def patch_dspy(handler: TelemetryHandler) -> None:
 
 def unpatch_dspy() -> None:
     """Remove patches from DSPy classes."""
-    import dspy
     import dspy.predict.react
 
     unwrap(dspy.Tool, "__call__")
@@ -453,8 +453,6 @@ def _start_retrieval_invocation(
 ) -> RetrievalInvocation:
     rm: Any = getattr(instance, "rm", None)
     if rm is None:
-        import dspy
-
         rm = getattr(dspy.settings, "rm", None)
 
     data_source_id: str | None = (
--- a/instrumentation/opentelemetry-instrumentation-genai-dspy/src/opentelemetry/instrumentation/genai/dspy/patch.py
+++ b/instrumentation/opentelemetry-instrumentation-genai-dspy/src/opentelemetry/instrumentation/genai/dspy/patch.py
@@ -455,6 +455,8 @@ def _start_retrieval_invocation(
     if rm is None:
         rm = getattr(dspy.settings, "rm", None)
 
+    # DSPy retrieval models lack a uniform identifier schema, so inspect common index and provider
+    # attributes across the Retrieve instance and configured RM.
     data_source_id: str | None = (
         getattr(instance, "data_source_id", None)
         or getattr(instance, "index_name", None)
--- a/instrumentation/opentelemetry-instrumentation-genai-dspy/src/opentelemetry/instrumentation/genai/dspy/patch.py
+++ b/instrumentation/opentelemetry-instrumentation-genai-dspy/src/opentelemetry/instrumentation/genai/dspy/patch.py
@@ -12,7 +12,6 @@ from copy import copy, deepcopy
 from importlib import import_module
 from typing import TYPE_CHECKING, Any, cast
 
-import dspy
 from wrapt import (
     BoundFunctionWrapper,
     FunctionWrapper,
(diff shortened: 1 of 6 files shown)


instrumentation/opentelemetry-instrumentation-genai-dspy/src/opentelemetry/instrumentation/genai/dspy/patch.py
```diff
@@ -180,6 +182,20 @@ def patch_dspy(handler: TelemetryHandler) -> None:
...
         )
 
+    retrieve_module = dspy.Retrieve.__module__
+    retrieve_name = dspy.Retrieve.__name__
+    _wrap_function(
+        retrieve_module,
+        f"{retrieve_name}.forward",
+        _retrieve_forward(handler),
+    )
+    if hasattr(dspy.Retrieve, "aforward"):
```
lmolkova: [`dspy.Retrieve`](https://github.com/stanfordnlp/dspy/blob/main/dspy/retrievers/retrieve.py#L43-L65) does not define `aforward` - it only has `__call__` and `forward`.

Because of this:
- The `_retrieve_aforward` wrapper and unwrap are dead code.
- `test_async_retrieve_execution` skips on real DSPy installs.
- The other two async tests only pass because they monkeypatch their own `aforward` onto the class.

Drop the async path until DSPy adds native async support for `Retrieve`.

instrumentation/opentelemetry-instrumentation-genai-dspy/src/opentelemetry/instrumentation/genai/dspy/patch.py
```diff
@@ -396,3 +416,174 @@ async def traced_method(
...
+
+    if passages is None:
+        return
+
+    documents: list[dict[str, Any]] = []
+    for psg in passages:
+        # Standard dspy.Retrieve flattens RM passages to strings via psg.long_text.
+        if isinstance(psg, str):
+            documents.append({"content": psg})
+        elif isinstance(psg, Mapping):
```
lmolkova: These branches are unreachable. [`Retrieve.forward`](https://github.com/stanfordnlp/dspy/blob/main/dspy/retrievers/retrieve.py#L63-L65) flattens RM passages to strings before returning:

```python
passages = [psg.long_text for psg in passages]
return Prediction(passages=passages)
```

The wrapper only ever receives `Prediction` with `list[str]`. The `Mapping`, `id`, and `score` branches never run.

RMs that carry IDs or scores (like [`DatabricksRM`](https://github.com/stanfordnlp/dspy/blob/main/dspy/retrievers/databricks_rm.py#L207)) override `forward()`, so patching `dspy.Retrieve.forward` does not intercept them anyway.

instrumentation/opentelemetry-instrumentation-genai-dspy/src/opentelemetry/instrumentation/genai/dspy/patch.py
```diff
@@ -396,3 +416,174 @@ async def traced_method(
...
+    # DSPy retrieval models lack a uniform identifier schema, so inspect common index and provider
+    # attributes across the Retrieve instance and configured RM.
+    data_source_id: str | None = (
+        getattr(instance, "data_source_id", None)
+        or getattr(instance, "index_name", None)
+        or (getattr(rm, "data_source_id", None) if rm is not None else None)
+        or (getattr(rm, "index_name", None) if rm is not None else None)
+        or (getattr(rm, "collection_name", None) if rm is not None else None)
+    )
+    provider: str | None = (
```
lmolkova: DSPy RM classes do not define `provider` or `provider_name` attributes (e.g. [`WeaviateRM`](https://github.com/stanfordnlp/dspy/blob/main/dspy/retrievers/weaviate_rm.py#L52) uses `_weaviate_collection_name`, [`DatabricksRM`](https://github.com/stanfordnlp/dspy/blob/main/dspy/retrievers/databricks_rm.py#L143) uses `databricks_index_name`). Probing them only works on the synthetic `_DummyRM` in tests and returns `None` in real DSPy usage.

Per the [semantic conventions for retrievals](https://github.com/open-telemetry/semantic-conventions-genai/blob/main/docs/gen-ai/gen-ai-spans.md#retrievals), `gen_ai.provider.name` is not required. Leave it unset unless mapping from known RM types.

instrumentation/opentelemetry-instrumentation-genai-dspy/src/opentelemetry/instrumentation/genai/dspy/patch.py
```diff
@@ -396,3 +416,174 @@ async def traced_method(
...
+        instance: Retrieve,
+        args: tuple[Any, ...],
+        kwargs: dict[str, Any],
+    ) -> Any:
+        invocation = _start_retrieval_invocation(
+            handler, instance, args, kwargs
+        )
+        with invocation:
+            result = wrapped(*args, **kwargs)
+            _set_retrieval_invocation_documents(handler, invocation, result)
```
lmolkova: nit: `_set_retrieval_invocation_documents` runs inside `with invocation:`, so if extraction raises (e.g. `str(psg)` on a passage with a broken `__str__`) the span is marked failed for a call that actually succeeded. Build the documents outside the context, or after the result is captured.

instrumentation/opentelemetry-instrumentation-genai-dspy/src/opentelemetry/instrumentation/genai/dspy/patch.py
DylanRussell: Ack.. removed

instrumentation/opentelemetry-instrumentation-genai-dspy/src/opentelemetry/instrumentation/genai/dspy/patch.py
DylanRussell: Good catch.. removing this dead code

instrumentation/opentelemetry-instrumentation-genai-dspy/src/opentelemetry/instrumentation/genai/dspy/patch.py
DylanRussell: Removing this dead code as well !

lmolkova (approved):

PR #596, opened 2026-09-03
Align opentelemetry-util-genai with latest GenAI semantic conventions

lmolkova:
## What does this change do?
Updates `opentelemetry-util-genai` to align with the latest GenAI semantic conventions, adding support for cache write tokens, per-modality breakdowns, reasoning level, continuation response IDs, conversation compaction, prompt template attributes, and dedicated agent/tool duration metrics. It also deprecates `agent.id`, `agent.version`, and cache token attributes on internal agent spans, changes `RetrievalInvocation.top_k` to an integer, and makes `OutputMessage.finish_reason` optional.

## Why?
GenAI semantic conventions introduced dedicated duration metrics for agent and tool invocations, renamed cache creation tokens, added multimodal token accounting, and refined agent attribute lifecycles.

## Any known gaps or concerns
Agent call count metrics (`gen_ai.invoke_agent.inference_calls`, `gen_ai.invoke_agent.tool_calls`) and event definitions are omitted.


util/opentelemetry-util-genai/.changelog/596.added
```diff
@@ -0,0 +1 @@
+Add support for cache write tokens, per-modality token breakdown, reasoning level, continuation response ID, conversation compaction, prompt template attributes, and agent/tool duration metrics.
```
lmolkova: need to call out breaking change - we stopped reporting some attributes and metrics. also each affected lib should have it in changelog - conformance test changes is a good place to find which libs where affected

instrumentation/opentelemetry-instrumentation-genai-langchain/.changelog/596.changed
```diff
@@ -0,0 +1 @@
+Record span attribute `gen_ai.usage.cache_write.input_tokens` instead of `gen_ai.usage.cache_creation.input_tokens` on inference spans, and stop setting span attribute `gen_ai.agent.id` on internal agent spans.
```
lmolkova: metric changes too

util/opentelemetry-util-genai/.changelog/596.deprecated
```diff
@@ -0,0 +1 @@
+Deprecate span attribute `gen_ai.usage.cache_creation.input_tokens` on inference spans in favor of `gen_ai.usage.cache_write.input_tokens`.
```
lmolkova: it's not deprecated, it's changed, API to set it is deprecated - both should be here

util/opentelemetry-util-genai/.changelog/596.removed
```diff
@@ -0,0 +1 @@
+Stop reporting metric `gen_ai.client.operation.duration` on internal agent, tool, and workflow invocations in favor of metrics `gen_ai.invoke_agent.duration`, `gen_ai.execute_tool.duration`, and `gen_ai.invoke_workflow.duration`. Stop reporting span attribute `gen_ai.conversation.compacted` on workflow and agent spans. Stop reporting span attributes `gen_ai.agent.id`, `gen_ai.agent.version`, and `gen_ai.usage.cache_*.input_tokens` on internal agent spans.
```
lmolkova: > Stop reporting span attribute `gen_ai.conversation.compacted` on workflow and agent spans.

is this right?

util/opentelemetry-util-genai/src/opentelemetry/util/genai/_agent_invocation.py
```diff
@@ -36,6 +43,24 @@ class AgentInvocation(GenAIInvocation):
         Internal span: https://github.com/open-telemetry/semantic-conventions/blob/main/docs/gen-ai/gen-ai-agent-spans.md#invoke-agent-internal-span
     """
 
+    agent_id: str | None = None
+    """
+    .. deprecated:: 1.3b0
```
lmolkova: makes no sense, it'd internal API for both - we should have it in handler (if necessary)

Issue #491, opened 2026-08-29
util-genai: streamed invocations leak the span and the OTel context when the caller walks away


lmolkova:
A stream the caller never drains and never closes leaks both the span and the OTel context.

Every stream wrapper derives from `SyncStreamWrapper` / `AsyncStreamWrapper`, so this is not package specific: openai, anthropic, portkey, google-genai, smolagents, qwen-agent.

### Steps to Reproduce

```python
import gc

from opentelemetry.sdk.trace import TracerProvider
from opentelemetry.sdk.trace.export import SimpleSpanProcessor
from opentelemetry.sdk.trace.export.in_memory_span_exporter import (
    InMemorySpanExporter,
)
from opentelemetry.util.genai.handler import TelemetryHandler
from opentelemetry.util.genai.stream import SyncStreamWrapper

exporter = InMemorySpanExporter()
provider = TracerProvider()
provider.add_span_processor(SimpleSpanProcessor(exporter))
handler = TelemetryHandler(tracer_provider=provider)
tracer = provider.get_tracer("test")


class Wrapper(SyncStreamWrapper):
    def __init__(self, stream, invocation):
        super().__init__(stream)
        self._self_inv = invocation

    def _process_chunk(self, chunk):
        pass

    def _on_stream_end(self):
        self._self_inv.stop()

    def _on_stream_error(self, error):
        self._self_inv.fail(error)


def chunks():
    yield "a"
    yield "b"


def abandon():
    inv = handler.invoke_local_agent(agent_name="A", request_model="m")
    for _ in Wrapper(chunks(), inv):
        break


abandon()
gc.collect()

with tracer.start_as_current_span("later"):
    pass

spans = {s.name: s for s in exporter.get_finished_spans()}
print("agent span ended:", "invoke_agent A" in spans)
print("later.parent:", spans["later"].parent)
```

### Expected Result

```
agent span ended: True
later.parent: None
```

### Actual Result

```
agent span ended: False
later.parent: SpanContext(trace_id=0x..., span_id=0x..., ...)
```

`later.parent` is the worse half: every span on that thread from then on reparents under an invocation that never ended.

### Additional context

Needs investigation, the constraints pull against each other. Whatever lands should satisfy:

- An abandoned stream ends its span, with the stream's duration, not the collector's.
- An abandoned stream leaves nothing attached to the caller's context.
[...]

PR #610, opened 2026-09-04
[`opentelemetry-instrumentation-genai-openai`] Record response telemetry for async `with_streaming_response`

### LINKED ISSUE #624: openai: async with_streaming_response.parse() produces a span with no response data
### Describe your environment

OS: Rocky Linux 9.8
Python version: Python 3.14.7
Package version: opentelemetry-instrumentation-genai-openai 1.2b0.dev (main)
GenAI library (e.g. anthropic, openai) and version: openai 3.1.0

### What happened?

`AsyncAPIResponse.parse()` is a coroutine, and it is the only response class
where that is true. `RawResponseStreamProxy.parse()` checks
`isinstance(parsed, (Stream, AsyncStream))`, so it rejects the coroutine, takes
its "not a stream we can drive" branch, and hands it back unwrapped. The caller
awaits an uninstrumented stream and no response telemetry is recorded.
Only the async client's `with_streaming_response` is affected; the sync client
and both `with_raw_response` paths return a stream synchronously and work.
This is the path the OpenAI Agents SDK takes for streamed runs
(`agents/models/openai_responses.py` resolves
`responses.with_streaming_response.create`, then awaits `parse()`), so every
`Runner.run_streamed` produces an empty `chat` span.

### Steps to Reproduce

```python
import asyncio
from openai import AsyncOpenAI
from opentelemetry.instrumentation.genai.openai import OpenAIInstrumentor
OpenAIInstrumentor().instrument()
async def main():
    client = AsyncOpenAI()
    async with client.responses.with_streaming_response.create(
        model="gpt-4o-mini", input="Say this is a test.", stream=True
    ) as raw_response:
        async for _event in await raw_response.parse():
            pass
asyncio.run(main())
```
Then inspect the finished span. `chat.completions.with_streaming_response`
behaves the same.

### Expected Result

The `chat` span carries the same response attributes as
`responses.create(..., stream=True)`: `gen_ai.output.messages`,
`gen_ai.usage.*`, `gen_ai.response.finish_reasons`, `gen_ai.response.id`,
`gen_ai.response.model`.

### Actual Result

None of them are set. The span carries request-side attributes only and ends
with status `UNSET`.

### Additional context

_No response_

Yes

### Tip

sfc-gh-zeningchen:
# Description

`RawResponseStreamProxy.parse()` assumed the wrapped response hands back a
stream. That holds for `LegacyAPIResponse` and `APIResponse`, but the async
client's `with_streaming_response` yields an `AsyncAPIResponse`, whose `parse()`
is a coroutine. The `isinstance` check rejected it, so the caller awaited an
uninstrumented stream: no `gen_ai.output.messages`, `gen_ai.usage.*`,
`gen_ai.response.finish_reasons`, `gen_ai.response.id` or
`gen_ai.response.model`. The close fallback still ended the span with status
`UNSET`, so it looked healthy while carrying nothing.

Of the four raw-response entry points, only this one is affected:

| client | entry point | response | `parse()` |
| --- | --- | --- | --- |
| `OpenAI` | `with_raw_response` | `LegacyAPIResponse` | sync |
| `AsyncOpenAI` | `with_raw_response` | `LegacyAPIResponse` | sync |
| `OpenAI` | `with_streaming_response` | `APIResponse` | sync |
| `AsyncOpenAI` | `with_streaming_response` | `AsyncAPIResponse` | **async** |

It is also the path the OpenAI Agents SDK takes for streamed runs
(`agents/models/openai_responses.py` resolves
`responses.with_streaming_response.create`, then awaits `parse()`), so every
`Runner.run_streamed` produced an empty chat span.

`parse()` now awaits the coroutine and wraps what it resolves to. It stays
awaitable, so callers are unchanged. `_raw_response` is shared, so this covers
chat completions and responses alike.

## Type of change

# How has this been tested?

New tests for `with_streaming_response.create()` on responses (sync and async)
and chat completions (async), plus three unit tests in
`test_raw_response_proxy.py` covering the awaitable branch: wrapping and
memoizing the awaited stream, handing back a non-stream untouched, and the
close fallback still finalizing when the coroutine is never awaited.

The two async tests fail on `main` and pass here; the sync one passes on both,
since that path was already correct.

- [x] `openai` suite: 252 passed on `py310-latest` and `py314-latest`
- [x] `py310-oldest`: 113 passed; the two Responses tests skip, as the API does
      not exist at the declared floor
- [x] `tox -e precommit` and `tox -e typecheck` clean

# Checklist

- [x] Unit tests added
- [ ] Documentation updated

Fixes #624


instrumentation/opentelemetry-instrumentation-genai-openai/src/opentelemetry/instrumentation/genai/openai/_raw_response.py
```diff
@@ -149,7 +150,22 @@ def parse(self, *args: Any, **kwargs: Any) -> object:
         # would consume it twice anyway.
         if self._self_parsed is not None:
             return self._self_parsed
-        stream = self.__wrapped__.parse(*args, **kwargs)
+        parsed = self.__wrapped__.parse(*args, **kwargs)
+        if inspect.isawaitable(parsed):
+            # Only ``AsyncAPIResponse`` (the async ``with_streaming_response``)
+            # has a coroutine ``parse()``.
+            return self._parse_awaited(parsed)
+        return self._wrap_parsed(parsed)
```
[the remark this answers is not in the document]
sfc-gh-zeningchen: good catch, fixed.

eternalcuriouslearner (approved):

eternalcuriouslearner:
great catch!! Just for my learning sake, how did you reproduce this?

sfc-gh-zeningchen:
> great catch!! Just for my learning sake, how did you reproduce this?

@eternalcuriouslearner thanks for the nudge! this should have had a tracking issue anyway, so I opened #624 with the symptom and reproduction steps.

instrumentation/opentelemetry-instrumentation-genai-openai/src/opentelemetry/instrumentation/genai/openai/_raw_response.py
```diff
@@ -148,8 +150,30 @@ def parse(self, *args: Any, **kwargs: Any) -> object:
...
+
+    async def _parsed_as_awaitable(self) -> object:
+        return self._self_parsed
+
+    async def _parse_awaited(self, pending: Awaitable[Any]) -> object:
+        stream = await pending
+        # parse() can hand out several coroutines before any of them resolves,
+        # so re-check the memo to keep one wrapper per response.
+        if self._self_parsed is not None:
```
lmolkova, line 202: Race: if the httpx response closes while this coroutine is still pending (e.g. the caller calls `parse()` without immediately awaiting it, as your own `test_close_before_awaiting_parse_finalizes_once` does), the close fallback already finalizes the span here, empty. When this coroutine then resolves anyway, `_wrap_parsed` still builds and returns a stream wrapper; draining it calls `invocation.stop()` again, which silently no-ops (span already ended), so whatever attributes it collected are dropped.

Fix: in `_wrap_parsed`, check whether the close fallback already fired before building the wrapper:
```python
def _wrap_parsed(self, stream: Any) -> object:
    if isinstance(stream, (Stream, AsyncStream)):
        if self._self_finalize is None:
            # Closed while parse() was pending; span is already gone,
            # wrapping now would only drop attributes silently.
            return stream
        self._self_parsed = self._self_wrap_stream(stream)
        return self._self_parsed
    ...
```
`_self_finalize` is only ever cleared by the close fallback, so this is a reliable signal without adding new state.

Regression test (fails on this branch, passes with the fix above):
```python
@pytest.mark.asyncio()
async def test_close_while_parse_pending_does_not_lose_the_stream():
    resume = asyncio.Event()
    stream = _FakeAsyncStream()

    class _SlowAsyncRawResponse(_AsyncRawResponse):
        async def parse(self, *, to=None):
            await resume.wait()
            return self._parse_result

    raw = _SlowAsyncRawResponse(stream)
    finalize_calls, wrap_calls = [], []
    proxy = RawResponseStreamProxy(
        raw,
        wrap_stream=lambda s: wrap_calls.append(s) or ("wrapped", s),
        finalize=lambda: finalize_calls.append(True),
    )

    pending = proxy.parse()
    raw.http_response.close()
    assert finalize_calls == [True]  # close fallback already finalized

    resume.set()
    result = await pending

    assert wrap_calls == []  # must not silently drop a live wrapper's data
    assert result is stream
```

sfc-gh-zeningchen: thanks for catching this! I might have a new idea, so currently for async case `self._self_parsed` is only set when the caller actually `await pending`, and because of that when the httpx response closes, `_finalize_close_fallback()` just closes the span, and the suggestion is avoiding wrapping a stream against that already-ended span.
It feels like we can do it the other way around? when `parse()` is called we can just mark a flag and use that flag in `_finalize_close_fallback()` to determine whether to call `_finalize_once()`, so that with the same case, `_finalize_once()` will not be called and the span will be open until the stream was consumed. That way we can save the data in span. Potentially this can leak span when no one awaits the coroutine, but that feels like a different issue and exists today in the sync case as well.

sfc-gh-zeningchen: Sorry for the deleted comment, I posted a suggestion and then went to verify it, and that turned out to be wrong. I've updated according to your suggestion, thanks!

instrumentation/opentelemetry-instrumentation-genai-openai/src/opentelemetry/instrumentation/genai/openai/_raw_response.py
```diff
@@ -148,9 +150,36 @@ def parse(self, *args: Any, **kwargs: Any) -> object:
...
+            if self._self_parse_returns_awaitable:
+                return self._parsed_as_awaitable()
             return self._self_parsed
-        stream = self.__wrapped__.parse(*args, **kwargs)
+        parsed = self.__wrapped__.parse(*args, **kwargs)
+        if inspect.isawaitable(parsed):
+            # Only ``AsyncAPIResponse`` (the async ``with_streaming_response``)
+            # has a coroutine ``parse()``.
+            self._self_parse_returns_awaitable = True
+            return self._parse_awaited(parsed)
```
lmolkova, line 192: This pulls the async path into #491: a stream the caller doesn't drain now emits no span at all, where before the close fallback still ended it (empty). Once `_self_parsed` is set the fallback steps aside and nothing else finalizes.

Fails here, passes on `main`:
```python
@pytest.mark.asyncio()
async def test_abandoned_async_streaming_response_still_emits_span(
    span_exporter, async_openai_client, instrument_with_content, vcr
):
    with vcr.use_cassette("test_chat_completion_with_raw_response_streaming.yaml"):
        async with async_openai_client.chat.completions.with_streaming_response.create(
            messages=USER_ONLY_PROMPT, model=DEFAULT_MODEL, stream=True,
            stream_options={"include_usage": True},
        ) as raw_response:
            async for _chunk in await raw_response.parse():
                break

```

#491 is hard in general because a plain stream gives no signal that the caller left. Here there is one - `__aexit__` closes the http response - and the anthropic package already uses it: see `_finalize_close_fallback` / `_afinalize_close_fallback` in `instrumentation/opentelemetry-instrumentation-genai-anthropic/src/opentelemetry/instrumentation/genai/anthropic/_raw_response.py`, which close the stream wrapper instead of suppressing themselves. Both stream wrappers guard on `_self_finalized`, so a drained stream is unaffected.

lmolkova (commented):
Thanks for the update! 

One more edge case and we should be good. I tend to ask AI to find edge cases and prove with failing test, it turns out to be relatively good at finding them around streaming.

instrumentation/opentelemetry-instrumentation-genai-openai/src/opentelemetry/instrumentation/genai/openai/_raw_response.py
sfc-gh-zeningchen: thanks! Rethinking this: instead of tracking the wrapper separately, I reused the existing `_self_finalize` callback. It starts as `invocation.stop` and is swapped for `wrapper.close` once `parse()` builds one, so whichever close hook fires just runs whatever is currently there. The field is cleared when the callback is taken, so finalization happens once and the re-entrant close finds nothing left to run.

instrumentation/opentelemetry-instrumentation-genai-openai/src/opentelemetry/instrumentation/genai/openai/_raw_response.py
```diff
@@ -125,21 +128,51 @@ async def _aclose(*args: object, **kwargs: object) -> object:
...
     def _finalize_once(self) -> None:
-        # ``stop()`` is not idempotent, so finalize at most once.
-        if self._self_finalize is not None:
-            finalize, self._self_finalize = self._self_finalize, None
+        finalize = self._take_finalize()
+        if finalize is None:
+            return
+        try:
             finalize()
+        except BaseException:  # pylint: disable=broad-exception-caught
```
lzchen: Nit: Should we narrow both handlers to `except Exception` so `BaseException` subclasses like `asyncio.CancelledError` can propagate?

sfc-gh-zeningchen: Good catch, thanks for pointing it out!

lmolkova (approved):

PR #617, opened 2026-09-04
fix(langchain): trace nested workflows

### LINKED ISSUE #578: [langchain] Add nested workflow span on on_chain_start
> [!NOTE]

Part of #543.

In `on_chain_start`, `_looks_like_workflow` currently suppresses child chains when `parent_run_id is not None`.

Semantic conventions specify that application-defined workflows should be reported even when nested (e.g. a sub-graph executed as a node of another graph).

- Map nested sub-graphs / workflows to GenAI `invoke_workflow` (`gen_ai.operation.name: "invoke_workflow"`) with span name `invoke_workflow {gen_ai.workflow.name}`
- Capture `gen_ai.workflow.name`
- Capture `gen_ai.input.messages` and `gen_ai.output.messages` when content capture is enabled

1fanwang:
# Description

Named LangGraph subgraphs disappear from traces, so their duration and outcome cannot be inspected separately. They now emit workflow spans while agent graphs keep one agent span and ordinary nested chains remain suppressed.

Graph entry points identify compiled graph roots, so inherited callback metadata does not create extra agent spans.

Fixes https://github.com/open-telemetry/opentelemetry-python-genai/issues/578

Async spans still lose their parents. https://github.com/open-telemetry/opentelemetry-python-genai/issues/513 tracks that defect; the parentage assertions remain explicitly xfailed.

## Type of change

- [ ] New feature (non-breaking change which adds functionality) (not applicable)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected) (not applicable)
- [ ] This change requires a documentation update (not applicable)

# How has this been tested?

## E2E

The reproducer uses real compiled graphs and the OpenTelemetry exporter with local message nodes. It requires no model service or credentials. The runtime was Python 3.12, LangChain 1.4.0, and LangGraph 1.2.11.

- [x] Test A: The compiled-graph reproducer shows the missing span before the fix and the emitted spans afterward.

| # | Scenario | Command | Result |
|---|---|---|---|
| 1 | Upstream source | Reproducer below | Only the outer workflow appears; the assertion fails. |
| 2 | This branch | Same reproducer | The nested workflow and bound agent appear once each. Sync parents are correct; async parents remain absent. |

Before, at https://github.com/open-telemetry/opentelemetry-python-genai/commit/60c59fa6147b8828be4cd03b8b4a584165408667:

```text
sync workflow: [('invoke_workflow outer', None)]
AssertionError: The nested graph span is missing or duplicated
upstream_repro_exit=1
```

After:

```text
sync workflow: [('invoke_workflow named_subgraph', 'invoke_workflow outer'), ('invoke_workflow outer', None)]
sync agent: [('invoke_agent my_agent', 'invoke_workflow outer'), ('invoke_workflow outer', None)]
```

<details>
<summary>Reproducer</summary>

Run from the repository root with its `py312-test-instrumentation-genai-langchain-latest` tox environment prepared.

```bash
.tox/py312-test-instrumentation-genai-langchain-latest/bin/python - <<'PY'
import asyncio
from typing import Any, TypedDict

from langchain_core.messages import AIMessage, BaseMessage, HumanMessage
from langgraph.graph import END, START, StateGraph
from opentelemetry.sdk.trace import TracerProvider
from opentelemetry.sdk.trace.export import SimpleSpanProcessor
from opentelemetry.sdk.trace.export.in_memory_span_exporter import InMemorySpanExporter
from opentelemetry.test_util_genai.instrumentor import instrument

class State(TypedDict):
    messages: list[BaseMessage]

def respond(_: State) -> State:
    return {"messages": [AIMessage(content="done")]}

def graph(node: Any, name: str) -> Any:
    builder = StateGraph(State)
    builder.add_node(node="step", action=node)
    builder.add_edge(start_key=START, end_key="step")
    builder.add_edge(start_key="step", end_key=END)
    return builder.compile(name=name)

for async_mode in (False, True):
    for case in ("workflow", "agent"):
        exporter = InMemorySpanExporter()
        tracer = TracerProvider()
        tracer.add_span_processor(SimpleSpanProcessor(exporter))
        child = graph(node=respond, name="named_subgraph")
        expected = "invoke_workflow named_subgraph"
        if case == "agent":
            child = child.with_config(metadata={"agent_name": "my_agent"})
            expected = "invoke_agent my_agent"
        outer = graph(node=child, name="outer")
        with instrument(
            LangChainInstrumentor(), tracer_provider=tracer,
            meter_provider=None, logger_provider=None,
            content_capture="SPAN_ONLY",
        ):
            inputs: State = {"messages": [HumanMessage(content="hello")]}
            if async_mode:
                asyncio.run(outer.ainvoke(input=inputs))
            else:
                outer.invoke(input=inputs)
        spans = exporter.get_finished_spans()
        names = {span.context.span_id: span.name for span in spans}
        print(f"{'async' if async_mode else 'sync'} {case}: "
              f"{[(span.name, names[span.parent.span_id] if span.parent else None) for span in spans]}",
              flush=True)
        assert sorted(span.name for span in spans) == sorted(
            [expected, "invoke_workflow outer"]
        ), "The nested graph span is missing or duplicated"
PY
```

</details>

<details>
<summary>Raw logs</summary>

After, at https://github.com/open-telemetry/opentelemetry-python-genai/commit/16e6267e77d5f62fa24b4c8b7f256f1970c61088:

```text
sync workflow: [('invoke_workflow named_subgraph', 'invoke_workflow outer'), ('invoke_workflow outer', None)]
sync agent: [('invoke_agent my_agent', 'invoke_workflow outer'), ('invoke_workflow outer', None)]
async workflow: [('invoke_workflow named_subgraph', None), ('invoke_workflow outer', None)]
async agent: [('invoke_agent my_agent', None), ('invoke_workflow outer', None)]
candidate_repro_exit=0
```

The async runs also emit the known context error:

```text
Failed to detach context
ValueError: <Token var=<ContextVar name='current_context' default={} at 0x7f8be7197560> at 0x7f8be677a800> was created in a different Context
```

</details>

The oldest dependency environment skips graph tests because it lacks LangGraph and the newer agent API.

# Checklist

- [x] Unit tests added
- [ ] Documentation updated (not applicable; no public API or configuration changes)

### DIFF
--- a/instrumentation/opentelemetry-instrumentation-genai-langchain/src/opentelemetry/instrumentation/genai/langchain/agent_context.py
+++ b/instrumentation/opentelemetry-instrumentation-genai-langchain/src/opentelemetry/instrumentation/genai/langchain/agent_context.py
@@ -1,17 +1,6 @@
 # Copyright The OpenTelemetry Authors
 # SPDX-License-Identifier: Apache-2.0
 
-"""Identify ``create_agent`` graph invocations from outside the callback API.
-
-LangChain callbacks cannot tell a nested ``create_agent`` root apart from any
-other named runnable invoked inside a tool: the enclosing agent's ``config`` is
-merged over the inner agent's own, so ``lc_agent_name`` and ``ls_integration``
-in the *callback* metadata describe the outer agent. The compiled graph's own
-bound config is never shadowed, so this module reads the marker there and has
-each graph entry point announce itself on a context stack that the callback
-handler consults when the root run starts.
-"""
-
 from __future__ import annotations
 
 from collections.abc import AsyncIterator, Callable, Iterator, Mapping
@@ -27,27 +16,19 @@ _REACT_AGENT_MODULE = "langgraph.prebuilt.chat_agent_executor"
 
 
 @dataclass
-class _PendingAgent:
-    """An agent graph that has started running but whose root run is not seen yet."""
-
+class _PendingGraph:
     name: str | None
+    is_agent: bool
     claimed: bool = False
 
 
-_pending: ContextVar[tuple[_PendingAgent, ...]] = ContextVar(
-    "otel_genai_pending_agents", default=()
+_pending: ContextVar[tuple[_PendingGraph, ...]] = ContextVar(
+    "otel_genai_pending_graphs", default=()
 )
 
 
-def claim_agent() -> _PendingAgent | None:
-    """Return the announcement if this run is a create_agent graph root.
-
-    The announcement is made as the graph starts, so the first chain run to see
-    it is the graph's root. Only the innermost announcement is claimable, and
-    only once, so internal nodes fall through to metadata-based classification.
-    The root run's own name is not checked - ``with_config(run_name=...)``
-    renames it without making it any less of an agent.
-    """
+def claim_graph() -> _PendingGraph | None:
+    """Return the announcement for the next graph root callback."""
     pending = _pending.get()
     if not pending:
         return None
@@ -98,14 +79,18 @@ def _agent_name(graph: Any) -> tuple[bool, str | None]:
     return False, None
 
 
-def _push(name: str | None) -> _PendingAgent:
-    """Announce ``name`` as the innermost running agent."""
-    entry = _PendingAgent(name)
+def _workflow_name(graph: Any) -> str | None:
+    name = getattr(graph, "name", None)
+    return str(name) if name else None
+
+
+def _push(name: str | None, is_agent: bool) -> _PendingGraph:
+    entry = _PendingGraph(name=name, is_agent=is_agent)
     _pending.set(_pending.get() + (entry,))
     return entry
 
 
-def _pop(entry: _PendingAgent) -> None:
+def _pop(entry: _PendingGraph) -> None:
     """Withdraw ``entry``, tolerating a stack the caller's context no longer owns."""
     pending = _pending.get()
     if pending and pending[-1] is entry:
@@ -118,14 +103,14 @@ def wrap_stream(
     args: tuple[Any, ...],
     kwargs: dict[str, Any],
 ) -> Any:
-    """Announce an agent graph for the duration of ``Pregel.stream``.
+    """Announce a graph for the duration of ``Pregel.stream``.
 
     ``Pregel.invoke`` runs through ``stream``, so this covers both entry points.
     """
     is_agent, name = _agent_name(instance)
     if not is_agent:
-        return wrapped(*args, **kwargs)
-    return _announce_at_stream_start(wrapped(*args, **kwargs), name)
+        name = _workflow_name(instance)
+    return _announce_at_stream_start(wrapped(*args, **kwargs), name, is_agent)
 
 
 def wrap_astream(
@@ -134,18 +119,20 @@ def wrap_astream(
     args: tuple[Any, ...],
     kwargs: dict[str, Any],
 ) -> Any:
-    """Announce an agent graph for the duration of ``Pregel.astream``.
+    """Announce a graph for the duration of ``Pregel.astream``.
 
     ``Pregel.ainvoke`` runs through ``astream``, so this covers both entry points.
     """
     is_agent, name = _agent_name(instance)
     if not is_agent:
-        return wrapped(*args, **kwargs)
-    return _announce_at_astream_start(wrapped(*args, **kwargs), name)
+        name = _workflow_name(instance)
+    return _announce_at_astream_start(wrapped(*args, **kwargs), name, is_agent)
 
 
 def _announce_at_stream_start(
-    stream: Iterator[Any], name: str | None
+    stream: Iterator[Any],
+    name: str | None,
+    is_agent: bool,
 ) -> Iterator[Any]:
     """Announce ``name`` for the first step of ``stream`` only.
 
@@ -158,7 +145,7 @@ def _announce_at_stream_start(
     # A generator body runs in the consumer's context, so the announcement lands
     # where the callbacks fire - on the first ``next()``, not here.
     iterator = iter(stream)
-    entry = _push(name)
+    entry = _push(name, is_agent)
     try:
         first = next(iterator)
     except StopIteration:
@@ -170,11 +157,13 @@ def _announce_at_stream_start(
 
 
 async def _announce_at_astream_start(
-    stream: AsyncIterator[Any], name: str | None
+    stream: AsyncIterator[Any],
+    name: str | None,
+    is_agent: bool,
 ) -> AsyncIterator[Any]:
     """Announce ``name`` for the first step of ``stream`` only."""
     iterator = stream.__aiter__()
-    entry = _push(name)
+    entry = _push(name, is_agent)
     try:
         first = await iterator.__anext__()
     except StopAsyncIteration:
--- a/instrumentation/opentelemetry-instrumentation-genai-langchain/src/opentelemetry/instrumentation/genai/langchain/agent_context.py
+++ b/instrumentation/opentelemetry-instrumentation-genai-langchain/src/opentelemetry/instrumentation/genai/langchain/agent_context.py
@@ -1,6 +1,13 @@
 # Copyright The OpenTelemetry Authors
 # SPDX-License-Identifier: Apache-2.0
 
+"""Identify graph roots outside the LangChain callback API.
+
+Callback metadata cannot distinguish a nested graph root from its internal
+runnables because parent config is inherited. Graph entry points announce the
+root on a context stack so only the matching callback can claim it.
+"""
+
 from __future__ import annotations
 
 from collections.abc import AsyncIterator, Callable, Iterator, Mapping
@@ -19,6 +26,7 @@ _REACT_AGENT_MODULE = "langgraph.prebuilt.chat_agent_executor"
 class _PendingGraph:
     name: str | None
     is_agent: bool
+    metadata: dict[str, Any]
     claimed: bool = False
 
 
@@ -28,7 +36,11 @@ _pending: ContextVar[tuple[_PendingGraph, ...]] = ContextVar(
 
 
 def claim_graph() -> _PendingGraph | None:
-    """Return the announcement for the next graph root callback."""
+    """Claim the innermost graph announcement for its root callback.
+
+    Only the newest announcement is claimable, and only once, so nested graphs
+    do not leak their classification to internal callbacks.
+    """
     pending = _pending.get()
     if not pending:
         return None
@@ -58,34 +70,50 @@ def _react_agent_name(graph: Any) -> str | None:
     return str(name) if name and name != "LangGraph" else ""
 
 
-def _agent_name(graph: Any) -> tuple[bool, str | None]:
+def _bound_metadata(graph: Any) -> dict[str, Any]:
+    config = getattr(graph, "config", None)
+    if not isinstance(config, Mapping):
+        return {}
+    metadata = cast("Mapping[str, Any]", config).get("metadata")
+    if not isinstance(metadata, Mapping):
+        return {}
+    return dict(cast("Mapping[str, Any]", metadata))
+
+
+def _agent_name(
+    graph: Any,
+    metadata: Mapping[str, Any],
+) -> tuple[bool, str | None]:
     """Return whether ``graph`` is an agent and its application-provided name."""
     config = getattr(graph, "config", None)
     create_agent_name = create_agent_graph_name(config)
     if create_agent_name:
         return True, create_agent_name
-    if isinstance(config, Mapping):
-        metadata = cast("Mapping[str, Any]", config).get("metadata")
-        if isinstance(metadata, Mapping):
-            typed_metadata = cast("Mapping[str, Any]", metadata)
-            if (
-                typed_metadata.get("ls_integration")
-                == "langchain_create_agent"
-            ):
-                return True, None
+    if metadata.get("ls_integration") == "langchain_create_agent":
+        return True, None
     react_name = _react_agent_name(graph)
     if react_name is not None:
         return True, react_name or None
+    if (
+        metadata.get("otel_agent_span")
+        or metadata.get("agent_type")
+        or metadata.get("agent_name")
+    ):
+        name = metadata.get("agent_name")
+        return True, str(name) if name else None
     return False, None
 
 
-def _workflow_name(graph: Any) -> str | None:
-    name = getattr(graph, "name", None)
-    return str(name) if name else None
-
-
-def _push(name: str | None, is_agent: bool) -> _PendingGraph:
-    entry = _PendingGraph(name=name, is_agent=is_agent)
+def _push(
+    name: str | None,
+    is_agent: bool,
+    metadata: dict[str, Any],
+) -> _PendingGraph:
+    entry = _PendingGraph(
+        name=name,
+        is_agent=is_agent,
+        metadata=metadata,
+    )
     _pending.set(_pending.get() + (entry,))
     return entry
 
@@ -107,10 +135,14 @@ def wrap_stream(
 
     ``Pregel.invoke`` runs through ``stream``, so this covers both entry points.
     """
-    is_agent, name = _agent_name(instance)
-    if not is_agent:
-        name = _workflow_name(instance)
-    return _announce_at_stream_start(wrapped(*args, **kwargs), name, is_agent)
+    metadata = _bound_metadata(instance)
+    is_agent, name = _agent_name(instance, metadata)
+    return _announce_at_stream_start(
+        wrapped(*args, **kwargs),
+        name,
+        is_agent,
+        metadata,
+    )
 
 
 def wrap_astream(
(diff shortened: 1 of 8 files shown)


instrumentation/opentelemetry-instrumentation-genai-langchain/src/opentelemetry/instrumentation/genai/langchain/callback_handler.py
```diff
@@ -103,25 +103,41 @@ def on_chain_start(
...
+            announced_workflow=announced_workflow,
         )
         conversation_id = _conversation_id(metadata)
         capture_content = self._telemetry_handler.should_capture_content()
         if operation == OperationName.INVOKE_WORKFLOW:
-            workflow_name = kwargs.get("name") or serialized.get("name")
+            workflow_name = (
+                kwargs.get("name")
+                or serialized.get("name")
+                or declared_workflow_name
```
lmolkova: If `serialized.get("name")` returns a default name (like `"LangGraph"`), it can shadow `declared_workflow_name`. Consider checking `declared_workflow_name` before fallback serialized names: `kwargs.get("name") or declared_workflow_name or serialized.get("name")`.

instrumentation/opentelemetry-instrumentation-genai-langchain/src/opentelemetry/instrumentation/genai/langchain/operation_mapping.py
```diff
@@ -276,13 +273,23 @@ def classify_chain_run(
...
     ):
         return OperationName.INVOKE_AGENT
 
-    # 3. Workflow / orchestration detection.
-    if _looks_like_workflow(serialized, metadata, parent_run_id):
+    if metadata and metadata.get(_META_WORKFLOW_SPAN):
+        return OperationName.INVOKE_WORKFLOW
+
+    # 3. A nested graph announcement is stronger than inherited agent metadata.
+    if announced_workflow and parent_run_id is not None:
```
lmolkova: Since `announced_workflow` confirms the graph is not an agent via `claim_graph()`, should it take precedence over `_has_agent_signals` consistently at both root and nested levels? If so, this and step 4 can be unified before `_has_agent_signals`.

instrumentation/opentelemetry-instrumentation-genai-langchain/tests/test_workflow.py
```diff
@@ -0,0 +1,145 @@
...
+    workflow_spans = _workflow_spans(span_exporter)
+    assert len(workflow_spans) == 2
+    workflow_spans_by_name = {
+        span.attributes[GenAI.GEN_AI_WORKFLOW_NAME]: span
+        for span in workflow_spans
+    }
+    assert set(workflow_spans_by_name) == {"LangGraph", "named_subgraph"}
+    inner_span = workflow_spans_by_name["named_subgraph"]
+    outer_span = workflow_spans_by_name["LangGraph"]
+    if not async_mode:
```
lmolkova: Is parentage expected to differ in async mode? If possible, we should assert span parentage in async mode as well or test the actual hierarchy.

instrumentation/opentelemetry-instrumentation-genai-langchain/src/opentelemetry/instrumentation/genai/langchain/agent_context.py
```diff
@@ -98,14 +79,18 @@ def _agent_name(graph: Any) -> tuple[bool, str | None]:
     return False, None
 
 
-def _push(name: str | None) -> _PendingAgent:
-    """Announce ``name`` as the innermost running agent."""
-    entry = _PendingAgent(name)
+def _workflow_name(graph: Any) -> str | None:
```
lmolkova: A subgraph that declares agent metadata on its own config now gets two spans instead of one - the graph root is announced as a workflow, and the agent signal is then picked up again by the inner node run:

```python
sub = _graph(_respond, name="my_agent").with_config({"metadata": {"agent_name": "my_agent"}})
_graph(sub).invoke({"messages": [HumanMessage(content="hello")]})

# main: invoke_workflow LangGraph > invoke_agent my_agent
# here: invoke_workflow LangGraph > invoke_workflow my_agent > invoke_agent my_agent
```

The graph's own bound config is the one place inheritance can't lie, so classify it here rather than in `classify_chain_run`:

```python
    react_name = _react_agent_name(graph)
    if react_name is not None:
        return True, react_name or None
    if isinstance(config, Mapping):
        metadata = cast("Mapping[str, Any]", config).get("metadata")
        if isinstance(metadata, Mapping):
            m = cast("Mapping[str, Any]", metadata)
            if m.get("otel_agent_span") or m.get("agent_type") or m.get("agent_name"):
                name = m.get("agent_name")
                return True, str(name) if name else None
    return False, None
```

With that, `classify_chain_run` needs no change and the existing tests still pass. Please add the case above as a test - agent metadata on the subgraph itself, not on the outer invoke.

instrumentation/opentelemetry-instrumentation-genai-langchain/src/opentelemetry/instrumentation/genai/langchain/callback_handler.py
lmolkova: `serialized` is `None` on every chain callback with current langgraph - this line only survives because `kwargs["name"]` is always set, and nested graphs now reach it. Combined with the ordering point above:

```python
            workflow_name = (
                kwargs.get("name")
                or declared_workflow_name
                or (serialized or {}).get("name")
            )
```

instrumentation/opentelemetry-instrumentation-genai-langchain/tests/test_workflow.py
lmolkova: In async the nested span is a second root span, not a child - that's #513, not a bug in this change. Please make it visible rather than skipping the assert:

```python
    if async_mode:
        pytest.xfail("nested spans are not parented in async - #513")
    assert inner_span.parent.span_id == outer_span.context.span_id
```

instrumentation/opentelemetry-instrumentation-genai-langchain/src/opentelemetry/instrumentation/genai/langchain/agent_context.py
```diff
@@ -1,17 +1,6 @@
...
-LangChain callbacks cannot tell a nested ``create_agent`` root apart from any
-other named runnable invoked inside a tool: the enclosing agent's ``config`` is
-merged over the inner agent's own, so ``lc_agent_name`` and ``ls_integration``
-in the *callback* metadata describe the outer agent. The compiled graph's own
-bound config is never shadowed, so this module reads the marker there and has
-each graph entry point announce itself on a context stack that the callback
-handler consults when the root run starts.
-"""
-
 from __future__ import annotations
```
lmolkova: nit: the module docstring and the body of `claim_graph`'s docstring explained why the announcement exists at all - callback metadata can't identify a nested graph root, only the innermost announcement is claimable, and only once. None of that is inferable from the code; please keep it, updated for graphs.

instrumentation/opentelemetry-instrumentation-genai-langchain/src/opentelemetry/instrumentation/genai/langchain/agent_context.py
lmolkova: A subgraph that declares agent metadata on its own config should be one `invoke_agent` span. It now gets two - the graph root is announced as a workflow, and the agent signal is picked up again by the inner node run:

```python
sub = _graph(_respond, name="my_agent").with_config({"metadata": {"agent_name": "my_agent"}})
_graph(sub).invoke({"messages": [HumanMessage(content="hello")]})

# main: invoke_workflow LangGraph > invoke_agent my_agent
# here: invoke_workflow LangGraph > invoke_workflow my_agent > invoke_agent my_agent
```

That metadata is on the graph's own bound config, which is the one place inheritance can't lie, so announce it as an agent here rather than reclassifying in `classify_chain_run`:

```python
    react_name = _react_agent_name(graph)
    if react_name is not None:
        return True, react_name or None
    if isinstance(config, Mapping):
        metadata = cast("Mapping[str, Any]", config).get("metadata")
        if isinstance(metadata, Mapping):
            m = cast("Mapping[str, Any]", metadata)
            if m.get("otel_agent_span") or m.get("agent_type") or m.get("agent_name"):
                name = m.get("agent_name")
                return True, str(name) if name else None
    return False, None
```

That collapses it back to a single `invoke_agent my_agent`, needs no change in `classify_chain_run`, and the existing tests still pass. Please add the case above as a test - agent metadata on the subgraph itself, not on the outer invoke.

instrumentation/opentelemetry-instrumentation-genai-langchain/src/opentelemetry/instrumentation/genai/langchain/callback_handler.py
```diff
@@ -103,25 +103,41 @@ def on_chain_start(
...
+            )
```
lmolkova: Both fallbacks here are dead. langchain_core always passes `name=config.get("run_name") or self.get_name()`, and for a `Pregel` `get_name()` is the graph name - the same value `_workflow_name` reads - so `kwargs["name"]` is always set and always equal to `declared_workflow_name`. `serialized.get("name")` is never reached either, which is the only reason it doesn't blow up: with current langgraph `serialized` is `None` on every chain callback.

```suggestion
            workflow_name = kwargs.get("name")
```

If that's right, `_workflow_name` and `_PendingGraph.name` for the workflow path can go too - the announcement only needs to carry `is_agent`. If there is a case where they differ, it needs a test.

instrumentation/opentelemetry-instrumentation-genai-langchain/tests/test_workflow.py
lmolkova: In async the nested span is a second root span, not a child - that's #513, not a bug in this change. Please make it visible rather than skipping the assert:

```python
    if async_mode:
        pytest.xfail("nested spans are not parented in async - #513")
    assert inner_span.parent.span_id == outer_span.context.span_id
```

instrumentation/opentelemetry-instrumentation-genai-langchain/src/opentelemetry/instrumentation/genai/langchain/agent_context.py
lmolkova, line 11: nit: the module docstring and the body of `claim_graph`'s docstring explained why the announcement exists at all - callback metadata can't identify a nested graph root, only the innermost announcement is claimable, and only once. None of that is inferable from the code; please keep it, updated for graphs.

instrumentation/opentelemetry-instrumentation-genai-langchain/src/opentelemetry/instrumentation/genai/langchain/callback_handler.py
1fanwang: Done in 66b7e04.

instrumentation/opentelemetry-instrumentation-genai-langchain/src/opentelemetry/instrumentation/genai/langchain/operation_mapping.py
1fanwang: Done in 66b7e04.

instrumentation/opentelemetry-instrumentation-genai-langchain/tests/test_workflow.py
1fanwang: Done in 66b7e04.

instrumentation/opentelemetry-instrumentation-genai-langchain/src/opentelemetry/instrumentation/genai/langchain/agent_context.py
1fanwang: Done in 66b7e04.

instrumentation/opentelemetry-instrumentation-genai-langchain/src/opentelemetry/instrumentation/genai/langchain/callback_handler.py
1fanwang: Done in 66b7e04.

instrumentation/opentelemetry-instrumentation-genai-langchain/tests/test_workflow.py
1fanwang: Done in 66b7e04.

instrumentation/opentelemetry-instrumentation-genai-langchain/src/opentelemetry/instrumentation/genai/langchain/agent_context.py
1fanwang: Done in 66b7e04.

instrumentation/opentelemetry-instrumentation-genai-langchain/src/opentelemetry/instrumentation/genai/langchain/agent_context.py
```diff
@@ -77,35 +70,55 @@ def _react_agent_name(graph: Any) -> str | None:
...
     if react_name is not None:
         return True, react_name or None
+    if (
+        metadata.get("otel_agent_span")
+        or metadata.get("agent_type")
+        or metadata.get("agent_name")
+    ):
+        name = metadata.get("agent_name")
+        return True, str(name) if name else None
     return False, None
```
lmolkova, line 104: Two items here:
- When `is_agent` is `False`, the function returns `(False, None)`. If the compiled graph has a name (`getattr(graph, "name", None)`), it should be returned here so the graph announcement carries the workflow name.
- `_bound_metadata` only checks `instance.config`. It should also inspect `kwargs.get("config")` to capture metadata passed dynamically at invocation time.

instrumentation/opentelemetry-instrumentation-genai-langchain/src/opentelemetry/instrumentation/genai/langchain/callback_handler.py
```diff
@@ -191,30 +193,62 @@ def on_chain_start(
...
+            announced_agent=announced_agent,
+            ancestor_agent_names=ancestor_agent_names,
+            announced_workflow=announced_workflow,
         )
         conversation_id = _conversation_id(metadata)
         capture_content = self._telemetry_handler.should_capture_content()
         if operation == OperationName.INVOKE_WORKFLOW:
-            workflow_name = kwargs.get("name") or serialized.get("name")
+            workflow_name = kwargs.get("name")
             workflow_name_override = (
```
lmolkova, line 248: Please restore `serialized.get("name")` as a fallback (`kwargs.get("name") or serialized.get("name")`). Dropping it causes `workflow_name` to resolve to `None` when `kwargs` does not contain `name`.

instrumentation/opentelemetry-instrumentation-genai-langchain/src/opentelemetry/instrumentation/genai/langchain/operation_mapping.py
```diff
@@ -87,6 +99,33 @@ def create_agent_graph_name(config: Any) -> str | None:
...
+
+def without_inherited_operation_metadata(
+    metadata: dict[str, Any] | None,
+    inherited_operation_metadata: tuple[Mapping[str, Any], ...] | None,
+) -> dict[str, Any] | None:
+    if not metadata or not inherited_operation_metadata:
+        return metadata
+    filtered_metadata = dict(metadata)
+    for inherited_metadata in inherited_operation_metadata:
+        for key, value in inherited_metadata.items():
```
lmolkova, line 123: Stripping keys based on value equality (`filtered_metadata[key] == value`) drops legitimate child metadata when parent and child share values (e.g. identical `workflow_name` or `agent_type`). Rather than stripping metadata across ancestor runs, let `announced_workflow` take precedence directly in `classify_chain_run`.

instrumentation/opentelemetry-instrumentation-genai-langchain/src/opentelemetry/instrumentation/genai/langchain/invocation_manager.py
```diff
@@ -14,6 +15,7 @@ class _InvocationState:
     invocation: GenAIInvocation | None
     children: list[UUID] = field(default_factory=lambda: list())
     parent_run_id: UUID | None = None
+    graph_metadata: tuple[dict[str, Any], ...] | None = None
```
lmolkova, line 18: Plumbing `graph_metadata` through `_InvocationState` adds unnecessary state tracking to `_InvocationManager`. If announcement precedence is handled in `classify_chain_run`, this extra state is not needed.

instrumentation/opentelemetry-instrumentation-genai-langchain/tests/test_callback_handler.py
```diff
@@ -162,15 +162,16 @@ def test_workflow_span_created(self):
             handler._invocation_manager.get_invocation(run_id) is workflow_inv
         )
 
-    def test_workflow_name_from_serialized(self):
+    def test_workflow_name_from_callback(self):
```
lmolkova, line 165: Please retain coverage for `serialized.get("name")` rather than replacing it with `kwargs["name"]`. Both resolution paths should be covered.

instrumentation/opentelemetry-instrumentation-genai-langchain/tests/test_workflow.py
```diff
@@ -0,0 +1,345 @@
+# Copyright The OpenTelemetry Authors
```
lmolkova, line 1: I put together a test suite on branch `langchain-nested-workflow-tests` covering nested topologies (2-level, 3-level, siblings), naming overrides, workflow/agent interactions, non-graph runnable suppression, and sync error propagation:
https://github.com/lmolkova/opentelemetry-python-genai/commit/283c65f1ab5f16c7880aa65d5ec6f668351ce780

Could you bring these tests into your branch and work on making them pass?

lmolkova (changes requested):

PR #622, opened 2026-09-04
[google-genai] Do not unconditionally set OTEL_INSTRUMENTATION_GENAI_EMIT_EVENT during instrumentation

### LINKED ISSUE #619: google-genai: instrument_generate_content() unconditionally sets OTEL_INSTRUMENTATION_GENAI_EMIT_EVENT=true, overriding config and user setting
### Package

`opentelemetry-instrumentation-google-genai` 1.0b0 (with `opentelemetry-util-genai` 1.0b0)

### What happens

[`instrument_generate_content()`](https://github.com/open-telemetry/opentelemetry-python-genai/blob/main/instrumentation/opentelemetry-instrumentation-google-genai/src/opentelemetry/instrumentation/google_genai/generate_content.py) unconditionally mutates a global process environment variable when instrumenting:

```python
def instrument_generate_content(
    generate_content_config_key_allowlist: AllowList,
) -> object:
    os.environ["OTEL_INSTRUMENTATION_GENAI_EMIT_EVENT"] = "true"
    ...
```

This has a few problems:

1. **It overrides the documented default behavior.** `opentelemetry.util.genai.utils.should_emit_event()` is documented to default based on `OTEL_INSTRUMENTATION_GENAI_CAPTURE_MESSAGE_CONTENT`:
   - `NO_CONTENT` or `SPAN_ONLY` → events default **off**
   - `EVENT_ONLY` or `SPAN_AND_EVENT` → events default **on**

   Because instrumenting force-sets `OTEL_INSTRUMENTATION_GENAI_EMIT_EVENT=true` (which `should_emit_event()` treats as highest priority), the `gen_ai.client.inference.operation.details` event is emitted for **every** request even in `NO_CONTENT` / `SPAN_ONLY` mode. In those modes the event carries no message content, so it just duplicates the span's attributes — redundant telemetry the user did not ask for.

2. **It overrides an explicit user setting.** A user who sets `OTEL_INSTRUMENTATION_GENAI_EMIT_EVENT=false` to disable the event has their choice silently overwritten by `instrument()`.

3. **It mutates global process state as a side effect of instrumenting.** Writing to `os.environ` affects the whole process (and any other GenAI instrumentation reading the same variable), is order-dependent, and is not thread-safe.

### Minimal reproduction

```python
import os

os.environ["OTEL_INSTRUMENTATION_GENAI_CAPTURE_MESSAGE_CONTENT"] = "SPAN_ONLY"
os.environ["OTEL_INSTRUMENTATION_GENAI_EMIT_EVENT"] = "false"  # explicitly disable the event

from opentelemetry.instrumentation.google_genai import GoogleGenAiSdkInstrumentor

GoogleGenAiSdkInstrumentor().instrument()

print(os.environ["OTEL_INSTRUMENTATION_GENAI_EMIT_EVENT"])
# -> "true"   (the user's "false" was overwritten)
```

After this, a `generate_content` call emits a `gen_ai.client.inference.operation.details` event even though the user asked for `SPAN_ONLY` capture and explicitly set `EMIT_EVENT=false`.

### Expected

`instrument()` should not write to `os.environ`. Whether the event is emitted should be decided by `should_emit_event()` from the (unmodified) configuration:
- default off for `NO_CONTENT` / `SPAN_ONLY`,
- default on for `EVENT_ONLY` / `SPAN_AND_EVENT`,
- and an explicit `OTEL_INSTRUMENTATION_GENAI_EMIT_EVENT` from the user always respected.

If there is a reason the event must be emitted regardless of capture mode, it would be clearer to pass that intent into the `TelemetryHandler` / invocation directly rather than mutating a global environment variable.

### Context

Found while integrating this instrumentation into [Pydantic Logfire](https://github.com/pydantic/logfire). As a downstream workaround we snapshot `OTEL_INSTRUMENTATION_GENAI_EMIT_EVENT` before calling `instrument()` and restore it afterwards, but that only works because `should_emit_event()` reads the variable per request rather than at instrument time.

somuai:
# Description

Fixes #619

In `instrumentation/opentelemetry-instrumentation-google-genai/src/opentelemetry/instrumentation/google_genai/generate_content.py`, `instrument_generate_content()` unconditionally mutated the global environment via:
```python
os.environ["OTEL_INSTRUMENTATION_GENAI_EMIT_EVENT"] = "true"
```

This overrides explicit user settings (such as `OTEL_INSTRUMENTATION_GENAI_EMIT_EVENT="false"`) as well as the default resolution logic from `OTEL_INSTRUMENTATION_GENAI_CAPTURE_MESSAGE_CONTENT` implemented in `opentelemetry.util.genai.utils.should_emit_event()`, while polluting `os.environ` for any other code running in the process.

This PR removes the unconditional assignment so `should_emit_event()` properly evaluates user configuration and capturing modes.

## Type of change

# How has this been tested?

- Added `test_instrument_does_not_mutate_emit_event_env` and `test_instrument_preserves_explicit_emit_event_env` in `instrumentation/opentelemetry-instrumentation-google-genai/tests/test_instrumentor.py`.
- Verified test suite passes: `pytest instrumentation/opentelemetry-instrumentation-google-genai/tests/test_instrumentor.py`.
- Formatted and linted via `ruff check` and `ruff format`.

# Checklist

- [x] Unit tests added
- [ ] Documentation updated

### DIFF
--- a/instrumentation/opentelemetry-instrumentation-google-genai/src/opentelemetry/instrumentation/google_genai/generate_content.py
+++ b/instrumentation/opentelemetry-instrumentation-google-genai/src/opentelemetry/instrumentation/google_genai/generate_content.py
@@ -842,7 +842,6 @@ def instrument_generate_content(
     telemetry_handler: TelemetryHandler,
     generate_content_config_key_allowlist: AllowList,
 ) -> object:
-    os.environ["OTEL_INSTRUMENTATION_GENAI_EMIT_EVENT"] = "true"
     snapshot = _MethodsSnapshot()
     wrapped = wrap_function_wrapper(
         "google.genai.models",
--- a/instrumentation/opentelemetry-instrumentation-google-genai/.changelog/619.fixed
+++ b/instrumentation/opentelemetry-instrumentation-google-genai/.changelog/619.fixed
@@ -1,0 +1,1 @@
+Do not unconditionally set OTEL_INSTRUMENTATION_GENAI_EMIT_EVENT during instrumentation.
--- a/instrumentation/opentelemetry-instrumentation-google-genai/tests/test_instrumentor.py
+++ b/instrumentation/opentelemetry-instrumentation-google-genai/tests/test_instrumentor.py
@@ -3,6 +3,8 @@
 
 """Tests for GoogleGenAiSdkInstrumentor."""
 
+import os
+
 from google.genai.models import AsyncModels, Models
 
 from opentelemetry.instrumentation.google_genai import (
@@ -14,6 +16,9 @@ from opentelemetry.instrumentation.google_genai.interactions import (
     InteractionsResource,
 )
 from opentelemetry.test_util_genai.instrumentor import instrument
+from opentelemetry.util.genai.environment_variables import (
+    OTEL_INSTRUMENTATION_GENAI_EMIT_EVENT,
+)
 
 
 def test_co_filename_on_wrapped_functions(
@@ -58,3 +63,29 @@ def test_co_filename_on_wrapped_functions(
         ), (
             f"Expected opentelemetry/instrumentation/google_genai removed from {co_filename} upon uninstrument"
         )
+
+
+def test_instrument_does_not_mutate_emit_event_env(
+    monkeypatch, tracer_provider, logger_provider, meter_provider
+):
+    monkeypatch.delenv(OTEL_INSTRUMENTATION_GENAI_EMIT_EVENT, raising=False)
+    with instrument(
+        GoogleGenAiSdkInstrumentor(),
+        tracer_provider=tracer_provider,
+        logger_provider=logger_provider,
+        meter_provider=meter_provider,
+    ):
+        assert OTEL_INSTRUMENTATION_GENAI_EMIT_EVENT not in os.environ
+
+
+def test_instrument_preserves_explicit_emit_event_env(
+    monkeypatch, tracer_provider, logger_provider, meter_provider
+):
+    monkeypatch.setenv(OTEL_INSTRUMENTATION_GENAI_EMIT_EVENT, "false")
+    with instrument(
+        GoogleGenAiSdkInstrumentor(),
+        tracer_provider=tracer_provider,
+        logger_provider=logger_provider,
+        meter_provider=meter_provider,
+    ):
+        assert os.environ.get(OTEL_INSTRUMENTATION_GENAI_EMIT_EVENT) == "false"


instrumentation/opentelemetry-instrumentation-google-genai/src/opentelemetry/instrumentation/google_genai/generate_content.py
```diff
@@ -842,7 +842,6 @@ def instrument_generate_content(
     telemetry_handler: TelemetryHandler,
     generate_content_config_key_allowlist: AllowList,
 ) -> object:
-    os.environ["OTEL_INSTRUMENTATION_GENAI_EMIT_EVENT"] = "true"
```
DylanRussell, line 845: I think this was added because we want to emit the event regardless of the CONTENT_CAPUTRE mode ( we want CONTENT_CAPTURE to just impact whether content is captured on events/spans) where as I think other instrumentations were using the content_capture mode / env var to determine whether an event is emitted at all -- can you check if that is still true or not ?

somuai: Hi @DylanRussell, thank you for the feedback. I investigated the entire codebase across all instrumentations and `opentelemetry-util-genai`. Here is what I verified:

1. **How other instrumentations operate**:
All instrumentations using `InferenceInvocation` and `TelemetryHandler` (`openai`, `bedrock`, `langchain`, `smolagents`, `portkey`, and `google-genai`) delegate event emission entirely to `InferenceInvocation._maybe_create_event()`, which calls `should_emit_event()`. None of the other instrumentations set `os.environ["OTEL_INSTRUMENTATION_GENAI_EMIT_EVENT"]`.

2. **The contract in `opentelemetry-util-genai/src/opentelemetry/util/genai/utils.py`**:
`should_emit_event()` explicitly implements:
- If `OTEL_INSTRUMENTATION_GENAI_EMIT_EVENT` is set, user preference takes highest priority (`"true"` -> True, `"false"` -> False).
- If `OTEL_INSTRUMENTATION_GENAI_EMIT_EVENT` is not set, it defaults based on `OTEL_INSTRUMENTATION_GENAI_CAPTURE_MESSAGE_CONTENT`:
  - `NO_CONTENT` or `SPAN_ONLY`: defaults to `False`.
  - `EVENT_ONLY` or `SPAN_AND_EVENT`: defaults to `True`.

3. **Content capture vs event emission**:
In `InferenceInvocation._maybe_create_event()`, message content serialization is decoupled via `get_content_attributes()`:
- When `OTEL_INSTRUMENTATION_GENAI_EMIT_EVENT=true` is set, `should_emit_event()` returns `True`, so the `gen_ai.client.inference.operation.details` event is emitted.
- If `CONTENT_CAPTURE` is `NO_CONTENT`, `get_content_attributes()` omits `input_messages`, `output_messages`, and `system_instruction` from the event payload, while preserving top-level operation attributes.
- Thus, users can already emit events without content by setting `OTEL_INSTRUMENTATION_GENAI_EMIT_EVENT=true` and `OTEL_INSTRUMENTATION_GENAI_CAPTURE_MESSAGE_CONTENT=NO_CONTENT`.

4. **Why hardcoding `os.environ["OTEL_INSTRUMENTATION_GENAI_EMIT_EVENT"] = "true"` was problematic**:
- In `generate_content.py`, mutating `os.environ` directly at instrumentation time overrides any explicit user setting (such as `OTEL_INSTRUMENTATION_GENAI_EMIT_EVENT=false`) and mutates the global process environment for all instrumentations.
- Removing the hardcoded assignment brings `google-genai` into exact alignment with all other instrumentations and restores respect for user-configured environment variables.

Let me know if you would prefer `google-genai` to retain a specific default when no environment variables are set, or if aligning with the shared `should_emit_event()` contract across all instrumentations is the desired direction.

DylanRussell: yeah that will be a breaking change for users of this instrumentation.. i don't like that the CAPTURE_CONTENT env var is responsible for event emission.. that is not intuitive.. i'd prefer we got rid of that logic instead...

somuai: Hi @DylanRussell,

Thank you for the review and for calling this out. You have raised two critical points:

1. Semantic Separation of Concerns & Public Documentation:
You are completely right that OTEL_INSTRUMENTATION_GENAI_CAPTURE_MESSAGE_CONTENT should not be governing event emission. To answer your question from the issue: this behavior is NOT documented or required anywhere in the public OpenTelemetry Semantic Conventions specification. The official GenAI semconv defines payload privacy (capturing prompt/completion content) and event emission (gen_ai.client.inference.operation.details) as orthogonal concepts. An event can and should record operational telemetry (model name, token counts, response status) even when content capture is off. The fallback was introduced internally in Python PR #3994 (opentelemetry-util-genai) as a local heuristic.

2. Backwards Compatibility for google-genai:
Removing the hardcoded os.environ assignment exposes that utility fallback, causing google-genai to stop emitting events by default for unconfigured users, which is indeed a breaking change for existing users of this package. At the same time, the original issue (#619) arose because unconditionally writing os.environ["OTEL_INSTRUMENTATION_GENAI_EMIT_EVENT"] = "true" clobbers explicit user opt-outs (e.g. when an operator or framework explicitly sets OTEL_INSTRUMENTATION_GENAI_EMIT_EVENT=false) and leaks across the entire process.

To resolve this cleanly, there are two paths:

Path A (Scoped, non-breaking fix for google-genai in this PR):
We preserve default event emission for google-genai when OTEL_INSTRUMENTATION_GENAI_EMIT_EVENT is unset (for example, using os.environ.setdefault("OTEL_INSTRUMENTATION_GENAI_EMIT_EVENT", "true") or resolving defaults in-memory), so existing users continue getting events without any disruption, while strictly honoring explicit user opt-outs like OTEL_INSTRUMENTATION_GENAI_EMIT_EVENT="false".

Path B (Decoupling in opentelemetry-util-genai):
As you suggested, we remove the CAPTURE_CONTENT fallback logic from should_emit_event() in opentelemetry-util-genai. If we do that, what should the global default for should_emit_event() be when OTEL_INSTRUMENTATION_GENAI_EMIT_EVENT is unset? Should it default to True (events on by default across all instrumentations, with message content gated by CAPTURE_CONTENT), or False?

Please let me know which direction you would prefer, and I will be happy to update this PR or submit a PR for opentelemetry-util-genai.

DylanRussell: My preference is to default it to `True` and remove the capture content fallback logic.. Someone else should weigh in on this though

somuai: Thanks @DylanRussell, that makes total sense. Defaulting event emission to `True` aligns well with standard telemetry expectations while keeping content payload capture decoupled under `CAPTURE_CONTENT`. Let's see if other maintainers / semconv WG members want to chime in on the global default. In the meantime, I can prepare a PR against `opentelemetry-util-genai` implementing that decoupling if you'd like, or wait for consensus here.

somuai:
Implemented Step 1 as discussed with @aabmass and @DylanRussell in #619.

Summary of changes:
1. In `opentelemetry-util-genai`:
   - Introduced in-memory process default (`set_emit_event_default()`, `get_emit_event_default()`) and context-local override (`set_emit_event_context()`, `reset_emit_event_context()`).
   - `should_emit_event()` respects explicit `OTEL_INSTRUMENTATION_GENAI_EMIT_EVENT` first, then context override, then process-level in-memory default, before falling back to `OTEL_INSTRUMENTATION_GENAI_CAPTURE_MESSAGE_CONTENT`.
2. In `opentelemetry-instrumentation-google-genai`:
   - `instrument_generate_content()` sets `set_emit_event_default(True)` without touching `os.environ`.
   - `_MethodsSnapshot.restore()` restores the prior default.
   - Explicit user settings like `OTEL_INSTRUMENTATION_GENAI_EMIT_EVENT="false"` are strictly honored without mutating global environment state.

All unit tests in both packages pass and ruff formatting/linting has been verified cleanly.

DylanRussell:
This feels pretty heavyweight.. i'm hoping we can get the "long term" solution in before the next release, which will make this obsolete.. Sorry but I dont think we should do this.. If you want to work on the "long term" fix which should be implementable when otel python is next released which should be any day now that would be great

somuai:
Understood @DylanRussell, that makes complete sense. Keeping the codebase clean without adding temporary fallback machinery is definitely the right call, especially with the next core release landing shortly.

I would be glad to work on the long-term fix. I will keep an eye out for the upcoming `opentelemetry-python` core release and implement the decoupled semantic conventions logging approach once it is out. Closing this interim PR for now.

PR #625, opened 2026-09-04
[`opentelemetry-instrumentation-genai-openai`] Record gen_ai.tool.definitions on Responses API spans

### LINKED ISSUE #626: [`opentelemetry-instrumentation-genai-openai`] Responses API never records gen_ai.tool.definitions
### Describe your environment

`opentelemetry-instrumentation-genai-openai` 1.2b0 (`main`), `openai` 2.51.0,
`OTEL_SEMCONV_STABILITY_OPT_IN=gen_ai_latest_experimental`,
`OTEL_INSTRUMENTATION_GENAI_CAPTURE_MESSAGE_CONTENT=SPAN_AND_EVENT`.

### What happened?

`gen_ai.tool.definitions` is never set on chat spans produced by the Responses
API, even when the request carries a `tools` array. The Chat Completions path
records it correctly, so the same tool definitions show up on one API surface and
are missing on the other.

This affects every entrypoint that records request attributes: sync and async
`responses.create()`, and sync and async `responses.stream()`.

### Steps to Reproduce

```python
client.responses.create(
    model="gpt-4.1",
    input="What's the weather in Seattle?",
    tools=[
        {
            "type": "function",
            "name": "get_current_weather",
            "description": "Get the current weather in a city.",
            "parameters": {
                "type": "object",
                "properties": {"city": {"type": "string"}},
            },
        }
    ],
)
```
Then inspect the exported span.

### Expected Result

The span carries gen_ai.tool.definitions holding the request's tool
definitions, as the Chat Completions path does.

### Actual Result

The attribute is absent.

### Additional context

extract_params() in response_extractors.py takes an explicit keyword for each
parameter it records. tools is not among them, so it lands in **_kwargs and is
discarded before apply_request_attributes() runs. Tool definitions are only read
off a fetched response, which covers responses.retrieve alone.

Yes

### Tip

sfc-gh-zeningchen:
# Description

`extract_params` did not accept `tools`, so the parameter was dropped before
`apply_request_attributes` could record it. Tool definitions were only ever read
off a fetched response, which covers `responses.retrieve` alone — every `create`
and `stream` call omitted `gen_ai.tool.definitions`, while the Chat Completions
path recorded it correctly.

This threads `tools` through the request params and shares the tool-definition
mapping between the request and response paths, covering sync and async `create`
and `stream()`. A second commit adds the Responses tool calling conformance
scenario that was missing, so Weaver validates the attribute on this path.

Fixes #626

## Type of change

# How has this been tested?

Unit tests cover parameter extraction, the flat Responses tool mapping (function
and built-in tools) and content-capture gating; span-level assertions cover all
four entrypoints. Each new assertion was confirmed to fail without the fix.

The new conformance scenario runs a two-turn Responses tool loop against a freshly
recorded cassette. The streaming tests reuse the existing streaming cassettes —
VCR does not match on the request body and this attribute is read off the request,
so the recorded response is irrelevant there.

- [x] `openai` suite: 254 passed (latest), 109 passed (oldest)
- [x] `openai` conformance: 9 passed

# Checklist

- [x] Unit tests added
- [ ] Documentation updated


instrumentation/opentelemetry-instrumentation-genai-openai/src/opentelemetry/instrumentation/genai/openai/response_extractors.py
```diff
@@ -95,6 +95,7 @@ class ResponseRequestParams:
     service_tier: str | None = None
     temperature: float | None = None
     output_type: str | None = None
+    tools: Sequence[object] | None = None
```
eternalcuriouslearner: question(blocking): Can we use a type here instead of object?

instrumentation/opentelemetry-instrumentation-genai-openai/src/opentelemetry/instrumentation/genai/openai/response_extractors.py
```diff
@@ -295,10 +298,10 @@ def _finish_reason_from_status(
     return None
 
 
-def get_tool_definitions_from_response(
-    response: Response | None,
+def get_tool_definitions(
```
eternalcuriouslearner, line 317: question(non-blocking):  Could this accept reusable non-Sequence iterables while explicitly skipping one-shot Iterators? The OpenAI SDK accepts `Iterable[ToolParam]` the current narrowing omits valid iterable collections from telemetry, while consuming a generator would exhaust SDK input.

eternalcuriouslearner (approved):

instrumentation/opentelemetry-instrumentation-genai-openai/src/opentelemetry/instrumentation/genai/openai/response_extractors.py
sfc-gh-zeningchen: Sure, thanks for catching this!

PR #631, opened 2026-09-05
[`opentelemetry-instrumentation-genai-openai`] Record `gen_ai.conversation.id` on Responses spans

sfc-gh-zeningchen:
# Description

The Responses request extractor never read the `conversation` parameter, so
`gen_ai.conversation.id` was absent from every chat span. Semconv marks the
attribute [conditionally required](https://github.com/open-telemetry/semantic-conventions-genai/blob/94f432d7126f5884d30a2cdde6f4e89908ebb6fd/docs/gen-ai/gen-ai-spans.md#inference)
on the inference span when the instrumented library has a conversation id
readily available, which the Responses API does via
[`conversation`](https://platform.openai.com/docs/api-reference/responses/create#responses_create-conversation);
its registry definition is
[here](https://github.com/open-telemetry/semantic-conventions-genai/blob/94f432d7126f5884d30a2cdde6f4e89908ebb6fd/model/gen-ai/registry.yaml#L412).

`extract_params` now reads `conversation` and `apply_request_attributes` sets
`invocation.conversation_id`, accepting both shapes the SDK declares: a bare id
string, and an object carrying the id under `id`.

`responses.retrieve` is deliberately unchanged — `gen_ai.conversation.id` is not
part of the fetch response span. `previous_response_id` is also left alone: it
identifies a response, not a conversation, and semconv forbids surrogate values.

## Type of change

# How has this been tested?

New tests cover sync and async `responses.create`, streamed `create`,
`responses.stream()`, the API-error path, both accepted value shapes, and the
absence of the attribute when `conversation` is not passed, plus unit cases for
the SDK's unset sentinels.

- [x] `openai` suite: 269 passed (latest), 109 passed (oldest floor, `openai==1.26.0`)
- [x] `openai` conformance: 8 scenarios passed

# Checklist

- [x] Unit tests added
- [ ] Documentation updated

### DIFF
--- a/instrumentation/opentelemetry-instrumentation-genai-openai/tests/test_responses.py
+++ b/instrumentation/opentelemetry-instrumentation-genai-openai/tests/test_responses.py
@@ -58,10 +58,12 @@ try:
     _create_params = set(inspect.signature(_Responses.create).parameters)
     _has_tools_param = "tools" in _create_params
     _has_reasoning_param = "reasoning" in _create_params
+    _has_conversation_param = "conversation" in _create_params
 except ImportError:
     HAS_RESPONSES_API = False
     _has_tools_param = False
     _has_reasoning_param = False
+    _has_conversation_param = False
 
 
 pytestmark = pytest.mark.skipif(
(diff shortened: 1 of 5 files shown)


instrumentation/opentelemetry-instrumentation-genai-openai/tests/test_responses.py
```diff
@@ -589,6 +591,130 @@ def test_responses_create_with_all_params(
...
+
+@requires_conversation_param
+@pytest.mark.parametrize(
+    "conversation",
+    [CONVERSATION_ID, {"id": CONVERSATION_ID}],
+    ids=["id", "object"],
+)
+@pytest.mark.cassette("test_responses_create_basic[content_mode0]")
+@pytest.mark.vcr()
+def test_responses_create_records_conversation_id(
```
lmolkova: nit: most of these 10 new tests don't add coverage. `conversation_id` is set once in `apply_request_attributes`, before the streaming/non-streaming and success/error paths diverge - so the streaming-create and error tests (x2 files) assert the same line as the basic one. The `{"id": ...}` shape is already covered by the extractor unit test, so the parametrize is redundant too.

Request-param coverage in this file already lives in `test_responses_create_with_all_params` + `_assert_request_attrs`. Suggest following that: add `conversation` there (and to the async twin), keep the `responses.stream()` test since it's a separate `extract_params` call site, and drop the rest.

lmolkova (approved):

instrumentation/opentelemetry-instrumentation-genai-openai/tests/test_responses.py
sfc-gh-zeningchen: Thanks! Updated as suggested. One other change: since `conversation` only landed in openai 2.0 and this package supports `openai >= 1.26.0`, the parameter is now only passed and asserted when the installed SDK has it.

PR #632, opened 2026-09-05
Emit telemetry under the instrumentation's own scope

### LINKED ISSUE #620: util-genai: telemetry is emitted under scope `opentelemetry.util.genai.handler` instead of the instrumentation library, losing attribution
### Package

`opentelemetry-util-genai` 1.0b0 — affects every instrumentation built on `TelemetryHandler` (e.g. `opentelemetry-instrumentation-google-genai` 1.0b0).

### What happens

[`TelemetryHandler.__init__`](https://github.com/open-telemetry/opentelemetry-python-genai/blob/main/util/opentelemetry-util-genai/src/opentelemetry/util/genai/handler.py) creates its tracer, meter, and logger using the util module's own `__name__`:

```python
class TelemetryHandler:
    def __init__(self, tracer_provider=None, meter_provider=None, logger_provider=None, completion_hook=None):
        schema_url = Schemas.V1_37_0.value
        self._tracer = get_tracer(__name__, __version__, tracer_provider, schema_url=schema_url)
        meter = get_meter(__name__, meter_provider=meter_provider, schema_url=schema_url)
        self._logger = get_logger(__name__, __version__, logger_provider, schema_url=schema_url)
```

So all spans, metrics, and log events end up under the instrumentation scope **`opentelemetry.util.genai.handler`**, regardless of which instrumentation library actually produced them.

### Why this is a problem

- **The instrumentation scope no longer identifies the instrumentation.** In `opentelemetry-instrumentation-google-genai` 0.x, spans were emitted under `opentelemetry.instrumentation.google_genai`. In 1.0b0 (built on `util-genai`) they're emitted under `opentelemetry.util.genai.handler`. This is a regression in attributability.
- **Every `util-genai`-based instrumentation shares one scope.** As more instrumentations migrate onto `TelemetryHandler`, google-genai, openai, anthropic, etc. all become indistinguishable by `otel.scope.name` — the standard OpenTelemetry mechanism for attributing telemetry to the library that produced it.
- **Downstream tooling that keys off the instrumentation scope breaks** — dashboards, per-library routing/filtering, sampling rules, enable/disable, and cost attribution can no longer tell these instrumentations apart.
- **There is no supported way to override it.** `TelemetryHandler` accepts only providers, and each instrumentor constructs its own handler internally (e.g. `GoogleGenAiSdkInstrumentor._instrument` builds `TelemetryHandler(...)`), so a consumer can't inject the correct instrumentation scope.

### Expected

Telemetry should carry the instrumentation scope of the library that produced it (e.g. `opentelemetry.instrumentation.google_genai`), consistent with how OpenTelemetry instrumentations normally call `get_tracer(__name__, __version__, ...)` with their own module name and with the semantic-convention expectation that the scope identifies the instrumentation.

A straightforward fix would be for `TelemetryHandler` to accept an instrumenting module name and version (and schema URL), which each instrumentation passes in when it constructs the handler — mirroring what `BaseInstrumentor` already threads through to `get_tracer`.

### Context

Found while integrating `opentelemetry-instrumentation-google-genai` 1.0b0 into [Pydantic Logfire](https://github.com/pydantic/logfire). As a downstream workaround we rewrite the span's instrumentation scope in a span processor, but that only fixes **spans** (not metrics or logs) and depends on the internal `opentelemetry.util.genai.handler` name. (Separately filed #619 about the same integration force-setting an env var.)

lmolkova:
Fixes #620.

`TelemetryHandler` now takes `instrumentation_scope_name`/`instrumentation_scope_version`, and every instrumentation passes its own, so spans, metrics and logs carry the instrumentation's scope instead of `opentelemetry.util.genai.handler`. 

`get_telemetry_handler` is deprecated  - no need to have shared state.

Known gaps:

- Proper validation for instrumentation scope will be coming through conformance scenarios - open-telemetry/semantic-conventions-conformance#179.


util/opentelemetry-util-genai/.changelog/632.deprecated
```diff
@@ -0,0 +1 @@
+Deprecate get_telemetry_handler, construct a TelemetryHandler with instrumentation_scope_name instead
```
lmolkova:
```suggestion
Deprecate get_telemetry_handler, construct and cache  TelemetryHandler instead.
```

.github/instructions/instrumentation.instructions.md
```diff
@@ -37,6 +37,10 @@ prefer opt-in or additive. Breaking changes need explicit justification in the P
 
 - Spans, logs, metrics, and events must go through `opentelemetry-util-genai`. Direct use of
   `Tracer`, `Meter`, `Logger`, or event APIs is not allowed.
+- `TelemetryHandler` must be constructed with `instrumentation_scope_name=__name__` and
```
xrmx, line 40: I'd add the possibility that sometime `__name__` is not what you want, e.g. google-genai instrumentation

AGENTS.md
```diff
@@ -191,6 +191,12 @@ Apply to packages under `instrumentation/`.
 - Spans, logs, metrics, and events should go through `opentelemetry-util-genai`. Do not call OTel
   `Tracer`/`Meter`/`Logger` directly, and import only its public surface — never an
   `opentelemetry.util.genai._*` module.
+- Construct the `TelemetryHandler` with `instrumentation_scope_name=__name__` and
```
xrmx, line 194: Same here, I'd say the name should be the dotted path of the instrumentation module.

xrmx (approved):
Thanks!

eternalcuriouslearner (approved):
LGTM!! Got a dumb question!! Can __name__ be null/none in any case?

.github/instructions/instrumentation.instructions.md
lmolkova: thanks, addressed

lmolkova:
> LGTM!! Got a dumb question!! Can **name** be null/none in any case?

Thanks! Not in normal circumstances, but technically possible to set it to None. Added tests to validate proper names.

PR #641, opened 2026-09-06
[opentelemetry-instrumentation-genai-anthropic] Record cache token usage via the shared invocation fields

Krishnachaitanyakc:
# Description

`messages_extractors` defined its own `gen_ai.usage.cache_creation.input_tokens` and `gen_ai.usage.cache_read.input_tokens` string constants and wrote them straight into `invocation.attributes`, so the attribute names were pinned in this package rather than tracking the semantic conventions. `InferenceInvocation` already carries `cache_creation_input_tokens` and `cache_read_input_tokens` fields for exactly this, and openai uses them (`response_extractors.py:661`).

This sets those fields instead and drops the two local constants. The emitted attributes are unchanged.

## Type of change

# How has this been tested?

The existing cache token tests now assert against the semconv constants rather than the removed local ones.

- [x] `anthropic` suite: 178 passed (latest), 171 passed with 7 skipped (oldest)
- [x] `anthropic` conformance: 5 passed
- [x] `tox -e precommit` and `tox -e typecheck` clean

# Checklist

- [x] Unit tests updated

### DIFF
--- a/instrumentation/opentelemetry-instrumentation-genai-anthropic/src/opentelemetry/instrumentation/genai/anthropic/messages_extractors.py
+++ b/instrumentation/opentelemetry-instrumentation-genai-anthropic/src/opentelemetry/instrumentation/genai/anthropic/messages_extractors.py
@@ -65,12 +65,6 @@ class MessageRequestParams:
     system: str | Iterable[TextBlockParam] | None = None
 
 
-GEN_AI_USAGE_CACHE_CREATION_INPUT_TOKENS = (
-    "gen_ai.usage.cache_creation.input_tokens"
-)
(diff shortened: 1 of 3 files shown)


instrumentation/opentelemetry-instrumentation-genai-anthropic/.changelog/641.fixed
```diff
@@ -0,0 +1 @@
+record cache token usage through the shared invocation fields instead of locally defined attribute name constants
```
lmolkova: no need for the changelog - the is no use facing change here

lmolkova (approved):

PR #648, opened 2026-09-08
fix(google-genai): respect event emission configuration

### LINKED ISSUE #619: google-genai: instrument_generate_content() unconditionally sets OTEL_INSTRUMENTATION_GENAI_EMIT_EVENT=true, overriding config and user setting

walleseve:
## Why

Google GenAI instrumentation unconditionally set `OTEL_INSTRUMENTATION_GENAI_EMIT_EVENT=true` during instrumentation. That overwrote an explicit `false`, changed the shared default, and mutated process-global configuration. Fixes #619.

## What

Remove the Google-specific environment mutation. Add regression coverage for preserving unset and explicit-false configuration, default no-event behavior when content capture is disabled, and explicit event opt-in where tests require events.

## Why the approach

`opentelemetry-util-genai` already owns event-selection policy through `should_emit_event()`. Deferring to that shared path avoids duplicate configuration logic, preserves the public interface, and limits production code to a one-line removal.

## Validation

- Google GenAI latest: 212 passed, 118 skipped
- Google GenAI oldest: 163 passed, 167 skipped
- Google GenAI conformance: 4 passed
- `tox -e typecheck`: 0 errors
- `tox -e precommit`: passed
- `tox -e changelog-preview`: passed

### DIFF
--- a/instrumentation/opentelemetry-instrumentation-google-genai/.changelog/648.fixed
+++ b/instrumentation/opentelemetry-instrumentation-google-genai/.changelog/648.fixed
@@ -1,0 +1,1 @@
+Respect the `OTEL_INSTRUMENTATION_GENAI_EMIT_EVENT` setting in Google GenAI instrumentation.
--- a/instrumentation/opentelemetry-instrumentation-google-genai/src/opentelemetry/instrumentation/google_genai/generate_content.py
+++ b/instrumentation/opentelemetry-instrumentation-google-genai/src/opentelemetry/instrumentation/google_genai/generate_content.py
@@ -842,7 +842,6 @@ def instrument_generate_content(
     telemetry_handler: TelemetryHandler,
     generate_content_config_key_allowlist: AllowList,
 ) -> object:
-    os.environ["OTEL_INSTRUMENTATION_GENAI_EMIT_EVENT"] = "true"
     snapshot = _MethodsSnapshot()
     wrapped = wrap_function_wrapper(
         "google.genai.models",
--- a/instrumentation/opentelemetry-instrumentation-google-genai/tests/generate_content/nonstreaming_base.py
+++ b/instrumentation/opentelemetry-instrumentation-google-genai/tests/generate_content/nonstreaming_base.py
@@ -359,6 +359,7 @@ class NonStreamingTestCase(TestCase):
         {
             "OTEL_GOOGLE_GENAI_GENERATE_CONTENT_CONFIG_INCLUDES": "gcp.gen_ai.operation.config.response_schema",
             "OTEL_INSTRUMENTATION_GENAI_CAPTURE_MESSAGE_CONTENT": "NO_CONTENT",
+            "OTEL_INSTRUMENTATION_GENAI_EMIT_EVENT": "true",
         },
     )
     def test_log_event_no_content_capture(self):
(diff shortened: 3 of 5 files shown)


DylanRussell:
Duplicate of https://github.com/open-telemetry/opentelemetry-python-genai/pull/622 -- i don't think we need this PR

PR #650, opened 2026-09-08
[`opentelemetry-instrumentation-genai-openai`] Record Responses API tool-loop history

sfc-gh-zeningchen:
# Description

Responses API input arrives as a flat item list, so `function_call` and `function_call_output` items carry no `role` and `get_input_messages` skipped them. A turn following a tool call recorded only the original user message, and no `tool_call_response` part ever reached `gen_ai.input.messages`. The extractor now dispatches on the item `type` before falling back to the role-based path: a tool call becomes an assistant message with a `tool_call` part, its result a `tool` message with a `tool_call_response` part, and `call_id` correlates the two. Consecutive tool-call items are one turn the flat list split up, so they merge back into a single assistant message, matching the Chat Completions shape. Custom tools take the same path on the response side as well as the request side; their `input` holds free-form text rather than a JSON arguments string.

`_tool_response_to_data` reduces a tool result to data the attributes can carry. It sits next to `_content_to_parts` in `utils.py` because both are content helpers this package shares, which also lets the Chat Completions path use it. That path could raise the same `TypeError` when a tool message held an SDK model, though only a caller who put one there would reach it.

The `openai-v1` test environment covers the 1.x range between the Responses API arriving in `1.66.0` and `custom_tool_call` arriving in `1.99.2`, which no environment reached before. It skips 10 tests whose features postdate `1.99.1`, 7 of them pre-existing — including three `tool_search` cases from #700, which the new environment surfaced.

## Type of change

# How has this been tested?

Unit tests in `test_response_extractors.py` cover the dispatch: plain dicts and replayed SDK models, structured tool output the span serializer cannot encode, parallel calls, custom tools, a missing `call_id`, unparsable arguments, and older util versions that lack the part types. Five cassette-backed tests replay `api.openai.com` recordings of both the function-tool and custom-tool loops, sync and async. `test_utils.py` covers the Chat Completions tool path. Every new test fails against the pre-change code.

- [x] `openai` suite: 347 passed at latest, 351 on `v2`, 334 on the new `v1`, 153 at the declared floor
- [x] `py314-test-instrumentation-genai-openai-conformance`: 12 passed
- [x] `tox -e precommit` (ruff, ruff-format), `spellcheck` and `typecheck` clean

# Checklist

- [x] Unit tests added
- [ ] Documentation updated


instrumentation/opentelemetry-instrumentation-genai-openai/src/opentelemetry/instrumentation/genai/openai/response_extractors.py
```diff
@@ -374,6 +460,37 @@ def get_tool_definitions_from_response(
...
 
 
+def _is_tool_call_item(item: object) -> bool:
+    """Whether a response output item is a client-side tool call."""
+    types = tuple(
+        model
+        for model in (ResponseFunctionToolCall, ResponseCustomToolCall)
+        if model is not None
+    )
+    return bool(types) and isinstance(item, types)
```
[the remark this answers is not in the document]
sfc-gh-zeningchen: good point, just updated this.

instrumentation/opentelemetry-instrumentation-genai-openai/src/opentelemetry/instrumentation/genai/openai/response_extractors.py
```diff
@@ -207,6 +214,119 @@ def get_system_instruction(instructions: str | None) -> list[TextPart]:
...
+    # The span serializer base64s `bytes` but has no encoder for `bytearray`.
+    if isinstance(value, bytearray):
+        return bytes(value)
+    model_dump = getattr(value, "model_dump", None)
+    if callable(model_dump):
+        return _json_safe(model_dump())
+    if isinstance(value, Mapping):
+        return {str(key): _json_safe(item) for key, item in value.items()}
+    if isinstance(value, Sequence):
+        return [_json_safe(item) for item in value]
```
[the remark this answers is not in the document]
sfc-gh-zeningchen: Reworked this rather than adding a branch for `set`. `_json_safe` receives three fields. `output` is declared on two item types:

| field | declared type |
|---|---|
| `function_call.arguments` | `Required[str]` |
| `custom_tool_call.input` | `Required[str]` |
| `function_call_output.output` | `Required[Union[str, ResponseFunctionCallOutputItemListParam]]` |
| `custom_tool_call_output.output` | `Required[Union[str, Iterable[OutputOutputContentList]]]` |

The list variant holds `ResponseInputTextContentParam`, `ResponseInputImageContentParam` or `ResponseInputFileContentParam`, and every field on those three is a `str`, a `Literal`, `None`, or a nested `PromptCacheBreakpoint`. Note `input_file.file_data` is `Optional[str]`, a base64 string, so even binary payloads arrive as text.

So the reachable value space is: a string, a null, a list, a dict, or an SDK model that a caller replayed from `response.output`. `_json_safe` now handles the five reachable shapes and returns `None` for everything else, instead of falling through to `str(value)`.

`str(value)` can raise when `__str__` does, and handling every possible type here is not feasible. Those values are not valid input anyway: they fall outside the shape the SDK declares.

instrumentation/opentelemetry-instrumentation-genai-openai/src/opentelemetry/instrumentation/genai/openai/response_extractors.py
```diff
@@ -39,6 +39,9 @@
 
 try:
     from openai.types.responses.response import Response
+    from openai.types.responses.response_custom_tool_call import (
+        ResponseCustomToolCall,
+    )
```
lmolkova, line 44:
```suggestion
```
Import `ResponseCustomToolCall` in an isolated `try...except ImportError` block.

Responses API and `ResponseFunctionToolCall` were added in `openai 1.66.0`, while `ResponseCustomToolCall` was added in `openai 1.99.2`. On versions `1.66.0` through `1.99.1`, bundling this import raises `ImportError` and sets `Response` and `ResponseFunctionToolCall` to `None`, completely disabling Responses API instrumentation.

Consider also adding a test environment to the tox matrix for `openai 1.x` (e.g. `openai==1.66.0`) like we already do for `v2`.

instrumentation/opentelemetry-instrumentation-genai-openai/src/opentelemetry/instrumentation/genai/openai/response_extractors.py
```diff
@@ -222,6 +228,102 @@ def get_system_instruction(instructions: str | None) -> list[TextPart]:
...
+    if value is None or isinstance(value, str):
+        return value
+    model_dump = getattr(value, "model_dump", None)
+    if callable(model_dump):
+        return _json_safe(model_dump())
+    if isinstance(value, Mapping):
+        return {str(key): _json_safe(item) for key, item in value.items()}
+    if isinstance(value, Sequence):
+        return [_json_safe(item) for item in value]
+    return None
```
lmolkova, line 263:
```suggestion
    if value is None or isinstance(value, (str, int, float, bool)):
        return value
    model_dump = getattr(value, "model_dump", None)
    if callable(model_dump):
        return _json_safe(model_dump())
    if isinstance(value, Mapping):
        return {str(key): _json_safe(item) for key, item in value.items()}
    if isinstance(value, Sequence) and not isinstance(value, (bytes, bytearray)):
        return [_json_safe(item) for item in value]
    return None
```
Why do we need `_json_safe` in instrumentation instead of handling SDK models in `opentelemetry-util-genai` (e.g. supporting `model_dump()` in `_GenAiJsonEncoder`)?

If kept here: the current logic converts `int`, `float`, and `bool` to `None`, and iterates `bytes` as a sequence.

sfc-gh-zeningchen: that's a good point! I moved it to `openai/utils.py` next to `_content_to_parts`, and import it here instead, also applied your change to the function. I kept it out of util-genai because it's specifically for translating openai SDK's payload shape. Thanks for catching that!

instrumentation/opentelemetry-instrumentation-genai-openai/src/opentelemetry/instrumentation/genai/openai/response_extractors.py
sfc-gh-zeningchen: thanks! updated this part and also added a test env for `openai 1.x`

instrumentation/opentelemetry-instrumentation-genai-openai/src/opentelemetry/instrumentation/genai/openai/response_extractors.py
```diff
@@ -228,6 +242,90 @@ def get_system_instruction(instructions: str | None) -> list[TextPart]:
...
+def _get_input_message(item: object) -> InputMessage | None:
+    """Convert one input item; a tool-call turn is flat, not a message."""
+    if InputMessage is None or Role is None:
+        return None
+
+    item_type = _get_field(item, "type")
+
+    if item_type in _TOOL_CALL_ARGUMENT_FIELDS and ToolCall is not None:
+        raw = _get_field(item, _TOOL_CALL_ARGUMENT_FIELDS[item_type])
+        return InputMessage(
```
lmolkova, line 286: In the Responses API ([API reference](https://platform.openai.com/docs/api-reference/responses/create#responses-create-input)), `input` is a flat list of items where each item has a `type`:
- `message`: standard user/system message with a `role`.
- `function_call` / `custom_tool_call`: tool call requested by the assistant.
- `function_call_output` / `custom_tool_call_output`: tool execution result.

When mapping this flat list to `gen_ai.input.messages`:
1. `message` items map 1:1 to `InputMessage`.
2. Parallel tool calls are generated by the assistant in a single turn, but arrive as consecutive `function_call` / `custom_tool_call` items. These consecutive items should be merged into a single `InputMessage(role="assistant", parts=[...])`.
3. Tool outputs (`function_call_output` / `custom_tool_call_output`) should remain individual `InputMessage(role="tool", parts=[...])` messages, one per result.

instrumentation/opentelemetry-instrumentation-genai-openai/src/opentelemetry/instrumentation/genai/openai/utils.py
```diff
@@ -234,6 +234,33 @@ def _document_to_part(file_obj: Any) -> MessagePart | None:
     return None
 
 
+def _tool_response_to_data(value: Any) -> Any:
```
lmolkova: Can we be more precise than Any for input param? 

For result it should be AnyValue

Should we move this to genai-util serialization logic?

instrumentation/opentelemetry-instrumentation-genai-openai/src/opentelemetry/instrumentation/genai/openai/utils.py
```diff
@@ -234,6 +234,33 @@ def _document_to_part(file_obj: Any) -> MessagePart | None:
...
+    """Reduce a tool result payload to data the GenAI attributes can carry.
+
+    A caller hands back whatever its tool produced, which may include SDK
+    models. Those are flattened via ``model_dump()``; anything that is neither
```
lmolkova: why do we need to model_dump here? we should pass complex object to tool result and let util put it on attributes (serialized if necessary).

instrumentation/opentelemetry-instrumentation-genai-openai/src/opentelemetry/instrumentation/genai/openai/response_extractors.py
sfc-gh-zeningchen: thanks for pointing it out! I also fixed the same issue for `gen_ai.output.messages`

lmolkova: Thanks for working on this!

One detail regarding `gen_ai.output.messages`: in the Responses API ([API reference](https://platform.openai.com/docs/api-reference/responses/object)), a single model response returns a flat `output` list containing all generated items (reasoning, message/text, and tool calls). For example, when the model generates assistant text and multiple tool calls:

```json
{
  "output": [
    {
      "type": "message",
      "role": "assistant",
      "content": [{"type": "output_text", "text": "I will check the weather for both cities."}]
    },
    {
      "type": "function_call",
      "call_id": "call_1",
      "name": "get_weather",
      "arguments": "{\"location\": \"Seattle, WA\"}"
    },
    {
      "type": "function_call",
      "call_id": "call_2",
      "name": "get_weather",
      "arguments": "{\"location\": \"Boston, MA\"}"
    }
  ]
}
```

In OpenTelemetry GenAI semconv, `gen_ai.output.messages` represents candidate/choice completions generated by the model. Since all these items are produced in a single generation, they should be mapped as parts of a single assistant message:

```json
[
  {
    "role": "assistant",
    "parts": [
      {
        "type": "text",
        "content": "I will check the weather for both cities."
      },
      {
        "type": "tool_call",
        "id": "call_1",
        "name": "get_weather",
        "arguments": {"location": "Seattle, WA"}
      },
      {
        "type": "tool_call",
        "id": "call_2",
        "name": "get_weather",
        "arguments": {"location": "Boston, MA"}
      }
    ],
    "finish_reason": "tool_call"
  }
]
```

Currently, the extractor merges consecutive tool calls, but leaves preceding text as a separate `OutputMessage`. Could we merge all output items for the turn into one `OutputMessage(role="assistant", parts=[...])`?

lmolkova: Thanks for working on this!

One detail regarding `gen_ai.output.messages`: in the Responses API ([API reference](https://platform.openai.com/docs/api-reference/responses/object)), a single model response returns a flat `output` list containing all generated items (reasoning, message/text, and tool calls). For example, when the model generates assistant text and multiple tool calls:

```json
{
  "output": [
    {
      "type": "message",
      "role": "assistant",
      "content": [{"type": "output_text", "text": "I will check the weather for both cities."}]
    },
    {
      "type": "function_call",
      "call_id": "call_1",
      "name": "get_weather",
      "arguments": "{\"location\": \"Seattle, WA\"}"
    },
    {
      "type": "function_call",
      "call_id": "call_2",
      "name": "get_weather",
      "arguments": "{\"location\": \"Boston, MA\"}"
    }
  ]
}
```

In OpenTelemetry GenAI semconv, `gen_ai.output.messages` represents candidate/choice completions generated by the model. Since all these items are produced in a single generation, they should be mapped as parts of a single assistant message:

```json
[
  {
    "role": "assistant",
    "parts": [
      {
        "type": "text",
        "content": "I will check the weather for both cities."
      },
      {
        "type": "tool_call",
        "id": "call_1",
        "name": "get_weather",
        "arguments": {"location": "Seattle, WA"}
      },
      {
        "type": "tool_call",
        "id": "call_2",
        "name": "get_weather",
        "arguments": {"location": "Boston, MA"}
      }
    ],
    "finish_reason": "tool_call"
  }
]
```

So what we did for putput messages was right (well aligned with what OpenAI thinks about it).

I see how it's asymmetrical with input messages and that we're forcing out input messages model onto openai Responses API where tool call requests are not messages per se.

In chat completions, each tool call request was individual message with unique tool call id, so the choice here is to stay consistent across openai flavors for input messages.

sfc-gh-zeningchen: thanks for the explanation! just wanted to clarify that with the original implementation the same scenario produced three separate messages, so I think that part still needed changes. I've updated it to merge them into one message.

lmolkova (approved):

PR #652, opened 2026-09-08
[`opentelemetry-instrumentation-genai-anthropic`] Record gen_ai.tool.definitions

sfc-gh-zeningchen:
# Description

The Anthropic instrumentation never recorded `gen_ai.tool.definitions`, so a
tool-calling request produced a span with no trace of the tools it offered the
model. `extract_params` reads the request arguments one by one and had no slot
for `tools`, so the list never reached the code that records telemetry.

A custom tool carries `input_schema` and maps to a `FunctionToolDefinition`.
Server tools (`web_search_20250305`, `bash_20250124`) and toolsets
(`computer_toolset_20260801`) map to a `GenericToolDefinition` keyed by their
versioned `type`; a toolset carries no `name`, so its `type` stands in for the
`name` the schema requires.

The instrumentation records the attribute only when
`OTEL_INSTRUMENTATION_GENAI_CAPTURE_MESSAGE_CONTENT` is set, matching
`genai-openai`. `spans.yaml` marks the attribute `opt_in` and places it in the
`attributes.gen_ai.content` group, so following the content setting is the
narrower reading.

Known gaps: `genai-langchain`, `genai-agno`, `genai-portkey` and
`genai-smolagents` set `tool_definitions` regardless of that setting, so one
call yields a different attribute depending on the provider. Separately,
`get_content_attributes` in util-genai serializes `description` and
`parameters` on its capture-disabled path, against the registry note that
non-required properties stay out by default. Both belong in util-genai and are
tracked in #655; once that lands, this gate can move or drop. Tools reached
through `mcp_servers` stay uncovered, since they never appear in the `tools`
array.

## Type of change

# How has this been tested?

Unit tests cover `get_tool_definitions` across the custom, server-tool and
unnamed-toolset shapes, tools that expose attributes instead of dict keys, and
the empty and `None` cases. Sync and async tests assert the serialized
attribute on `create` and `stream` spans with content capture on, and its
absence with capture off alongside the input and output messages. Both pairs
replay existing cassettes, since the attribute comes from the request and the
recorded response does not affect it.

- [x] `anthropic` oldest leg (`anthropic==0.51.0`): 184 passed, 7 skipped
- [x] `anthropic` latest leg (`anthropic==1.4.0`): 191 passed
- [x] `anthropic` conformance: 5 scenarios pass weaver live-check

# Checklist

- [x] Unit tests added
- [ ] Documentation updated


instrumentation/opentelemetry-instrumentation-genai-anthropic/src/opentelemetry/instrumentation/genai/anthropic/patch.py
```diff
@@ -249,6 +250,9 @@ def _create_invocation(
     invocation.system_instruction = (
         get_system_instruction(params.system) if capture_content else []
     )
+    invocation.tool_definitions = (
+        get_tool_definitions(params.tools) if capture_content else None
+    )
     invocation.attributes = attributes
```
lmolkova, line 271: The `gen_ai.tool.definitions` attribute is recommended (not opt-in). Per the [semconv schema](https://github.com/open-telemetry/semantic-conventions-genai/blob/main/model/gen-ai/gen-ai-tool-definitions.json), only the large optional properties (`description` and `parameters`) should be excluded by default, while `name` and `type` are required.

Please populate `invocation.tool_definitions` unconditionally with the required properties (`name`, `type`), while keeping the content-capture guard around `description` and `parameters` in the instrumentation for performance reasons.

sfc-gh-zeningchen: It seems like `gen_ai.tool.definitions` is still opt-in at this moment ([semconv](https://github.com/open-telemetry/semantic-conventions-genai/blob/main/docs/gen-ai/gen-ai-spans.md#L110)), I'm wondering if we're moving this to recommended in the near future.
I think one inconsistency here is that `gen_ai.tool.definitions` is `opt-in` but `description` and `parameters` have a nested level of `opt-in` based on that. But `ContentCapturingMode` isn't granular enough to distinguish this `opt-in` and `opt-in's opt-in`.
Maybe in this PR, I can proceed with only emitting `name` and `type` when content capture is off, then open an issue to discuss more granularity for `ContentCapturingMode`? Or we can put this PR on hold until we have a solution

instrumentation/opentelemetry-instrumentation-genai-anthropic/src/opentelemetry/instrumentation/genai/anthropic/patch.py
```diff
@@ -249,6 +250,9 @@ def _create_invocation(
     invocation.system_instruction = (
         get_system_instruction(params.system) if capture_content else []
     )
+    invocation.tool_definitions = (
+        get_tool_definitions(params.tools) if capture_content else None
+    )
```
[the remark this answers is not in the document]
sfc-gh-zeningchen: good catch, updated this.

instrumentation/opentelemetry-instrumentation-genai-anthropic/src/opentelemetry/instrumentation/genai/anthropic/patch.py
lmolkova: Sorry for the confusion, you're absolutely right. I created an issue to follow up in the conventions - https://github.com/open-telemetry/semantic-conventions-genai/issues/512

I think your PR is correct as is.

lmolkova (approved):
LGTM, please resolve conflicts and we should be good to go.

Thanks!

sfc-gh-zeningchen:
> LGTM, please resolve conflicts and we should be good to go.
> 
> Thanks!

Thanks! I've resolved the conflict

PR #656, opened 2026-09-08
Fix cancelled GenAI wrapper finalization

### LINKED ISSUE #646: Follow-up to #520: except Exception wrappers still leak spans on cancellation
### Describe your environment

OS: any
Python version: 3.10+ (CancelledError is BaseException since 3.8)
Package version: main incl. 59e6efe (#520)
GenAI library (e.g. anthropic, openai) and version: openai (any); same pattern in portkey / qwen-agent / smolagents / anthropic

### What happened?

#520 fixed the context-manager path (`GenAIInvocation.__exit__` now guards on `BaseException`), but as its own description notes, packages that finalize in a hand-rolled `except Exception:` handler are "worse off than unfixed": on `asyncio.CancelledError` neither `stop()` nor `fail()` runs, so the span is never ended, nothing is exported at all, and the context token leaks.

Sites verified on main, all `except Exception as error: invocation.fail(error); raise`:
- `instrumentation/opentelemetry-instrumentation-genai-openai/.../openai/patch.py:85,117,132,151` (chat sync/async + embeddings sync/async)
- `openai/patch_responses.py:111,184,289,359`
- `portkey/patch.py:114,147`
- `qwen_agent/patch.py:111`
- `smolagents/patch.py:396`
- `anthropic/patch.py:141,197`

The Google path is unaffected (uses `with handler.inference(...)`, covered by #520).

I verified this by executing the real `TelemetryHandler` + `InferenceInvocation` from this repo (SDK tracer, `InMemorySpanExporter`, pinned `api/sdk ~=1.43`, `semconv >=0.64b0`): driving the exact `openai/patch.py` try/except shape with a coroutine cancelled mid-await exports 0 spans and leaves the invocation's context token attached. Same harness with `except BaseException` exports 1 span with ERROR status. Full output in Steps to Reproduce.

Secondary, same pattern, same suspected outcome (traced, not executed): `AsyncStreamWrapper.__anext__` (`util/.../genai/stream.py:343-351`, sync twin `:184-192`) catches `except Exception` around the chunk read, so a task cancelled while awaiting the next chunk would skip `_finalize_failure` when iterating without `async with` (the `__aexit__` guard only helps context-managed iteration).

### Steps to Reproduce

```python
import asyncio
# Harness: real TelemetryHandler + InferenceInvocation from this repo,
# SDK TracerProvider + InMemorySpanExporter. Only the network call is faked.
# Replicates the EXACT try/except shape of openai/patch.py::traced_method:
async def openai_shaped_call(handler, sdk_coro):
    inv = handler.inference("openai", request_model="gpt-4o-mini")
    try:
        result = await sdk_coro()
        inv.stop()
        return result
    except Exception as error:  # <-- patch.py:85,117,132,151
        inv.fail(error)
        raise

async def cancelled_mid_await():
    task = asyncio.ensure_future(asyncio.sleep(60))
    task.cancel()
    await task  # raises CancelledError here
```

### Execution output (venv: api/sdk ~=1.43, semconv >=0.64b0):
```python
== 1. control: with-block path (#520) + cancel ==
   finished spans: 1, status: StatusCode.ERROR
== 2. CURRENT openai/patch.py shape + CancelledError ==
   finished spans: 0 (1 would mean fixed)
   context token still attached (leak): True
== 3. FIXED shape (except BaseException) + CancelledError ==
   finished spans: 1, status: StatusCode.ERROR,
   error.type: asyncio.exceptions.CancelledError, propagated: True
```

### Mechanism demo (stdlib only, no key needed):
```python
try:
    raise asyncio.CancelledError()
except Exception:
    print("caught")  # never prints -> fail() never runs
```

### Expected Result

Cancelled calls end spans with ERROR + error.type = "asyncio.exceptions.CancelledError" (fully qualified, per fq_exception_type), context detached, zero leaked spans for non-streaming, streaming, and mid-iteration cancellation alike.

### Actual Result

Non-streaming path: span never ends (0 spans exported, context token leaked). Bare-iteration streaming path: same pattern, same suspected outcome (see secondary note above).

### Additional context

Explicit follow-up invited in #520 ("Happy to follow up separately"); distinct from #630/#633, which cover the LangChain callback path. Suggested fix: widen the listed handlers to `except BaseException` (mirroring #520's merged guard — the GeneratorExit question was already litigated and tested there), fail, and re-raise unchanged; my harness confirms cancellation still propagates. Suggest scoping the PR to OpenAI + util `stream.py` first, fanning out to the other adapters as follow-ups if maintainers prefer it split. The red-before/green-after test strategy from #520 (`test_invocation_base_exception.py` pattern) applies directly, and I can contribute the failing test from this harness.

Yes

### Tip

bhushanasati25:
Fixes #646.

This widens telemetry-finalizing wrapper handlers from `Exception` to `BaseException` so cancellations such as `asyncio.CancelledError` fail and end invocations before being re-raised. It covers shared stream wrappers/managers plus OpenAI, Anthropic, Bedrock, Portkey, Qwen Agent, and Google GenAI hand-written wrapper paths.

Also updates instrumentation reviewer instructions to flag telemetry-finalizing `except Exception` handlers.

Known gaps: none.

### DIFF
--- a/.github/instructions/instrumentation.instructions.md
+++ b/.github/instructions/instrumentation.instructions.md
@@ -93,6 +93,8 @@ Flag, with a link to the rule:
 
 - When catching exceptions from the underlying library to record telemetry, always re-raise the
   original exception unmodified.
+- Flag telemetry-finalizing handlers that catch `Exception` instead of `BaseException`; cancellation
+  raises `asyncio.CancelledError`, which would otherwise leave spans open.
 - Do not raise **new** exceptions in instrumentation/telemetry code.
 
 ## 6. Tests
--- a/.github/instructions/instrumentation.instructions.md
+++ b/.github/instructions/instrumentation.instructions.md
@@ -93,8 +93,7 @@ Flag, with a link to the rule:
 
 - When catching exceptions from the underlying library to record telemetry, always re-raise the
   original exception unmodified.
-- Flag telemetry-finalizing handlers that catch `Exception` instead of `BaseException`; cancellation
-  raises `asyncio.CancelledError`, which would otherwise leave spans open.
+- Flag telemetry-finalizing handlers that catch `Exception` instead of `BaseException` to catch  `asyncio.CancelledError` and such.
 - Do not raise **new** exceptions in instrumentation/telemetry code.
 
 ## 6. Tests
--- a/instrumentation/AGENTS.md
+++ b/instrumentation/AGENTS.md
@@ -65,3 +65,9 @@ Before emitting an inference (`chat`) or `embeddings` span, ask:
 - **Yes** → emit an `invoke_workflow` span around the library API that runs the workflow.
 - **No** → don't emit a workflow span. A single agent run or a plain chain of calls is not an AI
   workflow.
+
+## Exception handling
+
+- When catching exceptions from an underlying library after starting telemetry,
+  catch `BaseException`, not `Exception`, so cancellation and interrupts
+  finalize the invocation before the original exception is re-raised.
--- a/instrumentation/opentelemetry-instrumentation-genai-anthropic/.changelog/656.fixed
+++ b/instrumentation/opentelemetry-instrumentation-genai-anthropic/.changelog/656.fixed
@@ -1,0 +1,1 @@
+Record failed Anthropic invocations when cancellation raises a `BaseException`.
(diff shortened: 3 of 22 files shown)


lmolkova (approved):
thanks for working on this!

it seems there is still `except Exception` in `bedrock` and `google-genai` - could you please update them too?

And please update agents.md in instrumentation folder and copilot review instructions for instrumentations to suggest BaseException and to flag narrow ones.

.github/instructions/instrumentation.instructions.md
```diff
@@ -93,6 +93,8 @@ Flag, with a link to the rule:
 
 - When catching exceptions from the underlying library to record telemetry, always re-raise the
   original exception unmodified.
+- Flag telemetry-finalizing handlers that catch `Exception` instead of `BaseException`; cancellation
+  raises `asyncio.CancelledError`, which would otherwise leave spans open.
```
lmolkova:
```suggestion
- Flag telemetry-finalizing handlers that catch `Exception` instead of `BaseException` to catch  `asyncio.CancelledError` and such.
```

lzchen (approved):

PR #663, opened 2026-09-09
Prototype of span / event / metric deduplication and context passing for inference span

DylanRussell:
# Description

For `InferenceInvocations` the outermost instrumentation puts an empty dict onto the context to suppress duplicate downstream instrumentation from emitting spans / events / metrics.

All inner instrumentations (usually just 1) put all span attributes (except content attributes, which we disable completely from inner instrumentations to avoid extra processing and other issues) onto the context upon completion.

The outer instrumentation reads all those attributes at completion time, and sets them as the default attribute set to put on the span/event, and then sets the attributes it knows about (overwriting the context ones when there's a conflict).

We have a [hardcoded a set](https://github.com/open-telemetry/opentelemetry-python-genai/pull/663/changes#diff-187d16cf07ea9c7f91750e0c713475d1e76603e7822d72d2eb332e2e29303d55R43) of these attributes that should be applied on metrics -- we also look for token count attributes on the context. 

There is a "metric_attributes" property on _invocation.. for now I'm not putting it onto the context, but maybe we want metric attributes to be passed similar to span/event attributes, we just need a way to differentiate them.. We could use a special prefix key on the attributes dict to do this.

## Type of change

# How has this been tested?

Unit tests

# Checklist

- [x] Unit tests added
- [x] Documentation updated


util/opentelemetry-util-genai/src/opentelemetry/util/genai/__init__.py
```diff
@@ -1,2 +1,14 @@
...
+    INFERENCE_ATTRIBUTES_KEY,
+    get_inference_attributes,
+    set_inference_attributes,
+)
+
+__all__ = [
+    "INFERENCE_ATTRIBUTES_KEY",
+    "get_inference_attributes",
+    "set_inference_attributes",
+]
```
lmolkova, line 14:
```suggestion
```

let's not expose context propagation helpers in the public API. Keep the module private (`_context.py`) and do not re-export it in `__init__.py`. All nesting and suppression logic should remain an internal implementation detail of `util-genai`.

util/opentelemetry-util-genai/src/opentelemetry/util/genai/context.py
```diff
@@ -0,0 +1,55 @@
...
+    """Return a Context with the given inference attributes dictionary attached.
+
+    Args:
+        attributes: The mutable inference attributes dictionary.
+        context: The context to attach to. Defaults to the current context.
+
+    Returns:
+        A new Context containing the inference attributes dictionary.
+    """
+    return set_value(_INFERENCE_ATTRIBUTES_KEY, attributes, context=context)
```
lmolkova, line 38: Storing a raw `dict[str, AttributeValue]` in context loses type information, drops structured data (`input_messages`, `output_messages`, `tool_definitions`), and forces manual string-key extraction for metrics (`_METRIC_SEMCONV_KEYS`).

Using a private dataclass to hold shared state avoids clobbering outer identity (like provider or operation name) and allows inner calls to enrich response and token attributes directly:

```suggestion
@dataclass
class _InferenceAttributes:
    provider: str | None = None
    request_model: str | None = None
    response_model_name: str | None = None
    response_id: str | None = None
    server_address: str | None = None
    server_port: int | None = None
    finish_reasons: list[str] | None = None
    input_tokens: int | None = None
    output_tokens: int | None = None
    input_messages: list[InputMessage] = field(default_factory=list)
    output_messages: list[OutputMessage] = field(default_factory=list)
    custom_attributes: dict[str, AttributeValue] = field(default_factory=dict)
```

util/opentelemetry-util-genai/src/opentelemetry/util/genai/_inference_invocation.py
```diff
@@ -193,6 +222,32 @@ def _get_attributes(self) -> dict[str, AttributeValue]:
...
+        return attrs
+
+    def publish_to_context(self) -> None:
+        """Publish the invocation's current attributes to the inference context.
+
+        Instrumentations call this on the request path when they have finished
+        setting request attributes on the invocation, certifying they are done
+        setting attributes and making them visible to downstream invocations.
+        """
+        self._context_attributes.update(self._get_context_attributes())
```
lmolkova, line 244: Instrumentations should not need custom branching or explicit calls to `publish_to_context()`.

If inner and outer invocations share the underlying `_InferenceAttributes` container, property setters update shared state in place automatically:

```suggestion
    # Delegate invocation properties directly to the shared _InferenceAttributes container.
    @property
    def response_model_name(self) -> str | None:
        return self._attrs.response_model_name

    @response_model_name.setter
    def response_model_name(self, value: str | None) -> None:
        self._attrs.response_model_name = value
```

util/opentelemetry-util-genai/src/opentelemetry/util/genai/_inference_invocation.py
```diff
@@ -193,6 +222,32 @@ def _get_attributes(self) -> dict[str, AttributeValue]:
...
+
+    def _finish_already_started(self, error: Error | None = None) -> None:
+        if error is not None:
+            self._apply_error_attributes(error)
+        self._context_attributes.update(self._get_context_attributes())
```
lmolkova, line 249: Marking error attributes on inner finish breaks error isolation. If an outer caller catches the inner error (e.g. for a retry or model fallback), the root span and metrics should not be marked as failed.

Error attributes should only be recorded if the error escapes the outer invocation unhandled:

```suggestion
    def _finish_already_started(self, error: Error | None = None) -> None:
        # Inner failures should not taint outer span or metrics;
        # outer invocation handles error recording if unhandled.
        pass
```

util/opentelemetry-util-genai/src/opentelemetry/util/genai/_invocation.py
```diff
@@ -173,9 +181,15 @@ def _on_stream_chunk(self, chunk_at: float) -> None:
...
         delta = max(chunk_at - last_chunk_at, 0.0)
-        attributes = self._get_metric_attributes()
-        if self._ttfc_seconds is None:
+        is_first_chunk = self._ttfc_seconds is None
+        if is_first_chunk:
             self._ttfc_seconds = delta
+
+        if self.already_started:
+            return
+
```
lmolkova, line 190: Inner stream wrappers still run `_process_chunk` and reconstruct messages, causing duplicate buffering in memory.

Inner wrappers should only handle stream pass-through and chunk timing metrics, skipping chunk accumulation:

```suggestion
        # In stream.py: skip chunk processing on inner layers
        if invocation is None or getattr(invocation, "should_accumulate_stream", True):
            self._process_chunk(chunk)
```

lmolkova (commented):
Great start!

Some ideas I was thinking of:

- we should keep typed properties rather than raw attribute in a dict
- we need to make it flexible and work for other invocations eventually
- it should be easy to read and maintain, so having something like noop/nestedInferenceInvocation would help with it. It would just proxy setters to inner typed attribute representation
- it should be completely invisible to instrumentation - it does not need to know about suppression like it does not care about parent spans.

util/opentelemetry-util-genai/src/opentelemetry/util/genai/__init__.py
DylanRussell: I'm fine with that. done.

util/opentelemetry-util-genai/src/opentelemetry/util/genai/_context.py
```diff
@@ -0,0 +1,55 @@
...
+    """Return a Context with the given inference attributes dictionary attached.
+
+    Args:
+        attributes: The mutable inference attributes dictionary.
+        context: The context to attach to. Defaults to the current context.
+
+    Returns:
+        A new Context containing the inference attributes dictionary.
+    """
+    return set_value(_INFERENCE_ATTRIBUTES_KEY, attributes, context=context)
```
DylanRussell: That makes it harder for instrumentations outside this repo to use this though, they would have to install util genai to get a bunch of these data classes. Also the mapping from class attribute -> sem conv attribute key is not clear this way

> Using a private dataclass to hold shared state avoids clobbering outer identity (like provider or operation name) and allows inner calls to enrich response and token attributes directly:

I don't follow this - provider and operation name can be put into a `dictionary[str, AnyValue]`, how are they clobbered ? Inner calls can modify the dictionary directly.

> forces manual string-key extraction for metrics (_METRIC_SEMCONV_KEYS).

I don't understand what this means, what exactly is bad about _METRIC_SEMCONV_KEYS ?

> drops structured data (input_messages, output_messages, tool_definitions)

It doesn't drop them it just passes the serialized versions. Does the inner instrumentation care about the outer instrumentation's values ? I think not. Does the outer instrumentation ever need to modify the attribute value the inner instrumentation set ? I think probably not, it just needs to take attributes it wasn't able to set and put them into the span.

util/opentelemetry-util-genai/src/opentelemetry/util/genai/_inference_invocation.py
DylanRussell: Ok good point but we can do the same thing with a dictionary.. I updated the code to show how this could be done

util/opentelemetry-util-genai/src/opentelemetry/util/genai/_inference_invocation.py
DylanRussell: Ack that makes sense. fixed

util/opentelemetry-util-genai/src/opentelemetry/util/genai/_invocation.py
DylanRussell: Ack, fixed it

DylanRussell:
> we should keep typed properties rather than raw attribute in a dict

My main concern with this one is it makes it hard to use for instrumentations outside this repo.. I see some benefit to this because we can pass unserialized data structures around the instrumentations, but I don't see a use case for that yet.


> we need to make it flexible and work for other invocations eventually

SGTM

> it should be easy to read and maintain, so having something like noop/nestedInferenceInvocation would help with it. It would just proxy setters to inner typed attribute representation

I don't follow this exactly -- can you clarify ?

> it should be completely invisible to instrumentation - it does not need to know about suppression like it does not care about parent spans.

Mostly SGTM.. It might be interesting for instrumentations to know so they can avoid doing costly work (i'm thinking of the content-capture parsing, but maybe there's other stuff too)

util/opentelemetry-util-genai/src/opentelemetry/util/genai/_inference_invocation.py
```diff
@@ -147,8 +239,42 @@ def __init__(
         # Rebuilt once per streaming chunk, so cache it and invalidate via
         # _invalidate_metric_attributes whenever an input changes.
         self._cached_metric_attributes: dict[str, AttributeValue] | None = None
+
+        existing_attrs = get_inference_attributes()
```
lmolkova: from the discussion we had on Tue call, I think the apprach we discussed was:

- outer just adds empty object on the context
- inner adds attributes
- on finish outer reconciles its attributes with inner

There is an edge case: when there are multiple inners. Both would write to the same object - this is an edge case that I don't think we need to care much about, so they can just overwrite each other.

util/opentelemetry-util-genai/src/opentelemetry/util/genai/_context.py
lmolkova: makes sense, let's park this and see how far we can go with raw attributes

lmolkova:
I did a small prototype to show noop/nestedinferenceinvocation to avoid fragile flag, PTAL:

https://github.com/lmolkova/opentelemetry-python-genai/commit/e0de3d878

util/opentelemetry-util-genai/src/opentelemetry/util/genai/_inference_invocation.py
DylanRussell: Nice that really simplifies things.. the implementation is pretty clean now:

inner only writes attributes once, at the very end, in `_finish_already_started`..
Outer consumes the dictionary only once, at the very end in `_apply_finish `

DylanRussell:
I took a look at your prototype. I like just having `self.already_started` (maybe a better name is needed) be the thing that differentiates inner/outer, I don't like having to have an entire new class.. I did steal a couple things from your prototype:

1. set self.span to get_current_span when `already_started` is true
2. Set content capture to false when `already_started` is true
3. Skip parsing streaming chunks on inner

util/opentelemetry-util-genai/src/opentelemetry/util/genai/_invocation.py
```diff
@@ -141,16 +142,37 @@ def _start(
             attributes: Initial span attributes available for sampling decisions.
             context: An optional OpenTelemetry Context to parent the span.
         """
+        if self.already_started:
```
lmolkova: Using `already_started` to skip `_start()` leaves `self._span_context` uninitialized on `InferenceInvocation`. Calling `invocation.context` or iterating a `SyncStreamWrapper` / `AsyncStreamWrapper` with a real inner `InferenceInvocation` raises `AttributeError: 'InferenceInvocation' object has no attribute '_span_context'`. (`test_stream.py` missed this because it used `MagicMock`.)

Additionally, `_finish_already_started()` in `_inference_invocation.py` calls `get_inference_attributes()` against ambient `get_current()` at finish time instead of the context active when the inner invocation started.

This partially initialized state and branching across methods is why a separate `SuppressedInferenceInvocation` class is cleaner and safer: a dedicated subclass costs nothing, never leaves attributes uninitialized, and does not require every method on `GenAIInvocation` and `InferenceInvocation` to check `already_started`.

Failing test that should pass:
```python
def test_nested_stream_wrapper_with_real_invocation(self) -> None:
    class _ChunkStream(SyncStreamWrapper[dict[str, int]]):
        def _process_chunk(self, chunk: dict[str, int]) -> None:
            self._self_invocation.output_tokens = chunk.get("tokens")

        def _on_stream_end(self) -> None:
            self._self_invocation.stop()

        def _on_stream_error(self, error: BaseException) -> None:
            self._self_invocation.fail(error)

    with self.handler.inference("outer-provider") as outer:
        inner = self.handler.inference("inner-provider", request_model="gpt-4o")
        stream = _ChunkStream(iter([{"tokens": 42}]), invocation=inner)
        # Currently raises AttributeError: 'InferenceInvocation' object has no attribute '_span_context'
        list(stream)

    spans = self.span_exporter.get_finished_spans()
    self.assertEqual(len(spans), 1)
    self.assertEqual(spans[0].attributes.get("gen_ai.usage.output_tokens"), 42)
```

util/opentelemetry-util-genai/src/opentelemetry/util/genai/stream.py
```diff
@@ -192,7 +192,12 @@ def __next__(self) -> ChunkT:
             raise
         invocation = self._self_invocation
         chunk_at = timeit.default_timer() if invocation is not None else None
-        self._process_chunk(chunk)
+        # Inner stream wrappers skip chunk accumulation to avoid duplicate
+        # message buffering in memory.
+        if invocation is None or not getattr(
```
lmolkova, line 197: My earlier suggestion to skip `_process_chunk(chunk)` when `already_started` is `True` has an unintended side effect: stream wrappers extract `input_tokens`, `output_tokens`, `response_model_name`, `response_id`, and `finish_reasons` inside `_process_chunk`. Skipping `_process_chunk` causes inner streaming calls to collect no response metadata or token usage, so nothing is propagated to the outer invocation when the stream ends.

Always call `self._process_chunk(chunk)`. Duplicate message buffering is already avoided because `invocation.should_capture_content` returns `False` for suppressed invocations.

util/opentelemetry-util-genai/src/opentelemetry/util/genai/_invocation.py
```diff
@@ -161,7 +183,7 @@ def _get_metric_token_counts(self) -> dict[str, int]:  # pylint: disable=no-self
 
     def record_stream_chunk(self) -> None:
         """Mark the request as streamed and record one output chunk arriving."""
-        if self._context_token is None:
+        if self.already_started or self._context_token is None:
```
lmolkova: This is another bug caused by scattering `already_started` checks across methods:
1. `record_stream_chunk()` checks `self.already_started` and returns early, so `_request_stream` and `_ttfc_seconds` are not tracked and `gen_ai.request.stream` / `gen_ai.response.time_to_first_chunk` cannot propagate to the outer span.
2. `_on_stream_chunk()` (called directly by `SyncStreamWrapper` / `AsyncStreamWrapper`) missed the `self.already_started` check, so it records `time_to_first_chunk` and `time_per_output_chunk` histogram metrics on inner invocations - causing duplicate streaming metrics (and crashing on missing `_span_context`).

With a separate `SuppressedInferenceInvocation` class, `_on_stream_chunk()` can simply record `_ttfc_seconds` without recording histogram instruments, avoiding fragile flag checks across base class methods.

Failing test that should pass:
```python
def test_nested_streaming_deduplicates_chunk_metrics_and_enriches_ttfc(self) -> None:
    with self.handler.inference("outer") as outer:
        with self.handler.inference("inner") as inner:
            inner.record_stream_chunk()
            inner.record_stream_chunk()
        outer.record_stream_chunk()
        outer.record_stream_chunk()

    metrics = self._harvest_metrics()
    # Should record only 1 time_to_first_chunk point from outer, not 2
    self.assertEqual(len(metrics.get("gen_ai.client.operation.time_to_first_chunk", [])), 1)
    # Inner TTFC and stream flag should propagate if outer did not set them
    spans = self.span_exporter.get_finished_spans()
    self.assertTrue(spans[0].attributes.get("gen_ai.request.stream"))
    self.assertIsNotNone(spans[0].attributes.get("gen_ai.response.time_to_first_chunk"))
```

util/opentelemetry-util-genai/src/opentelemetry/util/genai/_inference_invocation.py
```diff
@@ -375,6 +417,18 @@ def _get_attributes(self) -> dict[str, AttributeValue]:
         attrs.update({k: v for k, v in optional_attrs if v is not None})
         return attrs
 
+    def _finish_already_started(self, error: Error | None = None) -> None:
+        # Error attributes are not recorded on inner finish to isolate errors;
+        # the outer invocation records them only if the error escapes unhandled.
+        existing_attrs = get_inference_attributes()
+        if existing_attrs is not None:
```
lmolkova, line 407: `_finish_already_started()` copies `self.attributes` (span/event-only) into context, but drops `self.metric_attributes`. Low-cardinality custom metric attributes set by inner instrumentations are lost when the outer invocation records metrics.

Failing test that should pass:
```python
def test_nested_custom_metric_attributes_propagated(self) -> None:
    with self.handler.inference("outer"):
        with self.handler.inference("inner") as inner:
            inner.metric_attributes["custom.metric_tag"] = "low-cardinality-val"

    metrics = self._harvest_metrics()
    point = metrics["gen_ai.client.operation.duration"][0]
    self.assertEqual(point.attributes.get("custom.metric_tag"), "low-cardinality-val")
```

DylanRussell: yes i know but what's the solution ? I was thinking we could prefix the metric keys when they go into the context, so we know they should only be applied to metrics.. Or we could just leave out metric keys to begin with

util/opentelemetry-util-genai/src/opentelemetry/util/genai/_invocation.py
DylanRussell: Ack.. i updated the code to set span context to current context on the inner invocations.. i also cleaned up the code to do all the branching logic inside _start more clearly..

DylanRussell: I'm still not sold on the dedicated subclass

util/opentelemetry-util-genai/src/opentelemetry/util/genai/stream.py
DylanRussell: Ack, fixed

util/opentelemetry-util-genai/src/opentelemetry/util/genai/_invocation.py
DylanRussell: Ok i ended up removing that early return, and waiting until we get those attributes.. In our instrumentations I dont think we need those attributes b/c outer overrides inner and will have those attributes.. But for instrumentations outside the repo it could be useful..

DylanRussell:
Made a bunch of changes, thanks for the review.. I still think putting already_started inside the baseclass and adding some branching logic there is not bad, there isn't that much additional logic to parse..

I added some clarifications to the code:
- Change `already_started` to `_already_started` and move it to the base class, the child class doesn't need to know about it. All the child class has to do now is define `_context_attributes_key`, implement `_finish_already_started` and read the context in `_apply_finish` .
-  `_context_token` in the baseclass will be `None` for inner invocations, because it doesnt `attach` a new context, just uses the existing context.
- I added a `finished` bool that we set inside `finish` in the base class so that we dont have to use `context_token` as a proxy for `finished`, just make an explicit class var.
- Clarified that inside `apply_finish` in the child class we filter out `start_attributes` already applied by the outer instrumentation from the `context_attributes` because these have already been set on the span at start time and otherwise we might overwrite these. Also added a call here to `self._invalidate_metric_attributes()` because we may have picked up new metric attributes from the context..

util/opentelemetry-util-genai/src/opentelemetry/util/genai/_inference_invocation.py
lmolkova: We can put an object on the context that contains multiple things such as span+event attributes and metric attributes. Or, we could create a dataclass for all invocation props and put that on the context - suppressed invocations would put things into that (as it already does) and outer would reconcile.

I'm really interested to figure out right design so we could codegen it - ptal at the https://github.com/open-telemetry/opentelemetry-python-genai/pull/702 when you have a moment. Boilerplate and repetitions are not a concern at all for this.

DylanRussell: Alright that way looks pretty good too.. Sent out https://github.com/open-telemetry/opentelemetry-python-genai/pull/765

DylanRussell: To handle metric attributes the dictionary in the context now has this structure:

```
{ 
'spanevent_attributes': {}
'metric_attributes': {}
}
```

DylanRussell:
closing this in favor of https://github.com/open-telemetry/opentelemetry-python-genai/pull/765

PR #662, opened 2026-09-09
[opentelemetry-instrumentation-openai] Capture cached prompt tokens on chat completions

rads-1996:
# Description

Fixes https://github.com/open-telemetry/opentelemetry-python-genai/issues/546

## Type of change

Please delete options that are not relevant.

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update

# How has this been tested?

Please describe the tests that you ran to verify your changes. Provide
instructions so we can reproduce. List any relevant details for your test
configuration.

- [ ] Test A

# Checklist

See [CONTRIBUTING.md](https://github.com/open-telemetry/opentelemetry-python-genai/blob/main/CONTRIBUTING.md)
for the style guide, changelog guidance, and more.

- [x] Followed the style guidelines of this project
- [ ] Changelog updated if the change requires an entry
- [x] Unit tests added
- [ ] Documentation updated

Issue #546, opened 2026-09-02
[openai] Capture cached prompt tokens on chat completions


lmolkova:
> [!NOTE]
> This issue was generated with AI assistance and requires investigation and confirmation prior to being worked on.

Part of #541.

Extract and report cached prompt tokens for `chat.completions.create` and streaming chat responses.

- Extract from `usage.prompt_tokens_details.cached_tokens`.
- Semconv attribute: `gen_ai.usage.cache_read.input_tokens`.
- Affected files: `patch.py` and `chat_wrappers.py`.

rads-1996:
@lmolkova I can work on this. Could I please be assigned this issue.

lmolkova:
@rads-1996 sure! done!

PR #669, opened 2026-09-10
genai-openai: report gen_ai.usage.cache_read.input_tokens on chat completion spans

roy-tong:
## Description

The chat completions path of `opentelemetry-instrumentation-genai-openai` drops
the cached-token count that OpenAI reports in `usage.prompt_tokens_details.cached_tokens`,
so `gen_ai.usage.cache_read.input_tokens` is never emitted for
`chat.completions.create` — neither streaming nor non-streaming — even when the
API response contains it.

The Responses API path in the same package already maps this field
(`response_extractors.py:592-602` reads `input_tokens_details`), so today two
calls that hit the same prompt cache produce inconsistent telemetry depending on
which API surface they used. This also matches the Bedrock gap tracked in
opentelemetry-python-contrib#4614 for a different instrumentation.

Per the GenAI semantic conventions, `gen_ai.usage.cache_read.input_tokens` is
"The number of input tokens served from a provider-managed cache" and its value
SHOULD be included in `gen_ai.usage.input_tokens` — which holds here, since
OpenAI's `prompt_tokens` already includes cached tokens. The fix therefore only
adds the cache breakdown attribute; `gen_ai.usage.input_tokens` is unchanged.

This PR:

- `_set_chat_completion_response_properties` (`patch.py`): also read
  `usage.prompt_tokens_details.cached_tokens` (guarded with `getattr` since
  `prompt_tokens_details` is optional on some SDK versions) and set it on the
  invocation as `cache_read_input_tokens`.
- `_ChatStreamMixin` (`chat_wrappers.py`): capture the same field from the
  final usage chunk and apply it to the invocation at stream cleanup.

Verified against commit ea7fdf6d8b12c3f6b43531c697ad0b84337f3396.

## Testing

Added two VCR-based tests with synthetic cassettes derived from the existing
ones (only `cached_tokens` changed to a non-zero value):

- `test_chat_completion_reports_cached_tokens` (non-streaming, cached_tokens=9)
- `test_chat_completion_streaming_reports_cached_tokens` (streaming with
  `include_usage`, cached_tokens=7)

Before this change both tests fail (`gen_ai.usage.cache_read.input_tokens`
missing from span attributes); after the change both pass:

```
$ pytest tests/test_chat_completions.py -k cached_tokens -q
2 passed, 31 deselected
```

Example span attributes after the fix (streaming):
`gen_ai.usage.input_tokens: 12, gen_ai.usage.output_tokens: 12,
gen_ai.usage.cache_read.input_tokens: 7`
---

<sub>Repro fixtures: [AgentMeasure conformance pack](https://github.com/roy-tong/AgentMeasure/tree/main/conformance/pack) — plain JSONL + expected totals, no runtime install. Found during an [open ecosystem audit](https://github.com/roy-tong/AgentMeasure/blob/main/campaigns/measurement-casebook.md#ecosystem-audit-33-verified-usage-accounting-findings-2026-09-0607) of usage-accounting tools.</sub>

### DIFF
--- a/instrumentation/opentelemetry-instrumentation-genai-openai/.changelog/670.fixed
+++ b/instrumentation/opentelemetry-instrumentation-genai-openai/.changelog/670.fixed
@@ -1,0 +1,4 @@
+Report `gen_ai.usage.cache_read.input_tokens` on chat completion spans from
+`usage.prompt_tokens_details.cached_tokens`, for both streaming and
+non-streaming `chat.completions.create` calls, matching what the Responses
+API instrumentation already emits.
--- a/instrumentation/opentelemetry-instrumentation-genai-openai/src/opentelemetry/instrumentation/genai/openai/chat_wrappers.py
+++ b/instrumentation/opentelemetry-instrumentation-genai-openai/src/opentelemetry/instrumentation/genai/openai/chat_wrappers.py
@@ -40,6 +40,7 @@ class _ChatStreamMixin:
     _self_service_tier: str | None
     _self_prompt_tokens: int | None
     _self_completion_tokens: int | None
+    _self_cache_read_input_tokens: int | None
 
     def _set_response_model(self, chunk: ChatCompletionChunk) -> None:
         # Set eagerly so the per-chunk streaming timing metrics carry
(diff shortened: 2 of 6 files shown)


rads-1996:
@roy-tong An issue for this already exists and I already have a PR up for that - https://github.com/open-telemetry/opentelemetry-python-genai/issues/546. PR - https://github.com/open-telemetry/opentelemetry-python-genai/pull/662

lmolkova:
closing as duplicate. thanks @roy-tong for the contribution!

roy-tong:
Understood - thanks @lmolkova, and thanks @rads-1996 for pointing at #546/#662. Dropping this in favor of #662. If extra coverage helps there, the synthetic VCR cassettes from this branch (cached_tokens set on both streaming and non-streaming paths) are reusable as-is.

PR #672, opened 2026-09-10
Fix Anthropic streaming with anthropic>=1.5.0

lmolkova:
Fixes streaming message accumulation failure with Anthropic SDK >=1.5.0 caused by the added json_bufs parameter on accumulate_event. Also gracefully disables future content accumulation with a warning if accumulate_event fails unexpectedly.

### DIFF
--- a/instrumentation/opentelemetry-instrumentation-genai-anthropic/src/opentelemetry/instrumentation/genai/anthropic/wrappers.py
+++ b/instrumentation/opentelemetry-instrumentation-genai-anthropic/src/opentelemetry/instrumentation/genai/anthropic/wrappers.py
@@ -3,9 +3,12 @@
 
 from __future__ import annotations
 
+import inspect
+import logging
 from collections.abc import Callable
 from typing import (
     TYPE_CHECKING,
+    Any,
     Generic,
     Protocol,
     TypeVar,
@@ -27,6 +30,8 @@ from opentelemetry.util.genai.stream import (
 )
 
 from .messages_extractors import set_invocation_response_attributes
+
+_logger = logging.getLogger(__name__)
 
 try:
     from anthropic.lib.streaming._messages import (  # pylint: disable=no-name-in-module
@@ -57,6 +62,17 @@ if TYPE_CHECKING:
 
 ResponseFormatT = TypeVar("ResponseFormatT")
 accumulate_event = cast("Callable[..., Message] | None", _sdk_accumulate_event)
+
+_accumulate_takes_json_bufs = False
+if accumulate_event is not None:
+    try:
+        _accumulate_takes_json_bufs = (
+            "json_bufs" in inspect.signature(accumulate_event).parameters
+        )
+    except (ValueError, TypeError):
+        _accumulate_takes_json_bufs = False
+
+_accumulation_disabled = False
 
 
 class _StreamWrapperWithStream(Protocol):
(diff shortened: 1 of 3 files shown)


rads-1996 (approved):

instrumentation/opentelemetry-instrumentation-genai-anthropic/src/opentelemetry/instrumentation/genai/anthropic/wrappers.py
```diff
@@ -134,14 +152,29 @@ def _process_chunk(
...
             ),
-        )
+        }
+        if _accumulate_takes_json_bufs:
+            kwargs["json_bufs"] = self._self_json_bufs
+
+        try:
+            self._self_message = accumulate_event(**kwargs)
+        except BaseException:
+            _accumulation_disabled = True
```
eternalcuriouslearner, line 170: question(blocking): Should clear snapshot when accumulation fails midway? Say something like:

```python
self._self_message = None
```

instrumentation/opentelemetry-instrumentation-genai-anthropic/src/opentelemetry/instrumentation/genai/anthropic/wrappers.py
```diff
@@ -126,6 +143,7 @@ def _process_chunk(
         | ParsedMessageStreamEvent[ResponseFormatT],
     ) -> None:
         """Accumulate a final message snapshot from a streaming chunk."""
+        global _accumulation_disabled
```
eternalcuriouslearner, line 146: question(blocking): Is the module-level _accumulation_disabled thread-safe? Since one stream’s failure would disable accumulation for all concurrent and future streams, could this instead be per-wrapper state such as `self._self_accumulation_disabled`?

eternalcuriouslearner (approved):
LGTM!! Couple of questions:

1. I see anthropic failures on few other PRs. Does this resolve those errors?
2. In `util/opentelemetry-util-genai/src/opentelemetry/util/genai/stream.py`, should `SyncStreamWrapper.__next__()` and `AsyncStreamWrapper.__anext__()` include `_process_chunk(chunk)` in their exception handling? Currently, exceptions from the SDK iterator finalize the invocation as failed, but exceptions escaping `_process_chunk()` propagate without finalization. Should these methods `call_finalize_failure(error)` and re-raise in that case?

DylanRussell (approved):

instrumentation/opentelemetry-instrumentation-genai-anthropic/src/opentelemetry/instrumentation/genai/anthropic/wrappers.py
lmolkova: why? in practice it'd fail right away and there will be nothing. if there was something in the past, why don't we record it? yes, it's partial, but that's as much as we were able to get

instrumentation/opentelemetry-instrumentation-genai-anthropic/src/opentelemetry/instrumentation/genai/anthropic/wrappers.py
lmolkova: this is protection from unsupported SDK version, it won't change on the next request.

Assigning and reading a boolean in Python is atomic, so there is no thread-safety issue. The flag is intentionally global as a circuit breaker: if accumulate_event fails due to an unsupported SDK version, we want to stop attempting it process-wide rather than spamming warnings and wasting cycles on every every stream.

PR #673, opened 2026-09-10
[`opentelemetry-instrumentation-genai-agno`] Add instrumentation for `retrieval` and make various other improvements

DylanRussell:
# Description

- Add GenAI `retrieval` instrumentation for Agno's Knowledge.search/Knowledge.asearch.
- Add `_safe_wrap_function` to defer patching submodules via post-import hooks, preventing eager imports of heavy optional dependencies (e.g. FastAPI for workflows).
- Add support for streaming tool responses (tools that return sync or async generators/iterators).
- Added `gen_ai.request.model` and `gen_ai.agent.description` attributes to `AgentInvocation`.

## Type of change

# How has this been tested?

Unit tests and conformance tests

# Checklist

- [x] Unit tests added
- [x] Documentation updated

### DIFF
--- a/instrumentation/opentelemetry-instrumentation-genai-agno/src/opentelemetry/instrumentation/genai/agno/patch.py
+++ b/instrumentation/opentelemetry-instrumentation-genai-agno/src/opentelemetry/instrumentation/genai/agno/patch.py
@@ -81,6 +81,8 @@ _AGNO_KNOWLEDGE_MODULE = "agno.knowledge.knowledge"
 _KNOWLEDGE_CLASS = "Knowledge"
 
 
+# wrapt has no unregister API for post-import hooks; monotonic generations
+# invalidate deferred hooks registered during prior instrumentation cycles.
 _instrumentation_generation: int = 0
 _is_instrumented: bool = False
 
--- a/instrumentation/opentelemetry-instrumentation-genai-agno/src/opentelemetry/instrumentation/genai/agno/patch.py
+++ b/instrumentation/opentelemetry-instrumentation-genai-agno/src/opentelemetry/instrumentation/genai/agno/patch.py
@@ -3,40 +3,59 @@
 
 """Patching functions for Agno instrumentation."""
 
+# pylint: disable=import-outside-toplevel
+
 from __future__ import annotations
 
 import functools
 import logging
-from collections.abc import AsyncIterator, Awaitable, Callable, Iterator
+import sys
+from collections.abc import (
+    AsyncIterator,
+    Awaitable,
+    Callable,
+    Iterator,
+    Sequence,
+)
 from typing import TYPE_CHECKING, Any, cast
 
 if TYPE_CHECKING:
     from agno.agent import RunOutput
+    from agno.knowledge.document.base import Document
+    from agno.knowledge.knowledge import Knowledge
     from agno.run.workflow import WorkflowRunOutput
     from agno.team import TeamRunOutput
+    from agno.tools.function import FunctionCall, FunctionExecutionResult
 
     AgnoRunOutput = RunOutput | TeamRunOutput | WorkflowRunOutput
 
-from wrapt import wrap_function_wrapper
+from wrapt import register_post_import_hook, wrap_function_wrapper
 
 from opentelemetry.instrumentation.genai.agno.stream import (
     AgnoAgentStreamWrapper,
+    AgnoToolStreamWrapper,
     AgnoWorkflowStreamWrapper,
     AsyncAgnoAgentStreamWrapper,
+    AsyncAgnoToolStreamWrapper,
     AsyncAgnoWorkflowStreamWrapper,
 )
 from opentelemetry.instrumentation.genai.agno.utils import (
     _get_property_value,
     format_content,
+    format_retrieval_document,
     prepare_tool_definitions,
 )
 from opentelemetry.instrumentation.utils import unwrap
+from opentelemetry.semconv._incubating.attributes import (
+    gen_ai_attributes as GenAI,
+)
 from opentelemetry.semconv._incubating.attributes.error_attributes import (
     ErrorTypeValues,
 )
 from opentelemetry.util.genai.handler import TelemetryHandler
 from opentelemetry.util.genai.invocation import (
     AgentInvocation,
+    RetrievalInvocation,
     ToolInvocation,
     WorkflowInvocation,
 )
@@ -58,6 +77,34 @@ _AGNO_TOOLS_MODULE = "agno.tools.function"
 _FUNCTION_CALL_CLASS = "FunctionCall"
 _AGNO_WORKFLOW_MODULE = "agno.workflow.workflow"
 _WORKFLOW_CLASS = "Workflow"
+_AGNO_KNOWLEDGE_MODULE = "agno.knowledge.knowledge"
+_KNOWLEDGE_CLASS = "Knowledge"
+
+
+def _safe_wrap_function(
+    target_module: str,
+    target_name: str,
+    wrapper: Callable[..., Any],
+) -> None:
+    """Safely wrap a method if it exists, deferring if module is not yet imported."""
+
+    def _apply(mod: Any) -> None:
+        try:
+            parts = target_name.split(".")
+            curr = mod
+            for part in parts:
+                curr = getattr(curr, part)
+        except AttributeError:
+            # Target class or method may not exist across all supported Agno versions.
+            return
+        wrap_function_wrapper(mod, target_name, wrapper)
+
+    if target_module in sys.modules:
+        _apply(sys.modules[target_module])
+        return
+
+    # Defer wrapping to avoid eagerly importing submodules with heavy or optional dependencies.
+    register_post_import_hook(_apply, target_module)
 
 
 def patch_agent(handler: TelemetryHandler) -> None:
@@ -72,78 +119,92 @@ def patch_agent(handler: TelemetryHandler) -> None:
         f"{_AGENT_CLASS}.arun",
         _agent_arun(handler),
     )
-    try:
-        wrap_function_wrapper(
-            _AGNO_TEAM_MODULE,
-            f"{_TEAM_CLASS}.run",
-            _agent_run(handler),
-        )
-        wrap_function_wrapper(
-            _AGNO_TEAM_MODULE,
-            f"{_TEAM_CLASS}.arun",
-            _agent_arun(handler),
-        )
-    except (ImportError, AttributeError):
-        pass
-    try:
-        wrap_function_wrapper(
-            _AGNO_TOOLS_MODULE,
-            f"{_FUNCTION_CALL_CLASS}.execute",
-            _tool_call_execute(handler),
-        )
-        wrap_function_wrapper(
-            _AGNO_TOOLS_MODULE,
-            f"{_FUNCTION_CALL_CLASS}.aexecute",
-            _tool_call_aexecute(handler),
-        )
-    except (ImportError, AttributeError):
-        pass
-    try:
-        wrap_function_wrapper(
-            _AGNO_WORKFLOW_MODULE,
-            f"{_WORKFLOW_CLASS}.run",
-            _workflow_run(handler),
-        )
-        wrap_function_wrapper(
-            _AGNO_WORKFLOW_MODULE,
-            f"{_WORKFLOW_CLASS}.arun",
-            _workflow_arun(handler),
-        )
-    except (ImportError, AttributeError):
-        pass
+    _safe_wrap_function(
+        _AGNO_TEAM_MODULE,
+        f"{_TEAM_CLASS}.run",
+        _agent_run(handler),
+    )
+    _safe_wrap_function(
+        _AGNO_TEAM_MODULE,
+        f"{_TEAM_CLASS}.arun",
+        _agent_arun(handler),
+    )
+    _safe_wrap_function(
+        _AGNO_TOOLS_MODULE,
+        f"{_FUNCTION_CALL_CLASS}.execute",
+        _tool_call_execute(handler),
+    )
+    _safe_wrap_function(
+        _AGNO_TOOLS_MODULE,
+        f"{_FUNCTION_CALL_CLASS}.aexecute",
+        _tool_call_aexecute(handler),
+    )
+    _safe_wrap_function(
+        _AGNO_WORKFLOW_MODULE,
+        f"{_WORKFLOW_CLASS}.run",
+        _workflow_run(handler),
+    )
+    _safe_wrap_function(
+        _AGNO_WORKFLOW_MODULE,
+        f"{_WORKFLOW_CLASS}.arun",
+        _workflow_arun(handler),
+    )
+    # Knowledge.retrieve and aretrieve delegate to search and asearch, so wrapping
+    # search/asearch avoids duplicate spans.
+    _safe_wrap_function(
+        _AGNO_KNOWLEDGE_MODULE,
+        f"{_KNOWLEDGE_CLASS}.search",
+        _knowledge_search(handler),
+    )
+    _safe_wrap_function(
+        _AGNO_KNOWLEDGE_MODULE,
+        f"{_KNOWLEDGE_CLASS}.asearch",
+        _knowledge_asearch(handler),
+    )
 
 
 def unpatch_agent() -> None:
     """Remove patches from Agno class methods."""
-    try:
-        import agno.agent  # pylint: disable=import-outside-toplevel
+    if _AGNO_MODULE in sys.modules:
+        try:
+            import agno.agent
 
-        unwrap(agno.agent.Agent, "run")
-        unwrap(agno.agent.Agent, "arun")
-    except (ImportError, AttributeError):
-        pass
-    try:
-        import agno.team  # pylint: disable=import-outside-toplevel
+            unwrap(agno.agent.Agent, "run")
+            unwrap(agno.agent.Agent, "arun")
+        except (ImportError, AttributeError):
+            pass
+    if _AGNO_TEAM_MODULE in sys.modules:
+        try:
+            import agno.team
 
-        unwrap(agno.team.Team, "run")
-        unwrap(agno.team.Team, "arun")
-    except (ImportError, AttributeError):
-        pass
-    try:
-        import agno.tools.function  # pylint: disable=import-outside-toplevel
+            unwrap(agno.team.Team, "run")
+            unwrap(agno.team.Team, "arun")
+        except (ImportError, AttributeError):
+            pass
+    if _AGNO_TOOLS_MODULE in sys.modules:
+        try:
+            import agno.tools.function
 
-        unwrap(agno.tools.function.FunctionCall, "execute")
-        unwrap(agno.tools.function.FunctionCall, "aexecute")
-    except (ImportError, AttributeError):
-        pass
-    # Workflow depends on optional packages (like fastapi), may fail to import.
-    try:
-        import agno.workflow.workflow  # pylint: disable=import-outside-toplevel
+            unwrap(agno.tools.function.FunctionCall, "execute")
+            unwrap(agno.tools.function.FunctionCall, "aexecute")
+        except (ImportError, AttributeError):
+            pass
+    if _AGNO_WORKFLOW_MODULE in sys.modules:
+        try:
+            import agno.workflow.workflow
 
-        unwrap(agno.workflow.workflow.Workflow, "run")
-        unwrap(agno.workflow.workflow.Workflow, "arun")
-    except (ImportError, AttributeError):
-        pass
+            unwrap(agno.workflow.workflow.Workflow, "run")
+            unwrap(agno.workflow.workflow.Workflow, "arun")
+        except (ImportError, AttributeError):
+            pass
+    if _AGNO_KNOWLEDGE_MODULE in sys.modules:
+        try:
+            import agno.knowledge.knowledge
+
+            unwrap(agno.knowledge.knowledge.Knowledge, "search")
+            unwrap(agno.knowledge.knowledge.Knowledge, "asearch")
+        except (ImportError, AttributeError):
+            pass
 
 
 def _extract_input_content(input_val: Any) -> str:
@@ -173,30 +234,36 @@ def _extract_arguments_str(args_val: Any) -> str:
 
 def _set_tool_invocation_input(
     invocation: ToolInvocation,
-    instance: Any,
+    instance: FunctionCall,
     capture_content: bool,
 ) -> None:
     if capture_content:
-        arguments = getattr(instance, "arguments", None)
+        arguments = instance.arguments
         if arguments is not None:
             invocation.arguments = _extract_arguments_str(arguments)
 
 
+def _fail_tool_invocation(
+    invocation: ToolInvocation,
+    result: FunctionExecutionResult,
+) -> None:
+    error = result.error
+    invocation.fail(
+        Error(
+            type=ErrorTypeValues.OTHER.value,
+            message=str(error) if error else None,
+        )
+    )
+
+
 def _set_tool_invocation_output(
     invocation: ToolInvocation,
-    result: Any,
+    result: FunctionExecutionResult,
     capture_content: bool,
 ) -> None:
-    if getattr(result, "status", None) == "failure":
-        error = getattr(result, "error", None)
-        invocation.fail(
-            Error(
-                type=ErrorTypeValues.OTHER.value,
-                message=str(error) if error else None,
-            )
-        )
+    if result.status == "failure":
         return
-    if capture_content and result is not None:
+    if capture_content:
         invocation.tool_result = _extract_output_content(result)
 
 
@@ -241,6 +308,12 @@ def _set_invocation_output(
     session_id = getattr(result, "session_id", None)
     if session_id:
         invocation.conversation_id = str(session_id)
+    if isinstance(invocation, AgentInvocation):
+        if not invocation._request_model:
+            model = getattr(result, "model", None)
+            if model:
+                invocation._request_model = str(model)
+                invocation.attributes[GenAI.GEN_AI_REQUEST_MODEL] = str(model)
 
 
 def _start_agent_invocation(
@@ -251,7 +324,21 @@ def _start_agent_invocation(
     capture_content: bool,
 ) -> AgentInvocation:
     agent_name = getattr(instance, "name", None)
-    invocation = handler.invoke_local_agent(agent_name=agent_name)
+    model_obj = getattr(instance, "model", None)
+    request_model = None
+    if model_obj is not None:
+        request_model = getattr(model_obj, "id", None) or (
+            model_obj if isinstance(model_obj, str) else None
+        )
+
+    invocation = handler.invoke_local_agent(
+        agent_name=str(agent_name) if agent_name else None,
+        request_model=str(request_model) if request_model else None,
+    )
+    description = getattr(instance, "description", None)
+    if description:
+        invocation.agent_description = str(description)
+
     _set_invocation_input(invocation, instance, args, kwargs, capture_content)
     invocation.tool_definitions = prepare_tool_definitions(
         getattr(instance, "tools", None)
@@ -261,13 +348,13 @@ def _start_agent_invocation(
 
 def _start_tool_invocation(
     handler: TelemetryHandler,
-    instance: Any,
+    instance: FunctionCall,
     capture_content: bool,
 ) -> ToolInvocation:
-    function_obj = getattr(instance, "function", None)
+    function_obj = instance.function
     tool_name = getattr(function_obj, "name", None) or "tool"
     tool_desc = getattr(function_obj, "description", None)
-    tool_call_id = getattr(instance, "call_id", None)
+    tool_call_id = instance.call_id
 
     invocation = handler.tool(
         name=str(tool_name),
@@ -374,23 +461,57 @@ def _agent_arun(
     return traced_method
 
 
+def _handle_tool_result(
+    invocation: ToolInvocation,
+    instance: FunctionCall,
+    result: FunctionExecutionResult,
+    capture_content: bool,
+) -> FunctionExecutionResult:
+    if result.status == "failure":
+        _fail_tool_invocation(invocation, result)
+        return result
+
+    tool_res = result.result
+    wrapped_stream: Any = None
+    if isinstance(tool_res, AsyncIterator):
+        wrapped_stream = AsyncAgnoToolStreamWrapper(
+            cast(Any, tool_res), invocation, capture_content
+        )
+    elif isinstance(tool_res, Iterator):
+        wrapped_stream = AgnoToolStreamWrapper(
+            cast(Any, tool_res), invocation, capture_content
+        )
+
+    if wrapped_stream is not None:
+        result.result = wrapped_stream
+        instance.result = wrapped_stream
+        return result
+
+    _set_tool_invocation_output(invocation, result, capture_content)
+    invocation.stop()
+    return result
+
+
 def _tool_call_execute(
     handler: TelemetryHandler,
 ) -> Callable[..., Any]:
     capture_content = handler.should_capture_content()
 
     def traced_method(
-        wrapped: Callable[..., Any],
-        instance: Any,
+        wrapped: Callable[..., FunctionExecutionResult],
+        instance: FunctionCall,
         args: tuple[Any, ...],
         kwargs: dict[str, Any],
-    ) -> Any:
-        with _start_tool_invocation(
-            handler, instance, capture_content
-        ) as invocation:
+    ) -> FunctionExecutionResult:
+        invocation = _start_tool_invocation(handler, instance, capture_content)
+        try:
             result = wrapped(*args, **kwargs)
-            _set_tool_invocation_output(invocation, result, capture_content)
-            return result
+        except Exception as error:
+            invocation.fail(error)
+            raise
+        return _handle_tool_result(
+            invocation, instance, result, capture_content
+        )
 
     return traced_method
 
@@ -401,17 +522,20 @@ def _tool_call_aexecute(
     capture_content = handler.should_capture_content()
 
     async def traced_method(
-        wrapped: Callable[..., Awaitable[Any]],
-        instance: Any,
+        wrapped: Callable[..., Awaitable[FunctionExecutionResult]],
+        instance: FunctionCall,
         args: tuple[Any, ...],
         kwargs: dict[str, Any],
-    ) -> Any:
-        with _start_tool_invocation(
-            handler, instance, capture_content
-        ) as invocation:
+    ) -> FunctionExecutionResult:
+        invocation = _start_tool_invocation(handler, instance, capture_content)
+        try:
             result = await wrapped(*args, **kwargs)
-            _set_tool_invocation_output(invocation, result, capture_content)
-            return result
+        except Exception as error:
+            invocation.fail(error)
+            raise
+        return _handle_tool_result(
+            invocation, instance, result, capture_content
+        )
 
     return cast(Callable[..., Any], traced_method)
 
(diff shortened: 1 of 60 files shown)


instrumentation/opentelemetry-instrumentation-genai-agno/src/opentelemetry/instrumentation/genai/agno/patch.py
```diff
@@ -374,23 +492,57 @@ async def _await_result() -> object:
...
-        with _start_tool_invocation(
-            handler, instance, capture_content
-        ) as invocation:
+    ) -> FunctionExecutionResult:
+        invocation = _start_tool_invocation(handler, instance, capture_content)
+        try:
             result = wrapped(*args, **kwargs)
-            _set_tool_invocation_output(invocation, result, capture_content)
-            return result
+        except Exception as error:
```
lmolkova:
```suggestion
        except BaseException as error:
```

Catching `Exception` misses `BaseException` subclasses such as `asyncio.CancelledError`. If a task is cancelled, `invocation.fail(error)` is bypassed and the span and context token leak. Same on lines 564, 768, and 798.

Test demonstrating the leak on cancellation:
```python
async def test_tool_call_cancellation_fails_invocation(span_exporter):
    async def slow_tool():
        await asyncio.sleep(10)

    call = FunctionCall(function=Function.from_callable(slow_tool), arguments={})
    task = asyncio.create_task(call.aexecute())
    await asyncio.sleep(0.01)
    task.cancel()
    with pytest.raises(asyncio.CancelledError):
        await task

    assert len(spans) == 1
    assert spans[0].attributes.get("error.type") == "CancelledError"
```

instrumentation/opentelemetry-instrumentation-genai-agno/src/opentelemetry/instrumentation/genai/agno/stream.py
```diff
@@ -309,3 +319,57 @@ def __init__(
...
+class AgnoToolStreamWrapper(SyncStreamWrapper[Any]):
+    """Stream wrapper for synchronous tool executions that return iterators/generators."""
+
+    def __init__(
+        self,
+        stream: Any,
+        invocation: ToolInvocation,
+        capture_content: bool,
+    ) -> None:
+        super().__init__(stream)
```
lmolkova:
```suggestion
        super().__init__(stream, invocation=invocation)
```

Pass `invocation` to `super().__init__` so `self._self_invocation` is set on the base stream wrapper. Without it, stream chunk timing metrics are not recorded. Same on line 360 for `AsyncAgnoToolStreamWrapper`.

Test:
```python
def test_tool_stream_wrapper_receives_invocation():
    inv = handler.tool("test")
    wrapper = AgnoToolStreamWrapper(iter(["chunk"]), inv, True)
    assert wrapper._self_invocation is inv
```

instrumentation/opentelemetry-instrumentation-genai-agno/src/opentelemetry/instrumentation/genai/agno/stream.py
```diff
@@ -309,3 +319,57 @@ def __init__(
...
+        self._self_capture_content = capture_content
+        self._self_chunks: list[str] = []
+
+    def _process_chunk(self, chunk: Any) -> None:
+        if self._self_capture_content:
+            self._self_chunks.append(format_content(chunk))
+
+    def _on_stream_end(self) -> None:
+        if self._self_capture_content:
+            self._self_tool_invocation.tool_result = "".join(self._self_chunks)
```
lmolkova: `"".join(self._self_chunks)` assumes streamed chunks are text. When a tool yields structured items (e.g. dicts), `format_content` stringifies each item independently, resulting in concatenated fragments like `{"a": 1}{"b": 2}` instead of valid JSON. Same on line 371.

Test:
```python
def test_tool_stream_structured_chunks(instrument_agno_content_capture, span_exporter):
    def gen_dicts():
        yield {"chunk": 1}
        yield {"chunk": 2}

    call = FunctionCall(function=Function.from_callable(gen_dicts), arguments={})
    list(call.execute().result)
    result_str = span.attributes.get(GenAIAttributes.GEN_AI_TOOL_CALL_RESULT)
    assert json.loads(result_str) == [{"chunk": 1}, {"chunk": 2}]
```

instrumentation/opentelemetry-instrumentation-genai-agno/src/opentelemetry/instrumentation/genai/agno/patch.py
```diff
@@ -251,7 +353,23 @@ def _start_agent_invocation(
...
+        request_model = getattr(model_obj, "id", None) or (
+            model_obj if isinstance(model_obj, str) else None
+        )
+
+    invocation = handler.invoke_local_agent(
+        agent_name=str(agent_name) if agent_name else None,
+        request_model=str(request_model) if request_model else None,
+    )
+    if request_model:
+        invocation.attributes[GenAI.GEN_AI_REQUEST_MODEL] = str(request_model)
```
lmolkova:
```suggestion
```

`handler.invoke_local_agent(..., request_model=str(request_model))` already records `gen_ai.request.model` on span creation, so setting it manually here is redundant.

instrumentation/opentelemetry-instrumentation-genai-agno/src/opentelemetry/instrumentation/genai/agno/patch.py
```diff
@@ -241,6 +337,12 @@ def _set_invocation_output(
     session_id = getattr(result, "session_id", None)
     if session_id:
         invocation.conversation_id = str(session_id)
+    if isinstance(invocation, AgentInvocation):
+        model = getattr(result, "model", None)
+        if model:
+            invocation.attributes.setdefault(
+                GenAI.GEN_AI_REQUEST_MODEL, str(model)
+            )
```
lmolkova:
```suggestion
            invocation.attributes.setdefault(
                GenAI.GEN_AI_REQUEST_MODEL, str(model)
            )
            invocation.metric_attributes.setdefault(
                GenAI.GEN_AI_REQUEST_MODEL, str(model)
            )
```

Setting only `invocation.attributes` adds the model to the span, but duration metrics (`invoke_agent_duration`) read from `metric_attributes`. Same on line 57 of `stream.py`.

instrumentation/opentelemetry-instrumentation-genai-agno/src/opentelemetry/instrumentation/genai/agno/patch.py
```diff
@@ -58,92 +77,163 @@
...
+            return
+        try:
+            parts = target_name.split(".")
+            curr = mod
+            for part in parts:
+                curr = getattr(curr, part)
+        except AttributeError:
+            # Target class or method may not exist across all supported Agno versions.
+            return
+        if hasattr(curr, "__wrapped__"):
```
lmolkova:
```suggestion
        if isinstance(curr, BaseObjectProxy):
            return
```

`hasattr(curr, "__wrapped__")` returns `True` if upstream ever uses `@functools.wraps` on target methods, which would skip instrumentation.

instrumentation/opentelemetry-instrumentation-genai-agno/src/opentelemetry/instrumentation/genai/agno/utils.py
```diff
@@ -8,14 +8,32 @@
...
+def format_retrieval_document(doc: Document) -> dict[str, Any]:
+    """Format an Agno Document into a retrieval document dict."""
+    doc_dict: dict[str, Any] = {"content": doc.content}
+    if doc.id is not None:
+        doc_dict["id"] = str(doc.id)
+    if doc.reranking_score is not None:
+        try:
+            doc_dict["score"] = float(doc.reranking_score)
+        except (ValueError, TypeError):
+            pass
```
lmolkova:
```suggestion
    score = (
        doc.reranking_score
        if doc.reranking_score is not None
        else getattr(doc, "score", None)
    )
    if score is not None:
        try:
            doc_dict["score"] = float(score)
        except (ValueError, TypeError):
            pass
```

Fall back to `getattr(doc, "score", None)` for vector stores that populate document score without reranking.

instrumentation/opentelemetry-instrumentation-genai-agno/tests/test_tools.py
```diff
@@ -256,3 +257,226 @@ def test_agent_run_without_tools(
     span = spans[0]
     assert span.name == "invoke_agent test-no-tools-agent"
     assert GenAIAttributes.GEN_AI_TOOL_DEFINITIONS not in span.attributes
+
+
+def test_tool_call_execute_sync(
```
lmolkova: Missing test coverage for `FunctionCall`:
- `FunctionCall.aexecute` in non-streaming mode (only sync `FunctionCall.execute` is tested).
- Tool execution with content capture disabled for both sync and async (verifying `gen_ai.tool.call.result` and `arguments` are suppressed).

instrumentation/opentelemetry-instrumentation-genai-agno/tests/test_knowledge.py
```diff
@@ -0,0 +1,225 @@
...
+    assert (
+        span.attributes.get(GenAIAttributes.GEN_AI_DATA_SOURCE_ID)
+        == "private_kb"
+    )
+    assert span.attributes.get("gen_ai.retrieval.top_k") == 3
+    assert GenAIAttributes.GEN_AI_RETRIEVAL_QUERY_TEXT not in span.attributes
+    assert GenAIAttributes.GEN_AI_RETRIEVAL_DOCUMENTS not in span.attributes
+
+
+def test_knowledge_asearch_content_capture(
```
lmolkova: Missing test coverage for `Knowledge`:
- `Knowledge.asearch` with content capture disabled (only sync `Knowledge.search` is tested without content capture).
- `Knowledge.asearch` on error (only sync `Knowledge.search` error is tested).

instrumentation/opentelemetry-instrumentation-genai-agno/src/opentelemetry/instrumentation/genai/agno/patch.py
lmolkova: Catching `Exception` leaves cancelled tool invocations open and leaks their context. Catch `BaseException`, finalize the invocation, and re-raise unchanged in both tool wrappers and both retrieval wrappers.

This test fails on the PR and passes on its base (Agno 2.9):

```python
def test_cancelled_tool_finishes(instrument_agno, span_exporter):
    error = asyncio.CancelledError('cancelled')
    async def cancelled():
        raise error
    async def run():
        call = FunctionCall(function=Function.from_callable(cancelled), arguments={})
        with pytest.raises(asyncio.CancelledError) as caught:
            await call.aexecute()
        assert caught.value is error
        assert len(spans) == 1
        assert spans[0].attributes['error.type'] == 'asyncio.exceptions.CancelledError'
    asyncio.run(run())
```

instrumentation/opentelemetry-instrumentation-genai-agno/src/opentelemetry/instrumentation/genai/agno/stream.py
lmolkova:
```suggestion
        super().__init__(stream, invocation=invocation)
```

Pass `invocation` to `super().__init__` so `self._self_invocation` is set on the base stream wrapper. Without it, stream chunk timing metrics are not recorded. Same on line 360 for `AsyncAgnoToolStreamWrapper`.

Test:
```python
def test_tool_stream_wrapper_receives_invocation():
    inv = handler.tool("test")
    wrapper = AgnoToolStreamWrapper(iter(["chunk"]), inv, True)
    assert wrapper._self_invocation is inv
```

instrumentation/opentelemetry-instrumentation-genai-agno/src/opentelemetry/instrumentation/genai/agno/stream.py
lmolkova: `"".join(self._self_chunks)` assumes streamed chunks are text. When a tool yields structured items (e.g. dicts), `format_content` stringifies each item independently, resulting in concatenated fragments like `{"a": 1}{"b": 2}` instead of valid JSON. Same on line 371.

Test:
```python
def test_tool_stream_structured_chunks(instrument_agno_content_capture, span_exporter):
    def gen_dicts():
        yield {"chunk": 1}
        yield {"chunk": 2}

    call = FunctionCall(function=Function.from_callable(gen_dicts), arguments={})
    list(call.execute().result)
    result_str = span.attributes.get(GenAIAttributes.GEN_AI_TOOL_CALL_RESULT)
    assert json.loads(result_str) == [{"chunk": 1}, {"chunk": 2}]
```

instrumentation/opentelemetry-instrumentation-genai-agno/src/opentelemetry/instrumentation/genai/agno/patch.py
lmolkova: nit: `invoke_local_agent(request_model=...)` already records this attribute. Remove the redundant assignment.

instrumentation/opentelemetry-instrumentation-genai-agno/src/opentelemetry/instrumentation/genai/agno/patch.py
lmolkova:
```suggestion
```

`gen_ai.request.model` represents the model requested upfront and should only be set at span creation time via `handler.invoke_local_agent(request_model=...)`. Extracting it post-hoc from `RunOutput` (here) or stream chunks (`stream.py:55-60`) conflates request and response metadata and works around the util's design where `request_model` is intentionally an immutable start parameter for consistent sampling and metrics. Drop late model extraction here and in `stream.py`.

instrumentation/opentelemetry-instrumentation-genai-agno/src/opentelemetry/instrumentation/genai/agno/patch.py
lmolkova:
```suggestion
        if isinstance(curr, BaseObjectProxy):
            return
```

`hasattr(curr, "__wrapped__")` returns `True` if upstream ever uses `@functools.wraps` on target methods, which would skip instrumentation.

instrumentation/opentelemetry-instrumentation-genai-agno/src/opentelemetry/instrumentation/genai/agno/utils.py
lmolkova:
```suggestion
    score = (
        doc.reranking_score
        if doc.reranking_score is not None
        else getattr(doc, "score", None)
    )
    if score is not None:
        try:
            doc_dict["score"] = float(score)
        except (ValueError, TypeError):
            pass
```

Fall back to `getattr(doc, "score", None)` for vector stores that populate document score without reranking.

instrumentation/opentelemetry-instrumentation-genai-agno/tests/test_tools.py
lmolkova, line 263: Missing test coverage for `FunctionCall`:
- `FunctionCall.aexecute` in non-streaming mode (only sync `FunctionCall.execute` is tested).
- Tool execution with content capture disabled for both sync and async (verifying `gen_ai.tool.call.result` and `arguments` are suppressed).

instrumentation/opentelemetry-instrumentation-genai-agno/tests/test_knowledge.py
lmolkova, line 115: Missing test coverage for `Knowledge`:
- `Knowledge.asearch` with content capture disabled (only sync `Knowledge.search` is tested without content capture).
- `Knowledge.asearch` on error (only sync `Knowledge.search` error is tested).

instrumentation/opentelemetry-instrumentation-genai-agno/src/opentelemetry/instrumentation/genai/agno/patch.py
```diff
@@ -374,23 +492,57 @@ async def _await_result() -> object:
...
+        )
+    elif isinstance(tool_res, Iterator):
+        wrapped_stream = AgnoToolStreamWrapper(
+            cast(Any, tool_res), invocation, capture_content
+        )
+
+    if wrapped_stream is not None:
+        result.result = wrapped_stream
+        instance.result = wrapped_stream
+        return result
```
lmolkova, line 510: Returning a tool stream leaves its span active in the caller context, so unrelated work before the stream is drained becomes its child. Restore the caller context before returning, and activate the tool context while consuming the stream.

Add this test to `tests/test_tools.py`. It fails on the PR and passes on its base (Agno 2.9):

```python
def test_tool_stream_restores_caller_context(instrument_agno) -> None:
    from collections.abc import Iterator
    from opentelemetry.trace import get_current_span

    def streaming_tool() -> Iterator[str]:
        yield "chunk"

    caller = get_current_span()
    call = FunctionCall(
        function=Function.from_callable(streaming_tool), arguments={}
    )
    result = call.execute()
    current_after_return = get_current_span()
    assert list(result.result) == ["chunk"]
    assert current_after_return is caller
```

instrumentation/opentelemetry-instrumentation-genai-agno/src/opentelemetry/instrumentation/genai/agno/patch.py
lmolkova: Agno sets `RunOutput.model` from `agent.model.id` before execution, so this is configured-model metadata, not a provider response model ([source](https://github.com/agno-agi/agno/blob/v2.9.0/libs/agno/agno/agent/_run.py#L1313)). Capture it at invocation start when applicable; remove this fallback rather than rename it to `gen_ai.response.model`, which is not defined for [internal agent spans](https://github.com/open-telemetry/semantic-conventions-genai/blob/main/docs/gen-ai/gen-ai-agent-spans.md#invoke-agent-internal-span).

```suggestion
```

instrumentation/opentelemetry-instrumentation-genai-agno/src/opentelemetry/instrumentation/genai/agno/stream.py
```diff
@@ -48,6 +52,12 @@ def _process_chunk(self, chunk: Any) -> None:
         if session_id and not self._self_agent_invocation.conversation_id:
             self._self_agent_invocation.conversation_id = str(session_id)
 
+        model = getattr(chunk, "model", None)
+        if model:
+            self._self_agent_invocation.attributes.setdefault(
+                GenAI.GEN_AI_REQUEST_MODEL, str(model)
+            )
```
lmolkova: Please remove the stream fallback too. The run model comes from Agno's configured `agent.model.id`, not a provider response. Keep configured-model capture at invocation start; do not replace this with `gen_ai.response.model`.

```suggestion
```

instrumentation/opentelemetry-instrumentation-genai-agno/src/opentelemetry/instrumentation/genai/agno/patch.py
DylanRussell: Ack.. switched to BaseException

instrumentation/opentelemetry-instrumentation-genai-agno/src/opentelemetry/instrumentation/genai/agno/patch.py
DylanRussell: Ack

instrumentation/opentelemetry-instrumentation-genai-agno/tests/test_tools.py
DylanRussell: Ack added tests

instrumentation/opentelemetry-instrumentation-genai-agno/tests/test_knowledge.py
DylanRussell: Ack added tests

instrumentation/opentelemetry-instrumentation-genai-agno/src/opentelemetry/instrumentation/genai/agno/patch.py
DylanRussell: Great point.. I ended up creating new classes `SyncToolStreamWrapper` and `AsyncToolStreamWrapper` in utils to handle this case but it is tricky. 

Check out commit [8a15a66](https://github.com/open-telemetry/opentelemetry-python-genai/pull/673/commits/8a15a66161fc1cd95ba93170e07a25881b14c8d6)

In Agno, synchronous streaming tools execute sequentially and are iterated to completion on the same thread. In async runs, tools execute concurrently in separate tasks/threads; because context is copied upon creation, each tool correctly inherits the parent context and is recorded as a sibling span.

But these classes don't work well if someone calls `next` just once on the iterator/generator and then leaves the stream for a while and comes back later (be cause the tool call will be the parent span in the context).. It's hard to design this without knowing exactly what the use case is though.. Agno is the only one doing this right now..

I also directly access `GenAiInvocation._context_token` inside the stream wrapper classes, maybe theres a better way to do that..

instrumentation/opentelemetry-instrumentation-genai-agno/src/opentelemetry/instrumentation/genai/agno/patch.py
DylanRussell: Ack good catch. fixed

instrumentation/opentelemetry-instrumentation-genai-agno/src/opentelemetry/instrumentation/genai/agno/stream.py
DylanRussell: yep removed

instrumentation/opentelemetry-instrumentation-genai-agno/src/opentelemetry/instrumentation/genai/agno/patch.py
lmolkova: Maybe we should not attach context at all for streaming tools? Yes, correlation with child spans created inside the generator might be lost, but leaking context is not great.

`_activate_context()` leaks the tool span into the caller thread during iteration, between chunks, or if iteration breaks early. Here is a test showing the leak:

```python
def test_tool_stream_partial_iteration_restores_caller_context(instrument_agno) -> None:
    from collections.abc import Iterator
    from opentelemetry.trace import get_current_span

    def streaming_tool() -> Iterator[str]:
        yield "chunk1"
        yield "chunk2"

    caller = get_current_span()
    call = FunctionCall(
        function=Function.from_callable(streaming_tool), arguments={}
    )
    result = call.execute()
    it = iter(result.result)
    assert next(it) == "chunk1"
    assert get_current_span() is caller
```

If you want to try to prevent leaks, you can use the `.context` property on the invocation instead of private `_context_token` - just added in one of the recent PRs.

DylanRussell: I added a `suspend` method and `activate` contextmanager to the base GenAIInvocation  so that we can wrap all the streaming tool methods and toggle the context on and off.. let me know what you think

util/opentelemetry-util-genai/src/opentelemetry/util/genai/_invocation.py
```diff
@@ -151,6 +152,35 @@ def _start(
...
+        it, and a no-op once the invocation has finished.
+        """
+        if self._finished or self._context_token is not None:
+            yield
+            return
+        self._context_token = attach(self._span_context)
+        try:
+            yield
+        finally:
+            self.suspend()
```
lmolkova, line 182:
```suggestion
        if self._finished:
            yield
            return
        token = attach(self.context)
        try:
            yield
        finally:
            detach(token)
```
We can use the public `self.context` property here instead of `self._span_context`.

Also, using a local `token` inside `activate()` instead of mutating `self._context_token` keeps context activation safe across concurrent async tasks or threads (avoiding token detachment conflicts or `ValueError`).

instrumentation/opentelemetry-instrumentation-genai-agno/src/opentelemetry/instrumentation/genai/agno/patch.py
```diff
@@ -374,23 +483,57 @@ async def _await_result() -> object:
...
+    tool_res = result.result
+    wrapped_stream: Any = None
+    if isinstance(tool_res, AsyncIterator):
+        wrapped_stream = AsyncAgnoToolStreamWrapper(
+            cast(Any, tool_res), invocation, capture_content
+        )
+    elif isinstance(tool_res, Iterator):
+        wrapped_stream = AgnoToolStreamWrapper(
+            cast(Any, tool_res), invocation, capture_content
+        )
```
lmolkova, line 505:
```suggestion
    if isinstance(tool_res, AsyncIterator):
        wrapped_stream = AsyncToolStreamWrapper(cast(Any, tool_res), invocation)
    elif isinstance(tool_res, Iterator):
        wrapped_stream = SyncToolStreamWrapper(cast(Any, tool_res), invocation)
```
`SyncToolStreamWrapper` and `AsyncToolStreamWrapper` in util already handle chunk accumulation and check `invocation.should_capture_content`. Can we use them directly here rather than maintaining `AgnoToolStreamWrapper` and `AsyncAgnoToolStreamWrapper` with redundant `capture_content` parameters?

instrumentation/opentelemetry-instrumentation-genai-agno/src/opentelemetry/instrumentation/genai/agno/patch.py
```diff
@@ -522,3 +668,133 @@ async def _await_result() -> object:
...
+        try:
+            result = wrapped(*args, **kwargs)
+        except BaseException as error:
+            invocation.fail(error)
+            raise
+
+        if capture_content and result is not None:
+            invocation.documents = [
+                format_retrieval_document(doc) for doc in result
+            ]
```
lmolkova, line 766:
```suggestion
        if invocation.should_capture_content and result is not None:
            invocation.documents = [
                format_retrieval_document(doc) for doc in result
            ]
```
Check `invocation.should_capture_content` directly rather than snapshotting `capture_content = handler.should_capture_content()` at patch time (and similarly in `_knowledge_asearch`).

instrumentation/opentelemetry-instrumentation-genai-agno/src/opentelemetry/instrumentation/genai/agno/utils.py
```diff
@@ -8,14 +8,32 @@
...
 
 from opentelemetry.util.genai.types import (
     FunctionToolDefinition,
     ToolDefinition,
 )
 
 
+def format_retrieval_document(doc: Document) -> dict[str, Any]:
+    """Format an Agno Document into a retrieval document dict."""
+    doc_dict: dict[str, Any] = {"content": doc.content}
```
lmolkova, line 24: let's use util's model here too?

util/opentelemetry-util-genai/src/opentelemetry/util/genai/_invocation.py
DylanRussell: Good idea.. done

instrumentation/opentelemetry-instrumentation-genai-agno/src/opentelemetry/instrumentation/genai/agno/patch.py
DylanRussell: good point.. done

instrumentation/opentelemetry-instrumentation-genai-agno/src/opentelemetry/instrumentation/genai/agno/patch.py
DylanRussell: done

instrumentation/opentelemetry-instrumentation-genai-agno/src/opentelemetry/instrumentation/genai/agno/utils.py
DylanRussell: Ack.. getting rid of content / metadata and using utils model.

util/opentelemetry-util-genai/src/opentelemetry/util/genai/stream.py
```diff
@@ -365,6 +368,126 @@ async def __anext__(self) -> ChunkT:
...
+                self._self_chunks.append(chunk)
+            else:
+                try:
+                    self._self_chunks.append(gen_ai_json_dumps(chunk))
+                except Exception:
+                    self._self_chunks.append(str(chunk))
+
+    def _on_stream_end(self) -> None:
+        if self._self_tool_invocation.should_capture_content:
+            if all(isinstance(c, str) for c in self._self_chunks):
```
lmolkova, line 420: Serializing non-string chunks into strings in `_process_chunk` causes them to be concatenated into malformed JSON (e.g. `{"k1": "v1"}{"k2": "v2"}`) instead of preserving a list of items.

Here is a test showing the issue:

```python
def test_sync_tool_stream_wrapper_non_string_chunks():
    invocation, span_exporter = _started_tool_invocation()
    chunks = [{"k1": "v1"}, {"k2": "v2"}]
    stream = _FakeSyncStream(chunks=chunks)
    wrapper = SyncToolStreamWrapper(stream, invocation)

    assert list(wrapper) == chunks

    assert invocation.tool_result == chunks
    assert len(spans) == 1
    assert spans[0].attributes["gen_ai.tool.call.result"] == json.dumps(chunks)
```

instrumentation/opentelemetry-instrumentation-genai-agno/tests/test_tools.py
```diff
@@ -287,3 +289,471 @@ def sample_tool(location: str) -> str:
...
+    )
+    with pytest.raises(KeyboardInterrupt) as caught:
+        call.execute()
+    assert caught.value is error
+    spans = span_exporter.get_finished_spans()
+    assert len(spans) == 1
+    assert spans[0].attributes["error.type"] == "KeyboardInterrupt"
+
+
+def test_tool_stream_restores_caller_context(
```
lmolkova, line 687: Please add tests for Agno streaming tools with content capture enabled, as well as mid-iteration error and early break scenarios.

util/opentelemetry-util-genai/src/opentelemetry/util/genai/stream.py
DylanRussell: Ack, fixed this

instrumentation/opentelemetry-instrumentation-genai-agno/tests/test_tools.py
DylanRussell: Added tests

util/opentelemetry-util-genai/src/opentelemetry/util/genai/stream.py
```diff
@@ -402,6 +404,126 @@ async def __anext__(self) -> ChunkT:
...
+        exc_val: BaseException | None,
+        exc_tb: TracebackType | None,
+    ) -> Literal[False]:
+        with self._self_tool_invocation.activate():
+            return super().__exit__(exc_type, exc_val, exc_tb)
+
+    def __del__(self) -> None:
+        try:
+            self._finalize_success()
+        except BaseException:  # pylint: disable=broad-exception-caught
```
lmolkova:
```suggestion
    def __del__(self) -> None:
        try:
            self._finalize_failure(
                GeneratorExit("Stream garbage collected before completion")
            )
        except BaseException:  # pylint: disable=broad-exception-caught
            pass
```
Finalizing as success on `__del__` treats abandoned streams as successful. Should this record a failure instead?

instrumentation/opentelemetry-instrumentation-genai-agno/src/opentelemetry/instrumentation/genai/agno/patch.py
```diff
@@ -59,92 +77,163 @@
...
+        try:
+            parts = target_name.split(".")
+            curr = mod
+            for part in parts:
+                curr = getattr(curr, part)
+        except AttributeError:
+            # Target class or method may not exist across all supported Agno versions.
+            return
+        if hasattr(curr, "__wrapped__"):
+            return
```
lmolkova:
```suggestion
        if isinstance(curr, ObjectProxy):
            return
```
`hasattr(curr, "__wrapped__")` is true for any method decorated with `@functools.wraps`. Check `isinstance(curr, ObjectProxy)` to specifically check for wrapt's wrapper.

instrumentation/opentelemetry-instrumentation-genai-agno/src/opentelemetry/instrumentation/genai/agno/patch.py
```diff
@@ -560,3 +696,126 @@ async def _await_result() -> object:
...
+        else None,
+        provider=str(provider).lower() if provider is not None else None,
+        request_model=str(request_model)
+        if request_model is not None
+        else None,
+    )
+
+    query = get_argument("query", wrapped, args, kwargs)
+    if query is not None:
+        invocation.query_text = str(query)
```
lmolkova:
```suggestion
    if invocation.should_capture_content:
        query = get_argument("query", wrapped, args, kwargs)
        if query is not None:
            invocation.query_text = str(query)
```
Extract and set `query_text` only when `invocation.should_capture_content` is enabled, matching document capture.

instrumentation/opentelemetry-instrumentation-genai-agno/tests/test_knowledge.py
```diff
@@ -0,0 +1,326 @@
...
+        span.attributes.get(GenAIAttributes.GEN_AI_OPERATION_NAME)
+        == "retrieval"
+    )
+    assert (
+        span.attributes.get(GenAIAttributes.GEN_AI_DATA_SOURCE_ID) == "test_kb"
+    )
+    assert (
+        span.attributes.get(GenAIAttributes.GEN_AI_PROVIDER_NAME) == "pgvector"
+    )
+    assert span.attributes.get("gen_ai.retrieval.top_k") == 5
```
lmolkova:
```suggestion
    assert (
        span.attributes.get(GenAIAttributes.GEN_AI_RETRIEVAL_TOP_K) == 5
    )
```
nit: use `GenAIAttributes.GEN_AI_RETRIEVAL_TOP_K` instead of literal strings here and across lines 103, 146, 190.

util/opentelemetry-util-genai/src/opentelemetry/util/genai/stream.py
lmolkova:
```suggestion
    def __del__(self) -> None:
        try:
            self._finalize_failure(
                GeneratorExit("Stream garbage collected before completion")
            )
        except BaseException:  # pylint: disable=broad-exception-caught
            pass
```
Finalizing as success on `__del__` treats abandoned streams as successful. Should this record a failure instead?

util/opentelemetry-util-genai/src/opentelemetry/util/genai/stream.py
```diff
@@ -402,6 +404,126 @@ async def __anext__(self) -> ChunkT:
...
+        invocation: ToolInvocation,
+    ) -> None:
+        super().__init__(stream)
+        self._self_tool_invocation = invocation
+        invocation.suspend()
+        self._self_chunks: list[Any] = []
+
+    def __del__(self) -> None:
+        try:
+            self._finalize_success()
```
lmolkova:
```suggestion
    def __del__(self) -> None:
        try:
            self._finalize_failure(
                GeneratorExit("Stream garbage collected before completion")
            )
        except BaseException:  # pylint: disable=broad-exception-caught
            pass
```
Same here; finalize with failure on `__del__`.

instrumentation/opentelemetry-instrumentation-genai-agno/src/opentelemetry/instrumentation/genai/agno/patch.py
lmolkova:
```suggestion
        if isinstance(curr, ObjectProxy):
            return
```
`hasattr(curr, "__wrapped__")` is true for any method decorated with `@functools.wraps`. Check `isinstance(curr, ObjectProxy)` to specifically check for wrapt's wrapper.

instrumentation/opentelemetry-instrumentation-genai-agno/src/opentelemetry/instrumentation/genai/agno/patch.py
lmolkova:
```suggestion
    if invocation.should_capture_content:
        query = get_argument("query", wrapped, args, kwargs)
        if query is not None:
            invocation.query_text = str(query)
```
please extract and set `query_text` only when `invocation.should_capture_content` is enabled, matching document capture.

instrumentation/opentelemetry-instrumentation-genai-agno/tests/test_knowledge.py
lmolkova:
```suggestion
    assert span.attributes.get(GEN_AI_RETRIEVAL_TOP_K) == 5
```
nit: use `GEN_AI_RETRIEVAL_TOP_K` (from `opentelemetry.util.genai.semconv.gen_ai`) instead of literal strings here and across lines 103, 146, 190.

util/opentelemetry-util-genai/src/opentelemetry/util/genai/stream.py
lmolkova: I think we discussed somewhere that we should treat those as failures rather than success?

util/opentelemetry-util-genai/src/opentelemetry/util/genai/stream.py
lmolkova:
```suggestion
    def __del__(self) -> None:
        try:
            self._finalize_failure(
                GeneratorExit("Stream garbage collected before completion")
            )
        except BaseException:  # pylint: disable=broad-exception-caught
            pass
```
Same here; finalize with failure on `__del__`.

instrumentation/opentelemetry-instrumentation-genai-agno/src/opentelemetry/instrumentation/genai/agno/patch.py
lmolkova:
```suggestion
    if invocation.should_capture_content:
        query = get_argument("query", wrapped, args, kwargs)
        if query is not None:
            invocation.query_text = str(query)
```
Please extract and set `query_text` only when `invocation.should_capture_content` is enabled, matching document capture.

instrumentation/opentelemetry-instrumentation-genai-agno/tests/test_knowledge.py
lmolkova:
```suggestion
    assert span.attributes.get(GEN_AI_RETRIEVAL_TOP_K) == 5
```
nit: use `GEN_AI_RETRIEVAL_TOP_K` (from `opentelemetry.util.genai.semconv.gen_ai`) instead of literal strings here and across lines 103, 146, 190.

lmolkova (approved):

util/opentelemetry-util-genai/src/opentelemetry/util/genai/stream.py
DylanRussell: yes thats right.. I think that was the bedrock instrumentation which was merged into the wrong branch -- https://github.com/open-telemetry/opentelemetry-python-genai/pull/776 can u take a look when u have a chance.. changed this to finalize with a failure

instrumentation/opentelemetry-instrumentation-genai-agno/src/opentelemetry/instrumentation/genai/agno/patch.py
DylanRussell: Ack, fixed

instrumentation/opentelemetry-instrumentation-genai-agno/tests/test_knowledge.py
DylanRussell: Ack, fixed

lzchen (approved):

PR #671, opened 2026-09-10
feat(langchain): extract cache_write and modality token usage attributes

SAMurai-16:
# Description

Extracts and records `gen_ai.usage.cache_write.input_tokens` and modality breakdown attributes (`gen_ai.usage.{text,image,audio}.{input,output}_tokens`) on `InferenceInvocation` spans from LangChain `usage_metadata`. It supports both `cache_write` and legacy `cache_creation` keys while omitting `0`, negative, boolean, and non-integer values.

This aligns the LangChain instrumentation with the latest GenAI semantic conventions following the updates in #613.

Fixes #609

## Type of change

- [x] New feature (non-breaking change which adds functionality)

# How has this been tested?

Added unit tests in `test_callback_handler.py` verifying cache write/read, reasoning, and multimodal token breakdowns (`text`, `image`, `audio`), as well as ensuring invalid/zero values are filtered out.

Run all tests and linting locally:

- `uv run tox -e py312-test-instrumentation-genai-langchain-latest`
- `uv run pre-commit run --all-files`

# Checklist

- [x] Followed the style guidelines of this project
- [x] Changelog updated if the change requires an entry (`609.added`)
- [x] Unit tests added
- [x] Documentation updated

PR #616, opened 2026-09-04
Split AgentInvocation into LocalAgentInvocation and RemoteAgentInvocation

lmolkova:
## What does this change do?

Splits `AgentInvocation` into `LocalAgentInvocation` (`SpanKind.INTERNAL`) for in-process framework agents and `RemoteAgentInvocation` (`SpanKind.CLIENT`) for external agent services, and updates agent instrumentations and conformance scenarios to expect the new metrics.

## Why?

Aligns agent spans and metrics with GenAI semantic conventions ([agent spans spec](https://github.com/open-telemetry/semantic-conventions-genai/blob/main/docs/gen-ai/gen-ai-agent-spans.md)):

- **Span kind distinction**: In-process framework agents use `INTERNAL` spans, while remote agent calls use `CLIENT` spans.
- **Attributes on local agents**:
  - `gen_ai.agent.id` is removed from local agent spans (only applies to remote agents with service-managed identifiers).
  - Provider and server attributes (`gen_ai.provider.name`, `server.address`, `server.port`) are excluded on local agent spans.
- **Attributes on remote agents**: Supports `gen_ai.agent.id`, `gen_ai.agent.version`, `gen_ai.request.previous_response.id`, and cache token usage (`gen_ai.usage.cache_write.input_tokens`, `gen_ai.usage.cache_read.input_tokens`).
- **Duration metric**: Records `gen_ai.invoke_agent.duration` (`s`) on local agent runs (replacing `gen_ai.client.operation.duration`), tagged with `gen_ai.agent.name` and optional `gen_ai.request.model`.

PR #613, opened 2026-09-04
Add cache_write_input_tokens and modality token usage attributes

lmolkova:
## What does this change do?

Adds `cache_write_input_tokens` (deprecating `cache_creation_input_tokens`) and modality token breakdown attributes to invocations in `opentelemetry-util-genai`, updating `openai`, `google-genai`, `bedrock`, and `langchain` to record them where available.

**Deprecated attribute:**
- `gen_ai.usage.cache_creation.input_tokens` (replaced by `gen_ai.usage.cache_write.input_tokens`)

**New attributes:**
- `gen_ai.usage.cache_write.input_tokens`
- `gen_ai.usage.text.input_tokens`
- `gen_ai.usage.image.input_tokens`
- `gen_ai.usage.audio.input_tokens`
- `gen_ai.usage.text.output_tokens`
- `gen_ai.usage.image.output_tokens`
- `gen_ai.usage.audio.output_tokens`
- `gen_ai.usage.text.cache_read.input_tokens`
- `gen_ai.usage.image.cache_read.input_tokens`
- `gen_ai.usage.audio.cache_read.input_tokens`

## Why?

Aligns token usage attributes with GenAI semantic conventions ([spans spec](https://github.com/open-telemetry/semantic-conventions-genai/blob/main/docs/gen-ai/gen-ai-spans.md)).

PR #674, opened 2026-09-10
feat(google-genai): extract modality token usage attributes on generate_content

### LINKED ISSUE #564: [google-genai] Capture audio token counts on generate_content
> [!NOTE]

Part of #560.

Extract modality breakdown for audio tokens from `usage_metadata.prompt_tokens_details` and `usage_metadata.candidates_tokens_details` when available.

- Attributes: `gen_ai.usage.audio.input_tokens`, `gen_ai.usage.audio.output_tokens`

Krishnachaitanyakc:
Fixes #564.

`generate_content` now maps `usage_metadata.prompt_tokens_details`, `candidates_tokens_details` and `cache_tokens_details` onto the text, image and audio token usage attributes on the inference span.

The same breakdown already lands for `interactions` (#613) and for langchain (#671), so a `generate_content` call with audio or image input was reporting only the aggregate counts.

A streaming chunk redelivers cumulative usage, as `test_includes_token_counts_in_span_not_aggregated_from_responses` already records, so an arriving breakdown replaces the previous one. Merging would let a modality that drops out of a later chunk keep a stale count and push the breakdown over the total next to it.

Known gaps:

* `usage_metadata.tool_use_prompt_tokens_details` is not mapped, because `opentelemetry-util-genai` has no field for it.
* Video and document modalities have no attribute in `opentelemetry-util-genai`, so those entries are skipped.
* `interactions.py` keeps its own `_set_modality_tokens`. The payload shapes differ (`token_count` versus `tokens`, enum versus plain string) and #616 is already editing that file, so unifying them is better as a follow up.

### DIFF
--- a/instrumentation/opentelemetry-instrumentation-google-genai/src/opentelemetry/instrumentation/google_genai/generate_content.py
+++ b/instrumentation/opentelemetry-instrumentation-google-genai/src/opentelemetry/instrumentation/google_genai/generate_content.py
@@ -3,7 +3,7 @@
 
 import json
 import os
-from collections.abc import AsyncIterable, Callable, Iterable
+from collections.abc import AsyncIterable, Callable, Iterable, Mapping
 from typing import Any
 
 from google.genai.models import AsyncModels, Models
@@ -14,6 +14,7 @@ from google.genai.types import (
     GenerateContentConfig,
     GenerateContentConfigOrDict,
     GenerateContentResponse,
+    ModalityTokenCount,
     Tool,
     ToolUnionDict,
 )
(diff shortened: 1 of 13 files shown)


instrumentation/opentelemetry-instrumentation-google-genai/src/opentelemetry/instrumentation/google_genai/generate_content.py
```diff
@@ -346,6 +347,50 @@ def _apply_request_attributes(
...
+            continue
+        # modality is a MediaModality enum, so str() would yield
+        # "MediaModality.AUDIO" rather than the bare modality name.
+        field_name = fields.get(
+            str(getattr(modality, "value", modality)).lower()
+        )
+        if field_name is not None:
+            setattr(invocation, field_name, token_count)
+
+
```
lmolkova: This is the third copy of this mapping - `interactions.py` has one and langchain got one in #671. Let's move this onto `InferenceInvocation` in `opentelemetry-util-genai` instead of duplicating and trying to unify later.

Rough sketch (not final, needs details sorted out):

```python
class InferenceInvocation(...):
    def set_input_tokens(self, modality: Modality | str, tokens: int | None) -> None:
        self._set_modality_tokens(_INPUT, modality, tokens)

    def set_output_tokens(...) -> None: ...
    def set_cache_read_input_tokens(...) -> None: ...

    def _set_modality_tokens(self, bucket, modality, tokens) -> None:
        if not isinstance(tokens, int) or isinstance(tokens, bool) or tokens < 0:
            return
        name = str(getattr(modality, "value", modality)).lower()
        # drop unsupported modalities (video, doc) centrally
        ...
```

Call site here simplifies to:

```python
for entry in _get_response_property(response, "usage_metadata.prompt_tokens_details") or []:
    invocation.set_input_tokens(entry.modality, entry.token_count)
```

This removes the local dicts, enum/string normalization, and int checks. It also centralizes handling unmapped modalities and future token types across all instrumentations.

Note: `Modality` in `types.py` currently lacks `text`, need to add it.

It's a small change - let's add the util method and update `interactions.py` and langchain right away

Krishnachaitanyakc: Makes sense, pushed in d1a3a1a. Went with the whole iterable rather than per entry, so an arriving breakdown replaces rather than merges. Happy to switch.

`interactions.py` is in there, which #616 also touches.

DylanRussell (approved):

instrumentation/opentelemetry-instrumentation-google-genai/src/opentelemetry/instrumentation/google_genai/generate_content.py
```diff
@@ -423,6 +436,15 @@ def _apply_response_attributes(
...
         ) + thinking_tokens
+    invocation.set_input_tokens(
+        _modality_tokens(response, "usage_metadata.prompt_tokens_details")
+    )
+    invocation.set_output_tokens(
+        _modality_tokens(response, "usage_metadata.candidates_tokens_details")
+    )
+    invocation.set_cache_read_input_tokens(
+        _modality_tokens(response, "usage_metadata.cache_tokens_details")
+    )
```
lmolkova:
```suggestion
    if response.usage_metadata is not None:
        invocation.set_input_tokens(
            _modality_tokens(response, "usage_metadata.prompt_tokens_details")
        )
        invocation.set_output_tokens(
            _modality_tokens(response, "usage_metadata.candidates_tokens_details")
        )
        invocation.set_cache_read_input_tokens(
            _modality_tokens(response, "usage_metadata.cache_tokens_details")
        )
```
Avoids walking property paths on every intermediate streaming chunk that carries no usage.

lmolkova (approved):
LGTM with a small optimization comment. Thanks!

PR #698, opened 2026-09-12
Add LlamaIndex retrieval instrumentation

eternalcuriouslearner:
## Summary

Adds OpenTelemetry retrieval spans for LlamaIndex synchronous and asynchronous BaseRetriever operations. Captures query text and retrieved documents when content capture is enabled, records top-k, preserves retrieval errors, and adds tests, conformance coverage, and documentation.\n\nProvider model calls remain delegated to provider instrumentations.


instrumentation/opentelemetry-instrumentation-genai-llama-index/src/opentelemetry/instrumentation/genai/llama_index/_handler.py
```diff
@@ -298,6 +301,55 @@ def _request_model(agent: BaseWorkflowAgent) -> str | None:
...
+        return query.query_str
+    return None
+
+
+def _retrieval_top_k(retriever: BaseRetriever) -> int | None:
+    """Read the common top-k setting without requiring a retriever subtype."""
+    try:
+        top_k = getattr(retriever, "similarity_top_k", None)
+    except BaseException:
+        return None
```
lmolkova: This exception guard adds special handling for a hypothetical getter failure; other attribute reads in this handler are unguarded. Keep the same pattern here.

```suggestion
    top_k = getattr(retriever, "similarity_top_k", None)
```

instrumentation/opentelemetry-instrumentation-genai-llama-index/tests/test_retrieval.py
```diff
@@ -0,0 +1,161 @@
...
+    assert _retrieval_documents([object()]) is None
+
+
+def test_similarity_top_k_getter_failure_is_ignored() -> None:
+    class _FailingRetriever:
+        @property
+        def similarity_top_k(self) -> int:
+            raise KeyboardInterrupt
+
+    assert _retrieval_top_k(cast(BaseRetriever, _FailingRetriever())) is None
```
lmolkova: This artificial `KeyboardInterrupt` case only supports the unnecessary exception guard. Remove the test along with it.

```suggestion
```

lmolkova (approved):

PR #661, opened 2026-09-09
[`opentelemetry-instrumentation-genai-anthropic`] Instrument beta.messages create, stream, and parse (#550)

lmolkova:
Fixes #550

Instruments `anthropic.resources.beta.messages.Messages` and `AsyncMessages` (`create`, `stream`, and `parse`), including raw and streaming response helpers. 

Anthropic beta messages provide access to new API features (such as prompt caching, computer use, and MCP tools) while producing standard chat inference spans and metrics.

PR #699, opened 2026-09-12
fix(anthropic): classify server tool message parts

lmolkova:
Fixes Anthropic server tool calls and results being captured as client-side tool parts.

Anthropic-executed tools now use ServerToolCallPart and ServerToolCallResponsePart for response objects and replayed dictionary content. Current stable server-result variants are supported without importing version-specific SDK classes.

Known gap: beta message blocks remain scoped to #661.


instrumentation/opentelemetry-instrumentation-genai-anthropic/src/opentelemetry/instrumentation/genai/anthropic/utils.py
```diff
@@ -162,22 +208,24 @@ def _convert_content_block_to_part(
...
             block.thinking if isinstance(block, ThinkingBlock) else block.data
         )
         return ReasoningPart(content=content)
 
-    if isinstance(block, WebSearchToolResultBlock):
-        return ToolCallResponsePart(
-            response=block.model_dump().get("content"),
-            id=block.tool_use_id,
-        )
+    if getattr(
```
lmolkova: needed?

eternalcuriouslearner (approved):

PR #704, opened 2026-09-13
[opentelemetry-instrumentation-genai-llama-index] Simplify tool context propagation and use invocation context

lmolkova:
Uses `invocation.context` and passes `context` directly to `TelemetryHandler.tool(...)`, eliminating manual context attach/detach and tracking tokens in LlamaIndex span handler. Replaces untyped `GenAIInvocation` with a concrete invocation union.

Simplifies context propagation and avoids accessing internal `invocation.span`.


eternalcuriouslearner (approved):
Thanks a lot for adding this. I was planning to add this later in a separate PR as this is much cleaner and simple.

PR #705, opened 2026-09-13
[opentelemetry-instrumentation-genai-langchain] Set finish_reasons and clean up invocation state management

lmolkova:
Populates finish_reasons on InferenceInvocation from LangChain chat generation metadata per [gen_ai.response.finish_reasons](https://github.com/open-telemetry/semantic-conventions-genai/blob/0c87594975195608dc91b3f702e250a7b240c151/docs/gen-ai/gen-ai-spans.md?plain=1#L87). Replaces AgentInvocation with LocalAgentInvocation, avoids accessing internal invocation.span.is_recording(), and tracks agent names in _InvocationManager instead of reading internal invocation state.


instrumentation/opentelemetry-instrumentation-genai-langchain/src/opentelemetry/instrumentation/genai/langchain/callback_handler.py
```diff
@@ -522,6 +516,7 @@ def on_llm_end(
                             name=name_str,
                         )
                     output_messages.append(output_message)
+                    finish_reasons.append(finish_reason or "error")
```
lzchen: Nit: does it always make sense to report absent finish reasons as error? Should we use "unknown" like above?

instrumentation/opentelemetry-instrumentation-genai-langchain/src/opentelemetry/instrumentation/genai/langchain/callback_handler.py
```diff
@@ -775,23 +766,23 @@ def on_retriever_error(
...
             visited.add(current)
             entity = self._invocation_manager.get_invocation(current)
-            if isinstance(entity, AgentInvocation):
-                if nearest_agent is None:
-                    nearest_agent = entity
-                if entity.agent_name:
-                    ancestor_agent_names.add(entity.agent_name.lower())
+            if isinstance(entity, LocalAgentInvocation):
+                agent_name = self._invocation_manager.get_agent_name(current)
+                if nearest_agent_name is None:
```
lzchen: Can an agent be encountered but it's name is `None`? If we have unnamed agent and the next one is named, would this function return the next one as "nearest" incorrectly in this case?

lzchen (approved):
Two non-blocking questions/comments.

instrumentation/opentelemetry-instrumentation-genai-langchain/src/opentelemetry/instrumentation/genai/langchain/callback_handler.py
lmolkova: great catch! created #742 - turns out we don't record unnamed agents. Changed check here to a boolean flag (to work properly once we support unnamed agents)

instrumentation/opentelemetry-instrumentation-genai-langchain/src/opentelemetry/instrumentation/genai/langchain/callback_handler.py
lmolkova: semconv says to record and fallback to error

https://github.com/open-telemetry/semantic-conventions-genai/blob/c88d504ab3d9879f8e50d3cc87e69775e11db234/model/gen-ai/registry.yaml#L217-L224

thanks for catching this btw,  i updated 'unknown' to 'error'.

PR #710, opened 2026-09-13
[opentelemetry-instrumentation-genai-openai-agents] Do not record gen_ai.provider.name on tool execution metrics

lmolkova:
Removes setting gen_ai.provider.name on ToolInvocation.metric_attributes.

Tool execution does not invoke a GenAI provider, so per note [3] on [gen_ai.provider.name for gen_ai.client.operation.duration](https://github.com/open-telemetry/semantic-conventions-genai/blob/0c87594975195608dc91b3f702e250a7b240c151/docs/gen-ai/gen-ai-metrics.md?plain=1#L207-L208), it is only required when the operation involves a call to a GenAI provider.


eternalcuriouslearner (approved):
LGTM!!

PR #723, opened 2026-09-15
[`opentelemetry-instrumentation-genai-bedrock`] Add Bedrock Agent Runtime instrumentation

DylanRussell:
# Description

Adds instrumentation for AWS Bedrock Agent Runtime, supporting `invoke_agent` (remote agent invocations) and `retrieve` (Knowledge Base vector retrieval) across synchronous and asynchronous clients.

## Type of change

# How has this been tested?

Unit tests (sync and async)

# Checklist

- [x] Unit tests added
- [x] Documentation updated


instrumentation/opentelemetry-instrumentation-genai-bedrock/src/opentelemetry/instrumentation/genai/bedrock/extractors.py
```diff
@@ -925,3 +927,114 @@ def extract_embedding_response(
...
+    session_id = api_params.get("sessionId")
+    if session_id:
+        invocation.conversation_id = str(session_id)
+
+    # InvokeAgent takes an alias, not a version, and the response doesn't report the
+    # version the alias resolved to. The alias is the closest available identifier of
+    # which agent revision served the request.
+    agent_alias_id = api_params.get("agentAliasId")
+    if agent_alias_id:
+        invocation.agent_version = str(agent_alias_id)
```
lmolkova:
```suggestion
```
`agent_version` is already passed to `handler.invoke_remote_agent()` in `_start_invoke_agent`, so extracting and re-assigning it here is redundant.

instrumentation/opentelemetry-instrumentation-genai-bedrock/tests/test_agent.py
```diff
@@ -0,0 +1,731 @@
+# Copyright The OpenTelemetry Authors
+# SPDX-License-Identifier: Apache-2.0
+
+"""Tests for Amazon Bedrock Agent Runtime (invoke_agent and retrieve)."""
+
```
lmolkova, line 5: Please add conformance scenarios for `invoke_agent` and `retrieve`

instrumentation/opentelemetry-instrumentation-genai-bedrock/tests/test_agent.py
```diff
@@ -0,0 +1,731 @@
...
+    with stubber:
+        response = agent_client.invoke_agent(
+            agentId="agent-12345",
+            agentAliasId="alias-12345",
+            sessionId="session-123",
+            inputText="Hi",
+        )
+        with pytest.raises(ConnectionError):
+            list(response["completion"])
```
lmolkova, line 282: Please also add coverage for caller-side errors raised while consuming the stream to verify the span is closed with `error.type`.

```python
with pytest.raises(RuntimeError):
    for chunk in response["completion"]:
        raise RuntimeError("caller error")

assert spans[0].attributes.get(ErrorAttributes.ERROR_TYPE) == "RuntimeError"
```

instrumentation/opentelemetry-instrumentation-genai-bedrock/tests/test_agent.py
```diff
@@ -0,0 +1,731 @@
...
+        "ResourceNotFoundException",
+        "botocore.errorfactory.ResourceNotFoundException",
+    )
+
+
+# --- Async invoke_agent and retrieve tests ---
+
+
+@pytest.mark.asyncio
+async def test_async_invoke_agent_with_content(
```
lmolkova, line 569: Please drive async tests through the instrumented client instead of calling `_make_aio_api_call_wrapper` directly, and use the shared `instrument()` context manager instead of `monkeypatch`.

```python
with instrument(
    BedrockInstrumentor(),
    tracer_provider=tracer_provider,
    content_capture="SPAN_ONLY",
):
    ...
```

instrumentation/opentelemetry-instrumentation-genai-bedrock/src/opentelemetry/instrumentation/genai/bedrock/patch.py
```diff
@@ -357,6 +367,184 @@ async def _handle_async_invoke_model(
...
+        agent_id=agent_id,
+        agent_version=agent_version,
+        server_address=server_address,
+        server_port=server_port,
+    )
+    capture_content = handler.should_capture_content()
+    extract_invoke_agent_request(
+        api_params, invocation, capture_content=capture_content
+    )
+    return invocation, capture_content
```
lmolkova: should_capture_content is now available on the invocation, no need to return tuple

instrumentation/opentelemetry-instrumentation-genai-bedrock/src/opentelemetry/instrumentation/genai/bedrock/patch.py
```diff
@@ -357,6 +367,184 @@ async def _handle_async_invoke_model(
...
+    server_address, server_port = _server_address_and_port(instance)
+    raw_agent_id = api_params.get("agentId")
+    agent_id = str(raw_agent_id) if raw_agent_id else None
+    raw_alias_id = api_params.get("agentAliasId")
+    agent_version = str(raw_alias_id) if raw_alias_id else None
+
+    invocation = handler.invoke_remote_agent(
+        provider=GenAiProviderNameValues.AWS_BEDROCK.value,
+        agent_id=agent_id,
+        agent_version=agent_version,
```
lmolkova: these two friends are not sampling-relevant https://github.com/open-telemetry/semantic-conventions-genai/blob/c88d504ab3d9879f8e50d3cc87e69775e11db234/docs/gen-ai/gen-ai-agent-spans.md?plain=1#L450-L455, removing them in https://github.com/open-telemetry/opentelemetry-python-genai/pull/729  

Agent name on the other side is important and sampling-relevant. Is it available?

instrumentation/opentelemetry-instrumentation-genai-bedrock/src/opentelemetry/instrumentation/genai/bedrock/patch.py
DylanRussell: Ack removed

instrumentation/opentelemetry-instrumentation-genai-bedrock/src/opentelemetry/instrumentation/genai/bedrock/patch.py
DylanRussell: no there's no agent name in the request / response for this method.. we'd have to make a separate api call to get it

DylanRussell: i moved those to after the invocation start

instrumentation/opentelemetry-instrumentation-genai-bedrock/tests/test_agent.py
DylanRussell: This ended up catching an issue that a caller-side raise or break from inside a bare for loop never ended the span. This is because when that happens none of the finalization methods (__exit__, close) are hit..  

To fix this we added a method `_StreamTelemetry__del__` which gets invoked when the object is garbage collected.. Note that when the object is garbage collected is dependent on the users code, and also the exception is lost (if there was an exception, a break also causes the issue), it doesn't reach the wrapper

instrumentation/opentelemetry-instrumentation-genai-bedrock/src/opentelemetry/instrumentation/genai/bedrock/extractors.py
```diff
@@ -929,3 +931,114 @@ def extract_embedding_response(
...
+    capture_content: bool = True,
+) -> None:
+    if not capture_content:
+        return
+
+    results = response.get("retrievalResults")
+    if not _is_list(results):
+        return
+
+    docs: list[dict[str, Any]] = []
```
lmolkova: we should have a model for retrieved doc in util and I think it does not have content or metadata https://github.com/open-telemetry/semantic-conventions-genai/blob/main/model/gen-ai/gen-ai-retrieval-documents.json

instrumentation/opentelemetry-instrumentation-genai-bedrock/src/opentelemetry/instrumentation/genai/bedrock/stream.py
```diff
@@ -684,3 +685,80 @@ def __del__(self) -> None:
...
+        invocation: RemoteAgentInvocation,
+        *,
+        capture_content: bool = True,
+    ) -> None:
+        self._self_invocation = invocation
+        self._self_capture_content = capture_content
+        self._self_accumulated_text = []
+
+    def _process_chunk(self, chunk: dict[str, Any]) -> None:
+        if not _is_dict(chunk):
```
lmolkova: let's optimize and not buffer content if _self_capture_content is false?

util/opentelemetry-util-genai/src/opentelemetry/util/genai/stream.py
```diff
@@ -80,6 +80,27 @@ def _finalize_failure(self, error: BaseException) -> None:
...
+        hands the wrapper no other opportunity to end the span, so it would
+        otherwise leak. The caller's own exception is not recorded: it never
+        reaches the wrapper, and it is not the stream's failure.
+        """
+        # __del__ can run during interpreter shutdown or on an object whose
+        # __init__ raised, so nothing here may assume state exists.
+        if getattr(self, "_self_finalized", True):
+            return
+        try:
+            self._finalize_success()
```
lmolkova: maybe we should not consider it to be a success? it'd have random duration and would skew distribution a lot. Maybe we call it error.type = _OTHER (special constant that means we don't know but it's an error). We can also set span status description to something like "abandoned stream".

BTW is it guaranteed to be called in python during GC? In other languages I worked with GC might or might not invoke finalizers / destructors

DylanRussell: Not guaranteed: https://docs.python.org/3/reference/datamodel.html#object.__del__ but it seems like our best option.

Yeah setting it to error sounds good, made that change here, also inlined all this code under __del__..

instrumentation/opentelemetry-instrumentation-genai-bedrock/src/opentelemetry/instrumentation/genai/bedrock/stream.py
DylanRussell: ack, done

instrumentation/opentelemetry-instrumentation-genai-bedrock/src/opentelemetry/instrumentation/genai/bedrock/extractors.py
DylanRussell: Ack added a model.. it only has score and id.. There some cleanup todo.. pasting this genai comment here so I don't forget:

Because of that, other retrieval instrumentations written previously pass raw dictionaries:

DSPy: [{"content": str(psg)} for psg in passages]
LlamaIndex: [{"id": node.node_id, "content": node.get_content(), "score": ...}]
LangChain: [{"content": ..., "id": ..., "score": ...}]

lmolkova (approved):

instrumentation/opentelemetry-instrumentation-genai-bedrock/src/opentelemetry/instrumentation/genai/bedrock/extractors.py
lmolkova: thanks! created #743 to track, so we can merge this one once CI passes.

lmolkova:
@DylanRussell seems I'm merging into a wrong branch 🤐

PR #726, opened 2026-09-15
[`opentelemetry-instrumentation-genai-openai-agents`] Propagate `gen_ai.conversation.id` to nested spans

### LINKED ISSUE #725: openai-agents: run context does not propagate to spans owned by the model SDK instrumentation
### Describe your environment

OS: Amazon Linux 2023
Python version: Python 3.14.7
Package version: opentelemetry-instrumentation-genai-openai-agents 1.2b0.dev0, opentelemetry-instrumentation-genai-openai 1.2b0.dev0, opentelemetry-util-genai 1.2b0.dev0 (reproduces on main @ c6b63bd)
GenAI library (e.g. anthropic, openai) and version: openai-agents 0.22.2, openai 3.13.0

### What happened?

A trace of an agents run is produced by two independent instrumentation
packages. `opentelemetry-instrumentation-genai-openai-agents` owns
`invoke_workflow`, `invoke_agent` and `execute_tool`; the `chat` span is owned by
whichever model SDK instrumentation is installed —
`opentelemetry-instrumentation-genai-openai` on the default path, or
`...-anthropic` and friends when the run is routed through LiteLLM, any-llm, or
a caller-supplied `Model`.

There is no channel between them. Anything the agents instrumentation learns
from the framework stops at the spans it owns, and the model SDK
instrumentation, which only sees an HTTP call to a provider, cannot learn it.
OTel's Span API is write-only, so it cannot read the value off its parent span
either.

`gen_ai.conversation.id` is the case where this bites today. The agents library
carries a chat-thread identifier on `Trace.group_id`, set through
`RunConfig(group_id=...)` or `trace(..., group_id=...)`, and semconv defines the
attribute as conditionally required on inference, invoke_agent and
invoke_workflow. The agents instrumentation receives the `Trace` in
`on_trace_start` and ignores `group_id`, so no span carries the attribute today
— but only the `invoke_workflow` and `invoke_agent` halves are a local
oversight. The `chat` span is unreachable by construction, and that is the part
that needs a decision.

### Steps to Reproduce

```python
import asyncio

from agents import Agent, Runner
from agents.run import RunConfig

from opentelemetry.instrumentation.genai.openai import OpenAIInstrumentor
from opentelemetry.instrumentation.genai.openai_agents import (
    OpenAIAgentsInstrumentor,
)
from opentelemetry.sdk.trace import TracerProvider
from opentelemetry.sdk.trace.export import SimpleSpanProcessor
from opentelemetry.sdk.trace.export.in_memory_span_exporter import (
    InMemorySpanExporter,
)

exporter = InMemorySpanExporter()
provider = TracerProvider()
provider.add_span_processor(SimpleSpanProcessor(exporter))
OpenAIInstrumentor().instrument(tracer_provider=provider)
OpenAIAgentsInstrumentor().instrument(tracer_provider=provider)

agent = Agent(name="triage", instructions="Answer briefly.", model="gpt-4o-mini")
asyncio.run(
    Runner.run(
        agent,
        "What colour is the sky?",
        run_config=RunConfig(
            workflow_name="probe-workflow", group_id="chat-thread-1"
        ),
    )
)

for span in exporter.get_finished_spans():
    print(span.name, span.attributes.get("gen_ai.conversation.id"))
```

### Expected Result

```
chat gpt-4o-mini chat-thread-1
invoke_agent triage chat-thread-1
invoke_workflow probe-workflow chat-thread-1
```

### Actual Result

```
chat gpt-4o-mini None
invoke_agent triage None
invoke_workflow probe-workflow None
```

### Additional context

LangChain hit the same symptom in #474 (issue #475), but not the same problem:
LangChain is a single package that owns its whole tree and merges a run's config
metadata into every descendant, so each callback resolves the id from its own
metadata. Here the value has to cross a package boundary. The agents library
offers nothing to cross it with — `group_id` lives only on the `Trace`, and none
of the 14 types in `agents.tracing.span_data` exposes it.

Yes

### Tip

sfc-gh-zeningchen:
# Description

`opentelemetry-util-genai` gains a `conversation_context` module that holds
`gen_ai.conversation.id` on the OTel Context. Inference, agent and workflow
invocations read it at finish time when the caller set none explicitly, so a
`chat` span owned by a different instrumentation package picks it up. The
agents processor hands `Trace.group_id` over as the workflow invocation's
starting context.

`GenAIInvocation._start()` now uses the caller-supplied context as the base of
the invocation's own context rather than only to parent the span, and
`handler.workflow()` accepts one like `handler.tool()` does.

Fixes #725

## Type of change

# How has this been tested?

New unit tests in `util-genai` cover the fallback on all three invocation
types, explicit-over-ambient precedence, scope isolation, and that the id stays
off metric attributes. New `openai-agents` processor tests cover the `group_id`
hand-off and assert the processor writes no attributes itself.

Also driven end to end against a local Responses stand-in with both
instrumentations installed, asserting the `chat` span's instrumentation scope
really is `...genai.openai`; and against a run whose model calls Anthropic with
the openai instrumentation absent, confirming the fallback is not
OpenAI-specific.

- [x] `py3{10,14}-test-util-genai`: 432 passed
- [x] `py3{10,14}-test-instrumentation-genai-openai_agents-{oldest,latest}`: 35 passed
- [x] `py3{10,14}-test-instrumentation-genai-llama_index-{oldest,latest}`: 61 / 58 passed (the other `context=` caller)
- [x] `langchain` 395, `openai` 330, `smolagents` 125 passed
- [x] `tox -e typecheck` clean; ruff, ruff-format, rstcheck clean

# Checklist

- [x] Unit tests added
- [ ] Documentation updated


util/opentelemetry-util-genai/src/opentelemetry/util/genai/_invocation.py
```diff
@@ -147,10 +150,17 @@ def _start(
...
+        self._span_context = set_span_in_context(self.span, context)
         self._monotonic_start_s = timeit.default_timer()
         self._context_token = attach(self._span_context)
 
+    def _resolve_conversation_id(
+        self, conversation_id: str | None
+    ) -> str | None:
+        if conversation_id is not None:
+            return conversation_id
+        return get_ambient_conversation_id(self._span_context)
```
lmolkova: Resolving ambient `conversation_id` only at finish time leaves `self.conversation_id` unset on the invocation, and setting `invocation.conversation_id` directly does not propagate to nested spans.

```python
    def _start(
        self,
        attributes: dict[str, AttributeValue] | None = None,
        context: Context | None = None,
        conversation_id: str | None = None,
    ) -> None:
        self.conversation_id = conversation_id or get_ambient_conversation_id(context)
        if self.conversation_id:
            context = with_conversation_id(self.conversation_id, context=context)
        self.span = self._tracer.start_span(
            name=self._span_name,
            kind=self._span_kind,
            attributes=attributes,
            context=context,
        )
        self._span_context = set_span_in_context(self.span, context)
        self._monotonic_start_s = timeit.default_timer()
        self._context_token = attach(self._span_context)
```

instrumentation/opentelemetry-instrumentation-genai-openai-agents/src/opentelemetry/instrumentation/genai/openai_agents/processor.py
```diff
@@ -92,7 +95,18 @@ def on_trace_start(self, trace: Trace) -> None:
...
-        invocation = self._handler.workflow(name=trace.name)
+        #
+        # ``group_id`` is public on ``TraceImpl`` but not declared on the
+        # ``Trace`` ABC, hence the getattr.
+        group_id: str | None = getattr(trace, "group_id", None)
+        invocation = self._handler.workflow(
+            name=trace.name,
+            context=(
+                with_conversation_id(group_id)
+                if group_id is not None
```
lmolkova: Instrumentations should pass `conversation_id` through the handler abstraction rather than managing low-level context helpers directly.

```suggestion
        group_id: str | None = getattr(trace, "group_id", None)
        invocation = self._handler.workflow(
            name=trace.name,
            conversation_id=group_id,
        )
```

util/opentelemetry-util-genai/src/opentelemetry/util/genai/handler.py
```diff
@@ -600,6 +600,8 @@ def invoke_remote_agent(
     def workflow(
         self,
         name: str | None = None,
+        *,
+        context: Context | None = None,
```
lmolkova, line 625: `context` was added to `workflow()` but omitted from `invoke_local_agent()`, `invoke_remote_agent()`, and `inference()`, preventing agent-rooted traces from receiving ambient context.

```python
def invoke_local_agent(self, *, request_model: str | None = None, agent_name: str | None = None, context: Context | None = None) -> LocalAgentInvocation: ...
def invoke_remote_agent(self, provider: str, *, context: Context | None = None, ...) -> RemoteAgentInvocation: ...
def inference(self, provider: str, *, context: Context | None = None, ...) -> InferenceInvocation: ...
```

instrumentation/opentelemetry-instrumentation-genai-openai-agents/tests/test_processor.py
```diff
@@ -72,6 +86,54 @@ def test_trace_start_end_creates_and_stops_workflow() -> None:
...
+    processor.on_trace_start(trace)
+    span = _Span(AgentSpanData(name="triage"))
+    processor.on_span_start(span)
+    processor.on_span_end(span)
+    processor.on_trace_end(trace)
+
+    assert workflow_inv.conversation_id is None
+    assert agent_inv.conversation_id is None
+
+
```
lmolkova: This asserts that `conversation_id` is never assigned on invocation instances, codifying the missing attribute assignment.

```suggestion
    assert workflow_inv.conversation_id == "chat-thread-42"
    assert agent_inv.conversation_id == "chat-thread-42"
```

lmolkova (commented):
thanks! made a pass with AI, will take a proper look tomorrow. I think moving all context management to util would be great - it's hard to do right and prevent leaking. Would be great to do it for spans and extra attributes at the same time and in one place.

util/opentelemetry-util-genai/src/opentelemetry/util/genai/_invocation.py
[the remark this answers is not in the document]
sfc-gh-zeningchen: switched to go with a different approach

util/opentelemetry-util-genai/tests/test_conversation_context.py
```diff
@@ -0,0 +1,236 @@
...
+from opentelemetry.util.genai.conversation_context import (
+    get_ambient_conversation_id,
+    with_conversation_id,
+)
+from opentelemetry.util.genai.handler import TelemetryHandler
+
+
+class TestConversationContextPrimitives(unittest.TestCase):
+    def test_get_returns_none_when_nothing_attached(self):
+        assert get_ambient_conversation_id() is None
```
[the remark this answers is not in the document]
sfc-gh-zeningchen: updated

util/opentelemetry-util-genai/src/opentelemetry/util/genai/handler.py
```diff
@@ -600,6 +600,8 @@ def invoke_remote_agent(
...
     ) -> WorkflowInvocation:
         """Returns a Workflow invocation. Starts a span when called.
```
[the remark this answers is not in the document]
sfc-gh-zeningchen: done

instrumentation/opentelemetry-instrumentation-genai-openai-agents/tests/test_processor.py
sfc-gh-zeningchen: got it, added `handler.workflow.assert_called_once_with(name="Agent workflow", conversation_id="chat-thread-42")` to assert that since the assignment happens in `_start()` and workflow_inv here is a mock obj

util/opentelemetry-util-genai/src/opentelemetry/util/genai/handler.py
sfc-gh-zeningchen: thanks, added `context` and `conversation_id` to their signatures

instrumentation/opentelemetry-instrumentation-genai-openai-agents/src/opentelemetry/instrumentation/genai/openai_agents/processor.py
sfc-gh-zeningchen: thanks! updated

util/opentelemetry-util-genai/src/opentelemetry/util/genai/_invocation.py
sfc-gh-zeningchen: that's a cleaner solution, thanks! I've updated the code

instrumentation/opentelemetry-instrumentation-genai-openai/src/opentelemetry/instrumentation/genai/openai/response_extractors.py
```diff
@@ -662,7 +662,8 @@ def apply_request_attributes(
     invocation.attributes[OpenAIAttributes.OPENAI_API_TYPE] = (
         OpenAIAttributes.OpenaiApiTypeValues.RESPONSES.value
     )
-    invocation.conversation_id = params.conversation_id
+    if params.conversation_id is not None:
```
rads-1996, line 665: Since the invocation context is created in `handler.inference()`, setting the ID here does not propagate it to child spans. Could we include `conversation_id` in `get_inference_creation_kwargs()` instead?

lmolkova, line 665: +1 to Rad's point! Setting this after `handler.inference()` means `_start()` has already attached the context, so nested spans won't pick up the conversation id. Passing it via `get_inference_creation_kwargs()` would ensure it is attached from the start.

util/opentelemetry-util-genai/src/opentelemetry/util/genai/_invocation.py
```diff
@@ -134,20 +139,32 @@ def _start(
...
         """Start the invocation span and attach it to the current context.
 
         Args:
             attributes: Initial span attributes available for sampling decisions.
             context: An optional OpenTelemetry Context to parent the span.
+            conversation_id: Overrides the conversation id inherited from
+                ``context``. Either way the resolved value is put back on the
+                attached context, so nested invocations inherit it in turn.
         """
+        self.conversation_id = conversation_id or get_ambient_conversation_id(
```
lmolkova, line 154: Using `conversation_id or ...` relies on truthiness, so an explicit empty string falls back to the ambient ID. Could we check `if conversation_id is not None:` instead?

util/opentelemetry-util-genai/src/opentelemetry/util/genai/conversation_context.py
```diff
@@ -0,0 +1,42 @@
...
+from __future__ import annotations
+
+from opentelemetry.context import (
+    Context,
+    create_key,
+    get_value,
+    set_value,
+)
+
+CONVERSATION_ID_KEY = create_key("opentelemetry.util.genai.conversation_id")
```
lmolkova, line 21: Moving all context management into util is great - it's hard to do right and prevent leaking, and handling spans and extra attributes in one place will keep things clean.

Relatedly, Dylan has a PR for inner layer suppression and context passing (#663) that also touches context management in `_invocation.py`. Not a blocker, but worth taking a look when you have a moment so we can coordinate and ensure both context mechanisms work together reliably.

Also, since callers and instrumentations interact through `handler.*(conversation_id=...)`, could we keep this helper internal as `_conversation_context.py`?

instrumentation/opentelemetry-instrumentation-genai-openai/src/opentelemetry/instrumentation/genai/openai/response_extractors.py
sfc-gh-zeningchen: that's a good point, thanks! already updated this part

sfc-gh-zeningchen: thanks! already updated

util/opentelemetry-util-genai/src/opentelemetry/util/genai/_invocation.py
sfc-gh-zeningchen: just fixed this, thanks for calling it out!

util/opentelemetry-util-genai/src/opentelemetry/util/genai/_conversation_context.py
```diff
@@ -0,0 +1,42 @@
...
+from __future__ import annotations
+
+from opentelemetry.context import (
+    Context,
+    create_key,
+    get_value,
+    set_value,
+)
+
+CONVERSATION_ID_KEY = create_key("opentelemetry.util.genai.conversation_id")
```
sfc-gh-zeningchen: thanks! I've renamed it.
looking at #663 , it seems like it collects attributes from inner to outer instrumentation, this one hands the id in an opposite way (like head vs tail recursion), the code path should be pretty separate and could work together. I also noticed that it creates a `_context.py` file, happy to move what's in this file into that one once 663 is merged to make it cleaner!

lmolkova (approved):

PR #731, opened 2026-09-16
[`opentelemetry-instrumentation-genai-agno`] Add instrumentation for `continue.run` and `continue.arun`

DylanRussell:
# Description

Monkey patches `continue.run/arun` for teams, workflows and agents, we create a new agent / workflow invocation when these methods are called and use session id (resumption id) as conversation id.

Workflows, teams and agents all can be paused and then continued for human in the loop stuff / tool permission requests.  See https://docs.agno.com/reference-api/schema/agents/continue-agent-run for continuing agent runs..

## Type of change

# How has this been tested?

Unit tests

# Checklist

- [x] Unit tests added
- [x] Documentation updated

### DIFF
--- a/instrumentation/opentelemetry-instrumentation-genai-agno/src/opentelemetry/instrumentation/genai/agno/patch.py
+++ b/instrumentation/opentelemetry-instrumentation-genai-agno/src/opentelemetry/instrumentation/genai/agno/patch.py
@@ -81,6 +81,8 @@ _AGNO_KNOWLEDGE_MODULE = "agno.knowledge.knowledge"
 _KNOWLEDGE_CLASS = "Knowledge"
 
 
+# wrapt has no unregister API for post-import hooks; monotonic generations
+# invalidate deferred hooks registered during prior instrumentation cycles.
 _instrumentation_generation: int = 0
 _is_instrumented: bool = False
 
--- a/instrumentation/opentelemetry-instrumentation-genai-agno/src/opentelemetry/instrumentation/genai/agno/patch.py
+++ b/instrumentation/opentelemetry-instrumentation-genai-agno/src/opentelemetry/instrumentation/genai/agno/patch.py
@@ -3,40 +3,59 @@
 
 """Patching functions for Agno instrumentation."""
 
+# pylint: disable=import-outside-toplevel
+
 from __future__ import annotations
 
 import functools
 import logging
-from collections.abc import AsyncIterator, Awaitable, Callable, Iterator
+import sys
+from collections.abc import (
+    AsyncIterator,
+    Awaitable,
+    Callable,
+    Iterator,
+    Sequence,
+)
 from typing import TYPE_CHECKING, Any, cast
 
 if TYPE_CHECKING:
     from agno.agent import RunOutput
+    from agno.knowledge.document.base import Document
+    from agno.knowledge.knowledge import Knowledge
     from agno.run.workflow import WorkflowRunOutput
     from agno.team import TeamRunOutput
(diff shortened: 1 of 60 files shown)


instrumentation/opentelemetry-instrumentation-genai-agno/src/opentelemetry/instrumentation/genai/agno/patch.py
```diff
@@ -313,6 +359,166 @@ def _set_invocation_input(
...
+    for param in (
+        "input",
+        "additional_instructions",
+        "additionalInstructions",
+    ):
+        val = get_argument(param, wrapped, args, kwargs)
+        if val is not None:
+            return val
+    if args and isinstance(args[0], str):
+        return args[0]
```
lmolkova: If `continue_run` is called with a positional run ID string (e.g. `continue_run("run-123")`), `args[0]` gets recorded as a user input message; `get_argument` already handles positional mapping when `input` is in the signature.

```suggestion
```

instrumentation/opentelemetry-instrumentation-genai-agno/src/opentelemetry/instrumentation/genai/agno/patch.py
```diff
@@ -366,12 +574,26 @@ def _start_agent_invocation(
...
+    else:
+        _set_invocation_input(
+            invocation, instance, args, kwargs, capture_content, wrapped
+        )
+
+    tool_defs = prepare_tool_definitions(getattr(instance, "tools", None))
+    if not tool_defs:
+        tools_arg: Any = get_argument("tools", wrapped, args, kwargs)
+        if tools_arg is not None:
+            tool_defs = prepare_tool_definitions(tools_arg)
```
lmolkova: In `continue_run`, `tools` contains tool execution results rather than definitions, so fallback definition extraction should only run for regular invocations.

```suggestion
    if not tool_defs and not is_continue:
        tools_arg: Any = get_argument("tools", wrapped, args, kwargs)
        if tools_arg is not None:
            tool_defs = prepare_tool_definitions(tools_arg)
```

instrumentation/opentelemetry-instrumentation-genai-agno/src/opentelemetry/instrumentation/genai/agno/utils.py
```diff
@@ -141,12 +141,47 @@ def _extract_desc(tool: Any) -> str | None:
...
+        try:
+            parsed_coerced: Any = json.loads("".join(tools))
+            if isinstance(parsed_coerced, list):
+                raw_tools = cast(list[Any], parsed_coerced)
+            elif isinstance(parsed_coerced, dict):
+                raw_tools = [cast(dict[str, Any], parsed_coerced)]
+            else:
+                return None
+        except Exception:
+            return None
```
lmolkova: Joining 1-character string elements and JSON-parsing drops valid tools when tool names are single characters.

```suggestion
```

instrumentation/opentelemetry-instrumentation-genai-agno/src/opentelemetry/instrumentation/genai/agno/patch.py
DylanRussell: ack, applied suggestion

instrumentation/opentelemetry-instrumentation-genai-agno/src/opentelemetry/instrumentation/genai/agno/patch.py
DylanRussell: good point.. removed

PR #738, opened 2026-09-17
[`opentelemetry-instrumentation-genai-agno`] Add instrumentation for agno embedding methods

DylanRussell:
# Description

Monkey patch the built in `get_embedding` methods. Use a ContextVar (`_SUPPRESS_EMBEDDING`) to avoid double instrumenting in cases where one monkey patched embedding method delegates to another which can happen.

There's also code which monkey patches the users custom built Embeddors, both ones that are already initialized, and also ones that have yet to be initialized (we do this by monkey patching `__init__subclass__` hook to add monkey patching when the class is initialized...) -- this is a little excessive ! But I don't see anything wrong with the implementation. [Here](https://github.com/agno-agi/agno/blob/a71c4d0c607260d2483d73644a60dc49d036148c/libs/agno/agno/knowledge/embedder/base.py#L108) is a reference to the base Embedding class for reference..

## Type of change

# How has this been tested?

Unit tests

# Checklist

- [x] Unit tests added
- [x] Documentation updated


instrumentation/opentelemetry-instrumentation-genai-agno/src/opentelemetry/instrumentation/genai/agno/patch.py
```diff
@@ -91,6 +91,44 @@
 _ACTIVE_FOREGROUND_WORKFLOWS: contextvars.ContextVar[frozenset[int]] = (
     contextvars.ContextVar("_ACTIVE_FOREGROUND_WORKFLOWS", default=frozenset())
 )
+_SUPPRESS_EMBEDDING: contextvars.ContextVar[bool] = contextvars.ContextVar(
```
lmolkova, line 94: Let's wait for #663 to land for embeddings before we do this, similarly to #683 for inference. Otherwise, we risk emitting duplicate embedding spans when underlying client instrumentation is also present.

instrumentation/opentelemetry-instrumentation-genai-agno/src/opentelemetry/instrumentation/genai/agno/patch.py
```diff
@@ -253,6 +291,39 @@ def patch_agent(handler: TelemetryHandler) -> None:
...
+
+        # Wrap new embedder classes created during runtime
+        def _traced_init_subclass(cls: type[Embedder], **kwargs: Any) -> None:
+            super(Embedder, cls).__init_subclass__(**kwargs)
+            if _is_instrumented:
+                _wrap_embedder_class(cls, handler)
+
+        setattr(
+            Embedder, "__init_subclass__", classmethod(_traced_init_subclass)
+        )
```
lmolkova, line 323: Could we drop `__init_subclass__` and dynamic subclass patching, and stick to instrumenting the specific known classes in `_KNOWN_EMBEDDERS`? Intercepting `__init_subclass__` globally can fail if any subclass takes class kwargs (since `super(Embedder, cls).__init_subclass__(**kwargs)` forwards them to `object`). Leaving out custom user subclasses also keeps this aligned with how we handle other instrumentations in the repo.

instrumentation/opentelemetry-instrumentation-genai-agno/src/opentelemetry/instrumentation/genai/agno/patch.py
```diff
@@ -316,6 +408,276 @@ def _safe_unwrap(target: Any, attr: str) -> None:
...
+    if "agno.knowledge.embedder." in module:
+        sub = module.split("agno.knowledge.embedder.")[-1].split(".")[0]
+        if sub == "azure_openai":
+            return "azure.ai.openai"
+        if sub == "aws_bedrock":
+            return "aws.bedrock"
+        return sub.lower()
+    cls_name = embedder.__class__.__name__
+    cls_name = cls_name.removesuffix("Embedder")
+    return cls_name.lower() or "agno"
```
lmolkova, line 424: Please map known embedders to standard semconv provider values and fall back to `"unknown"` if unresolved (similarly to #683). We shouldn't fall back to `"agno"` or arbitrary class suffixes since Agno is the orchestrator, not a model provider.

instrumentation/opentelemetry-instrumentation-genai-agno/src/opentelemetry/instrumentation/genai/agno/patch.py
```diff
@@ -316,6 +408,276 @@ def _safe_unwrap(target: Any, attr: str) -> None:
...
+
+def _wrap_embedder_class(
+    cls: type[Any],
+    handler: TelemetryHandler,
+) -> None:
+    for method_name, wrapper_fn in _embedder_method_wrappers(handler):
+        if hasattr(cls, method_name):
+            target = getattr(cls, method_name)
+            if not hasattr(target, "__wrapped__"):
+                try:
```
lmolkova, line 672: If we do wrap subclasses, checking `method_name in cls.__dict__` instead of `hasattr(cls, method_name)` might be safer to avoid copying inherited base methods directly onto child classes and leaving them behind on unpatch.

instrumentation/opentelemetry-instrumentation-genai-agno/src/opentelemetry/instrumentation/genai/agno/patch.py
DylanRussell: good point that we have the duplicate embedding issue here too.. but this was added to resolve the issue that the embedding methods within `agno` itself can call eachother.. (ex `get_embedding` can call `get_embedding_and_usage`)... in this case i think it makes sense to disable the monkey patching altogether from the nested call when we detect it.. WDYT ?

instrumentation/opentelemetry-instrumentation-genai-agno/src/opentelemetry/instrumentation/genai/agno/patch.py
DylanRussell: i think we could resolve that issue but it's turning into a lot of code to add to patch something in sort of a hacky way.. ive removed it all

instrumentation/opentelemetry-instrumentation-genai-agno/src/opentelemetry/instrumentation/genai/agno/patch.py
DylanRussell: ack.. ended up removing this all

instrumentation/opentelemetry-instrumentation-genai-agno/src/opentelemetry/instrumentation/genai/agno/patch.py
DylanRussell: Ack.. added logic to resolve the provider..

PR #739, opened 2026-09-17
fix CI: declare companion bounds and exclude broken workflows 2.24.0

lmolkova:
Declares `llama-index-instrumentation` and `llama-index-workflows` in `instruments` with their supported lower bounds, excludes `llama-index-workflows==2.24.0`, and pins minor versions in `requirements.latest.txt`.

`llama-index-workflows==2.24.0` breaks workflow agent runs (`FunctionAgent`, `ReActAgent`), which fails even without instrumentation.


instrumentation/opentelemetry-instrumentation-genai-llama-index/tests/requirements.latest.txt
```diff
@@ -38,7 +38,9 @@
 # the newest supported version of external dependencies.
 
 llama-index-core~=0.14.24
-llama-index-llms-openai
+llama-index-workflows~=2.23.0
```
lmolkova, line 41: explicit dependencies so we won't fail on CI without explicit udpate

rads-1996: but this would fetch 2.24.0 still, right?

lmolkova: `~=2.23.0` fetches anything `2.23.*`,  `~=2.23` would fetch 2.24

also CI is green

rads-1996 (approved):

PR #740, opened 2026-09-17
Use `inspect.signature` and then `bind.partial` and then fetch argument instead of manually parsing args/kwargs

DylanRussell:
# Description

I think it's better to use `inspect.signature` and then `bind.partial`, and then fetch argument via it's real name, rather than doing manual inspection of args / kwargs.. It's possible an argument name changes but that would've broke the previous code too..

Some instrumentations don't need this because the library only uses kwargs (google-genai, anthropic, open ai).

Some instrumentations are using callbacks (langchain) and dont need to worry about args / kwargs at all.

## Type of change

# How has this been tested?

Unit tests

# Checklist

- [x] Unit tests added
- [x] Documentation updated


instrumentation/opentelemetry-instrumentation-google-genai/src/opentelemetry/instrumentation/google_genai/tool_call_wrapper.py
```diff
@@ -46,19 +47,10 @@ def _to_otel_value(python_value):
...
-            type(entry).__name__
-        )
-        function_arg_attr[f"code.function.parameters.{param_name}.value"] = (
-            _to_otel_value(entry)
-        )
-    for key, value in function_kwargs.items():
+    bound = bind_arguments(
+        wrapped_function, function_args, function_kwargs, apply_defaults=False
+    )
+    for key, value in bound.items():
```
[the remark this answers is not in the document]
DylanRussell: It does change the existing behavior a bit but i think it's an overall improvement to bind the arguments to the function first

util/opentelemetry-util-genai/src/opentelemetry/util/genai/utils.py
```diff
@@ -179,3 +183,66 @@ def default(self, o: Any) -> Any:
 )
 """Should be used by GenAI instrumentations when serializing objects that may contain
 bytes, datetimes, etc. for GenAI observability."""
+
+
+_SIGNATURE_CACHE_MAX_SIZE = 1024
+_signature_cache: dict[tuple[object, bool], inspect.Signature] = {}
```
eternalcuriouslearner, line 189: question(non-blocking): Can we make this dict thread safe by using lru_cache? Say something like

```python
import functools
import inspect
from collections.abc import Callable


@functools.lru_cache(maxsize=1024)
def _cached_signature(fn: Callable[..., object], drop_first: bool) -> inspect.Signature:
    sig = inspect.signature(fn)
    if drop_first:
        params = list(sig.parameters.values())[1:]
        sig = sig.replace(parameters=params)
    return sig


def _get_signature(func: Callable[..., object]) -> inspect.Signature:
    # Bound methods are new objects on every attribute access, so key on the
    # underlying function. Only long-lived definitions are cached; per-call
    # closures and callable instances would otherwise be pinned in the cache.
    underlying = getattr(func, "__func__", None)
    if underlying is not None and getattr(func, "__self__", None) is not None:
        return _cached_signature(underlying, True)
    if inspect.isfunction(func) and "<locals>" not in func.__qualname__:
        return _cached_signature(func, False)
    return inspect.signature(func)

File: util/opentelemetry-util-genai/src/opentelemetry/util/genai/utils.py [hunk 3]
```diff
@@ -179,3 +183,66 @@ def default(self, o: Any) -> Any:
...
+    except (TypeError, ValueError):
+        return dict(kwargs)
+
+
+def get_argument(
+    name: str,
+    func: Callable[..., object],
+    args: tuple[object, ...],
+    kwargs: Mapping[str, object],
+    default: Any = None,
```
eternalcuriouslearner, line 240: question(non-blocking): Can we avoid use of Any by doing something like

```python
def get_argument(name: str, func: Callable[..., object], args: tuple[object, ...],
                 kwargs: Mapping[str, object], default: object = None,
                 *, apply_defaults: bool = False) -> object: ...

File: instrumentation/opentelemetry-instrumentation-google-genai/src/opentelemetry/instrumentation/google_genai/tool_call_wrapper.py [hunk 4]
```diff
@@ -46,19 +47,10 @@ def _to_otel_value(python_value):
...
-        if index < len(params):
-            param_name = params[index].name
-        function_arg_attr[f"code.function.parameters.{param_name}.type"] = (
-            type(entry).__name__
-        )
-        function_arg_attr[f"code.function.parameters.{param_name}.value"] = (
-            _to_otel_value(entry)
-        )
-    for key, value in function_kwargs.items():
+    bound = bind_arguments(
```
eternalcuriouslearner, line 50: question(non-blocking): Is the telemetry shape change for variadic tool functions intentional?

For example:
 ```python
def weather(**kwargs):
         ...

 weather(city="Boston", units="celsius")
```

The existing implementation emits flat parameters:

```text
code.function.parameters.city.value
code.function.parameters.units.value ```

After this change, `bind_arguments()` produces:

```python
{"kwargs": {"city": "Boston", "units": "celsius"}}
```

so telemetry instead contains `code.function.parameters.kwargs.value`. Let me know if this understanding is right.

If yes, could we preserve the existing extraction for variadic signatures while using signature binding for ordinary parameters?

```python
signature = inspect.signature(wrapped_function)
parameters = list(signature.parameters.values())
 has_variadics = any(
     parameter.kind
     in (
         inspect.Parameter.VAR_POSITIONAL,
         inspect.Parameter.VAR_KEYWORD,
    )
    for parameter in parameters
 )

 if has_variadics:
     # Preserve the existing telemetry representation.
     bound = {
         (
            parameters[index].name
             if index < len(parameters)
             else f"args[{index}]"
         ): value
         for index, value in enumerate(function_args)
     }
     bound.update(function_kwargs)
 else:
    bound = bind_arguments(
         wrapped_function,
         function_args,
        function_kwargs,
        apply_defaults=False,
    )
 ```

Additionally can we add some regression tests for both a `*args` tool and a `**kwargs` tool so the chosen telemetry contract is explicit. Feel free to push back if this an overkill.

**Review by eternalcuriouslearner — APPROVED** [review]
LGTM!! leaving few non-blocking questions/suggestions. Feel free to push back.

File: util/opentelemetry-util-genai/src/opentelemetry/util/genai/utils.py (hunk 2 above)
Comment (DylanRussell) [reply util/opentelemetry-util-genai/src/opentelemetry/util/genai/utils.py]: yeah switched to use this..

File: util/opentelemetry-util-genai/src/opentelemetry/util/genai/utils.py (hunk 3 above)
Comment (DylanRussell) [reply util/opentelemetry-util-genai/src/opentelemetry/util/genai/utils.py]: sure.. is object much better than Any ? Anyway switched to use it

File: instrumentation/opentelemetry-instrumentation-google-genai/src/opentelemetry/instrumentation/google_genai/tool_call_wrapper.py (hunk 4 above)
Comment (DylanRussell) [reply instrumentation/opentelemetry-instrumentation-google-genai/src/opentelemetry/instrumentation/google_genai/tool_call_wrapper.py]: Switched to do it like you proposed and added a test.. I'd rather have this instrumentation just pass the bound arguments dict to utils where it's json serialized, that's more aligned with what the other instrumentations are doing.. Opened https://github.com/open-telemetry/opentelemetry-python-genai/issues/764 to track this

File: instrumentation/opentelemetry-instrumentation-genai-bedrock/src/opentelemetry/instrumentation/genai/bedrock/patch.py [hunk 5]
```diff
@@ -166,8 +167,8 @@ def _wrapper(
         if service_name != BEDROCK_RUNTIME:
             return wrapped(*args, **kwargs)
 
-        operation_name = args[0] if args else kwargs.get("operation_name")
-        raw_params = args[1] if len(args) > 1 else kwargs.get("api_params")
+        operation_name = get_argument("operation_name", wrapped, args, kwargs)
+        raw_params = get_argument("api_params", wrapped, args, kwargs)
```
lmolkova: Both arguments are typically passed positionally, so calling `get_argument` twice repeats `bind_partial` on every AWS call. We could bind once:

```suggestion
        bound = bind_arguments(wrapped, args, kwargs)
        operation_name = bound.get("operation_name")
        raw_params = bound.get("api_params")
```
(and import `bind_arguments` instead of `get_argument`).

instrumentation/opentelemetry-instrumentation-genai-qwen-agent/src/opentelemetry/instrumentation/genai/qwen_agent/patch.py
```diff
@@ -91,17 +92,18 @@ def wrap_agent_call_tool(
     handler: TelemetryHandler,
 ) -> Any:
     """Wrapper for ``Agent._call_tool()`` producing an ``execute_tool`` span."""
-    tool_name = str(args[0]) if args else str(kwargs.get("tool_name", ""))
-    tool_args = args[1] if len(args) > 1 else kwargs.get("tool_args")
+    tool_name = str(
+        get_argument("tool_name", wrapped, args, kwargs, default="")
+    )
+    tool_args: Any = get_argument("tool_args", wrapped, args, kwargs)
```
lmolkova: Since `tool_name`, `tool_args`, and `messages` are positional, calling `get_argument` repeats signature binding three times for each tool execution. We can bind once via `bind_arguments`:

```suggestion
    bound = bind_arguments(wrapped, args, kwargs)
    tool_name = str(bound.get("tool_name", ""))
    tool_args: Any = bound.get("tool_args")
```

util/opentelemetry-util-genai/src/opentelemetry/util/genai/utils.py
```diff
@@ -203,3 +207,73 @@ def default(self, o: Any) -> Any:
...
+
+def _get_signature(func: Callable[..., object]) -> inspect.Signature:
+    """Return the inspect.Signature for a callable, caching long-lived definitions.
+
+    Bound methods are new objects on every attribute access, so key on the
+    underlying function. Only long-lived definitions are cached; per-call
+    closures and callable instances would otherwise be pinned in the cache.
+    """
+    try:
+        underlying = getattr(func, "__func__", None)
```
lmolkova: Bound methods miss the `<locals>` check applied to functions on line 236, so methods on locally defined classes can still be retained in `_cached_signature`:

```suggestion
        underlying = getattr(func, "__func__", None)
        if (
            underlying is not None
            and getattr(func, "__self__", None) is not None
            and "<locals>" not in getattr(underlying, "__qualname__", "")
        ):
            return _cached_signature(underlying, True)
```

instrumentation/opentelemetry-instrumentation-google-genai/src/opentelemetry/instrumentation/google_genai/tool_call_wrapper.py
```diff
@@ -46,19 +47,40 @@ def _to_otel_value(python_value):
...
-    signature = inspect.signature(wrapped_function)
-    params = list(signature.parameters.values())
-    for index, entry in enumerate(function_args):
-        param_name = f"args[{index}]"
-        if index < len(params):
-            param_name = params[index].name
-        function_arg_attr[f"code.function.parameters.{param_name}.type"] = (
-            type(entry).__name__
+    try:
+        signature = inspect.signature(wrapped_function)
```
lmolkova: `inspect.signature` is called uncached here on every tool invocation, and then inspected again inside `bind_arguments` when `has_variadics` is False. If #764 is tracking standardizing this wrapper, should we reuse signature caching or note this as part of that follow-up?

instrumentation/opentelemetry-instrumentation-genai-bedrock/src/opentelemetry/instrumentation/genai/bedrock/patch.py
lmolkova: Both arguments are typically passed positionally, so calling `get_argument` twice repeats `bind_partial` on every AWS call. We could bind once:

```suggestion
        bound = bind_arguments(wrapped, args, kwargs)
        operation_name = bound.get("operation_name")
        raw_params = bound.get("api_params")
```
(and import `bind_arguments` instead of `get_argument`).

instrumentation/opentelemetry-instrumentation-genai-qwen-agent/src/opentelemetry/instrumentation/genai/qwen_agent/patch.py
lmolkova: Since `tool_name`, `tool_args`, and `messages` are positional, calling `get_argument` repeats signature binding three times for each tool execution. We can bind once via `bind_arguments`:

```suggestion
    bound = bind_arguments(wrapped, args, kwargs)
    tool_name = str(bound.get("tool_name", ""))
    tool_args: Any = bound.get("tool_args")
```

instrumentation/opentelemetry-instrumentation-genai-dspy/src/opentelemetry/instrumentation/genai/dspy/patch.py
```diff
@@ -246,13 +246,7 @@ def _extract_tool_arguments(
...
     func: Any = getattr(instance, "func", None)
     if func is not None and callable(func):
-        try:
-            sig = inspect.signature(func)
-            bound = sig.bind_partial(*args, **kwargs)
-            bound.apply_defaults()
-            return dict(bound.arguments)
-        except (TypeError, ValueError):
-            pass
+        return bind_arguments(func, args, kwargs, apply_defaults=True)
```
lmolkova: When `bind_arguments` fails on positional arguments (e.g. uninspectable built-ins like `dir` or mismatched argument counts), it returns `dict(kwargs)` which is `{}`. This causes telemetry to record `gen_ai.tool.call.arguments = "{}"` instead of omitting the attribute (`None`).

Could we preserve returning `None` when positional arguments are present but binding fails?

```suggestion
        bound = bind_arguments(func, args, kwargs, apply_defaults=True)
        if args and bound == kwargs:
            return None
        return bound
```

util/opentelemetry-util-genai/src/opentelemetry/util/genai/utils.py
lmolkova, line 235: Bound methods miss the `<locals>` check applied to functions on line 236, so methods on locally defined classes can still be retained in `_cached_signature`:

```suggestion
        underlying = getattr(func, "__func__", None)
        if (
            underlying is not None
            and getattr(func, "__self__", None) is not None
            and "<locals>" not in getattr(underlying, "__qualname__", "")
        ):
            return _cached_signature(underlying, True)
```

instrumentation/opentelemetry-instrumentation-google-genai/src/opentelemetry/instrumentation/google_genai/tool_call_wrapper.py
lmolkova: `inspect.signature` is called uncached here on every tool invocation, and then inspected again inside `bind_arguments` when `has_variadics` is False. If #764 is tracking standardizing this wrapper, should we reuse signature caching or note this as part of that follow-up?

DylanRussell: May as well use the signature caching.. updated it

util/opentelemetry-util-genai/src/opentelemetry/util/genai/utils.py
DylanRussell: Added suggestion

util/opentelemetry-util-genai/src/opentelemetry/util/genai/utils.py
```diff
@@ -203,3 +207,77 @@ def default(self, o: Any) -> Any:
...
+    func: Callable[..., object],
+    args: tuple[object, ...],
+    kwargs: Mapping[str, object],
+    default: object = None,
+    *,
+    apply_defaults: bool = False,
+) -> object:
+    """Extract a named argument from kwargs or args via signature binding."""
+    if name in kwargs:
+        return kwargs[name]
```
lmolkova, line 278:
```suggestion
    if name in kwargs:
        return kwargs[name]
    if not args and not apply_defaults:
        return default
```
When `name` is not in `kwargs` and no positional arguments or defaults are involved, signature binding cannot introduce `name`. Returning early avoids unnecessary signature binding on every call with omitted optional kwargs.

util/opentelemetry-util-genai/src/opentelemetry/util/genai/utils.py
```diff
@@ -203,3 +207,77 @@ def default(self, o: Any) -> Any:
...
+        ):
+            return _cached_signature(underlying, True)
+        if inspect.isfunction(func) and "<locals>" not in func.__qualname__:
+            return _cached_signature(func, False)
+    except TypeError:
+        pass
+    return _inspect_signature(func)
+
+
+_get_signature = get_signature
```
lmolkova:
```suggestion
```
`_get_signature` alias is redundant now that `get_signature` is public; we can remove it and update the call in `test_get_signature_unhashable_callable`.

instrumentation/opentelemetry-instrumentation-genai-dspy/src/opentelemetry/instrumentation/genai/dspy/patch.py
```diff
@@ -465,8 +460,8 @@ def _start_retrieval_invocation(
         else None,
     )
 
-    invocation.query_text = _extract_retrieval_query(args, kwargs)
-    invocation.top_k = _extract_retrieval_k(instance, args, kwargs)
+    invocation.query_text = _extract_retrieval_query(wrapped, args, kwargs)
+    invocation.top_k = _extract_retrieval_k(instance, wrapped, args, kwargs)
```
lmolkova: Both `_extract_retrieval_query` and `_extract_retrieval_k` invoke `get_argument` with the same `(wrapped, args, kwargs)`. When arguments are passed positionally, this binds the signature twice. Similar to Bedrock and Qwen-Agent, binding once avoids repeating signature binding twice for every retrieval call.

lmolkova (approved):

util/opentelemetry-util-genai/src/opentelemetry/util/genai/utils.py
DylanRussell: Ack makes sense.. applied suggestion

util/opentelemetry-util-genai/src/opentelemetry/util/genai/utils.py
DylanRussell: Ack, removed

instrumentation/opentelemetry-instrumentation-genai-dspy/src/opentelemetry/instrumentation/genai/dspy/patch.py
DylanRussell: Ack fixed

lmolkova:
@DylanRussell did you push the change ?

DylanRussell:
Woops forgot to push. Now it's pushed

PR #741, opened 2026-09-18
Minor utils cleanup: remove `start` and `get_start_attributes` methods, inline the code inside `__init__`

DylanRussell:
# Description

Remove the `_start` method from `GenAIInvocation` and `_get_start_attributes` from its subclasses, inlining span initialization into `__init__`. Make `start_attributes` a required public attribute on the base class initialized via `super().__init__`.

Just simplifies the code a bit.

## Type of change

# How has this been tested?

Unit tests

# Checklist

- [x] Unit tests added
- [x] Documentation updated


util/opentelemetry-util-genai/src/opentelemetry/util/genai/_invocation.py
```diff
@@ -95,12 +96,17 @@ def __init__(
             {} if metric_attributes is None else metric_attributes
         )
         """Additional attributes to set on metrics. Must be low cardinality. Not set on spans or events."""
-        self.span: Span = _INVALID_SPAN
-        self._span_context: Context
-        self._span_name: str = span_name
-        self._span_kind: SpanKind = span_kind
-        self._context_token: ContextToken | None = None
-        self._monotonic_start_s: float
+        self.start_attributes: dict[str, AttributeValue] = start_attributes
```
lmolkova, line 99: Could we keep `_start_attributes` private instead of exposing it publicly? Unlike `invocation.attributes` (which callers mutate before finishing), modifying `start_attributes` after `__init__` won't update the span. Keeping it internal avoids expanding the public API surface.

util/opentelemetry-util-genai/src/opentelemetry/util/genai/_inference_invocation.py
```diff
@@ -100,7 +100,17 @@ def __init__(
         operation_name = (
             operation_name or GenAI.GenAiOperationNameValues.CHAT.value
         )
-        """Use handler.inference(provider) rather than calling this directly."""
+        start_attributes: dict[str, AttributeValue] = {
+            k: v
```
lmolkova, line 104: Since `operation_name` is already passed to `super().__init__`, what do you think about having `GenAIInvocation.__init__` populate `GenAI.GEN_AI_OPERATION_NAME` automatically? That would save repeating it across all subclasses.

util/opentelemetry-util-genai/src/opentelemetry/util/genai/_inference_invocation.py
```diff
@@ -387,7 +385,7 @@ def _get_metric_attributes(self) -> dict[str, AttributeValue]:
         # Cached because this is rebuilt once per streaming chunk. Any mutation
         # of its inputs must call _invalidate_metric_attributes.
```
lmolkova, line 386: `self._provider`, `self._request_model`, etc. are stored in `__init__` above but no longer used now that this copies `start_attributes`. Could we build metric attributes from the instance fields (matching `EmbeddingInvocation` and `RetrievalInvocation`), or clean up the unused fields?

util/opentelemetry-util-genai/.changelog/741.changed
```diff
@@ -0,0 +1 @@
+Inline invocation span initialization into `GenAIInvocation.__init__` and expose `start_attributes` as a public attribute.
```
lmolkova, line 1: If `start_attributes` remains internal, this refactor doesn't introduce public API changes and we can remove this changelog fragment.

lmolkova (commented):
nice cleanup!! A few suggestions

util/opentelemetry-util-genai/src/opentelemetry/util/genai/_invocation.py
DylanRussell: yeah good idea. done

util/opentelemetry-util-genai/src/opentelemetry/util/genai/_inference_invocation.py
DylanRussell: yep true, done

util/opentelemetry-util-genai/src/opentelemetry/util/genai/_inference_invocation.py
DylanRussell: yep, done

util/opentelemetry-util-genai/.changelog/741.changed
DylanRussell: nice, done

lmolkova (approved):

PR #750, opened 2026-09-20
util-genai: wrap completion hooks in safe completion hook to suppress exceptions

lmolkova:
Wraps programmatic completion hooks in `_SafeCompletionHook` within `TelemetryHandler` so hook exceptions are suppressed across all hook configurations.

Also updates the smolagents completion hook failure test to verify suppression instead of expecting exceptions to escape to the caller.


lzchen (approved):

rads-1996 (approved):
LGTM.

PR #752, opened 2026-09-21
feat(google-genai): capture Interactions request configuration

### LINKED ISSUE #563: [google-genai] Capture generation_config request parameters on interactions.create
> [!NOTE]

Part of #560.

Capture request configuration parameters from `generation_config` when creating an interaction.

- Attributes: `gen_ai.request.temperature`, `gen_ai.request.top_p`, `gen_ai.request.max_tokens`, `gen_ai.request.seed`, `gen_ai.request.stop_sequences`, `gen_ai.output.type` (from `response_format` or `response_mime_type`)

JacksonWeber:
# Description

Captures explicitly supplied generation configuration and output-format attributes for Google GenAI Interactions calls, including sync, async, streaming, and failed requests.

These settings were previously omitted from telemetry. The implementation supports legacy keyword arguments and newer SDK `request.body` models, preserves zero values, and leaves SDK arguments and lazy iterators untouched.

Closes #563

Field availability follows the SDK version. Mixed output formats are omitted when no single semantic-convention output type represents them.

Assisted-by: GPT-6 Astra

## Type of change

- [x] This change requires a documentation update

# How has this been tested?

- [x] `tox -e precommit,typecheck`
- [x] Google oldest, v2, and latest suites, including Python 3.14 latest.
- [x] Google conformance scenarios with Weaver 0.26.1 and VCR replay.
- [x] Offline SDK-model and HTTP-transport coverage for parameters, streaming normalization, and error paths. No live Gemini calls or API key.

# Checklist

- [x] Unit tests added
- [x] Documentation updated


instrumentation/opentelemetry-instrumentation-google-genai/src/opentelemetry/instrumentation/google_genai/interactions.py
```diff
@@ -512,10 +517,114 @@ def _maybe_get_tool_definitions(
     return definitions if definitions else None
 
 
+def _interaction_request(kwargs: dict[str, Any]) -> object:
+    body = _get_field(kwargs.get("request"), "body")
+    if body is None:
+        return kwargs
+    # Read stored model fields without triggering SDK deprecation warnings.
+    return getattr(body, "__dict__", body)
+
```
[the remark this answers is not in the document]
JacksonWeber: Changed it to record only settings the caller explicitly supplied, not SDK defaults. It reads field metadata without serializing the request, so lazy content stays untouched.

eternalcuriouslearner (approved):

instrumentation/opentelemetry-instrumentation-google-genai/README.rst
```diff
@@ -115,6 +115,30 @@ captured using the following environment variables:
 If both variables are set, the includes list is applied first, then the
 excludes list filters the result further.
 
+Interactions request parameters
```
lmolkova: Can we de-AI this section a bit? I think it can be either removed (we don't list the details for each operation) or collapsed to 1-2 sentences.

instrumentation/opentelemetry-instrumentation-google-genai/src/opentelemetry/instrumentation/google_genai/interactions.py
```diff
@@ -512,10 +517,124 @@ def _maybe_get_tool_definitions(
...
+def _output_type_from_mime_type(mime_type: object) -> str | None:
+    if not isinstance(mime_type, str):
+        return None
+    mime_type = mime_type.partition(";")[0].strip().lower()
+    output_types = GenAIAttributes.GenAiOutputTypeValues
+    if mime_type == "application/json" or mime_type.endswith("+json"):
+        return output_types.JSON.value
+    if mime_type == "text/plain":
+        return output_types.TEXT.value
```
lmolkova:
```suggestion
    if mime_type.startswith("text/"):
        return output_types.TEXT.value
```
Prefix matching here (like `image/` and `audio/`) would also catch types like `text/markdown` or `text/x.enum`.

instrumentation/opentelemetry-instrumentation-google-genai/src/opentelemetry/instrumentation/google_genai/interactions.py
```diff
@@ -512,10 +517,124 @@ def _maybe_get_tool_definitions(
...
+    if format_type in (
+        "object",
+        "array",
+        "string",
+        "number",
+        "integer",
+        "boolean",
+        "null",
+    ):
+        return output_types.JSON.value
```
lmolkova, line 609:
```suggestion
    if format_type in (
        "object",
        "array",
        "string",
        "number",
        "integer",
        "boolean",
        "null",
        "json",
        "json_object",
        "json_schema",
    ):
        return output_types.JSON.value
```
Might be worth including common JSON format names too.

instrumentation/opentelemetry-instrumentation-google-genai/src/opentelemetry/instrumentation/google_genai/interactions.py
```diff
@@ -512,10 +517,124 @@ def _maybe_get_tool_definitions(
...
+) -> None:
+    config = _explicit_request_fields(_get_field(request, "generation_config"))
+    for name in ("temperature", "top_p"):
+        value = _get_field(config, name)
+        if (
+            isinstance(value, (int, float))
+            and not isinstance(value, bool)
+            and -float_info.max <= value <= float_info.max
+        ):
+            setattr(invocation, name, float(value))
```
lmolkova:
```suggestion
    for name in ("temperature", "top_p"):
        value = _get_field(config, name)
        if (
            isinstance(value, (int, float))
            and not isinstance(value, bool)
            and math.isfinite(value)
            and value >= 0
        ):
            setattr(invocation, name, float(value))
```
`math.isfinite` could be simpler than checking against `float_info`.

instrumentation/opentelemetry-instrumentation-google-genai/src/opentelemetry/instrumentation/google_genai/interactions.py
```diff
@@ -512,10 +517,124 @@ def _maybe_get_tool_definitions(
...
+    if isinstance(max_tokens, int) and not isinstance(max_tokens, bool):
+        invocation.max_tokens = max_tokens
+    seed = _get_field(config, "seed")
+    if isinstance(seed, int) and not isinstance(seed, bool):
+        invocation.seed = seed
+    stop_sequences = _get_field(config, "stop_sequences")
+    if isinstance(stop_sequences, (list, tuple)) and all(
+        isinstance(item, str) for item in stop_sequences
+    ):
+        invocation.stop_sequences = list(stop_sequences)
```
lmolkova:
```suggestion
    if (
        isinstance(stop_sequences, (list, tuple))
        and stop_sequences
        and all(isinstance(item, str) for item in stop_sequences)
    ):
        invocation.stop_sequences = list(stop_sequences)
```
Consider checking that `stop_sequences` is non-empty so `[]` is not recorded as an attribute.

eternalcuriouslearner:
Sorry overlooked one small thing, should we add assertions to verify these attributes to vcr tests and conformance test suite?

instrumentation/opentelemetry-instrumentation-google-genai/README.rst
JacksonWeber: Makes sense, removed this section to clean this up

instrumentation/opentelemetry-instrumentation-google-genai/src/opentelemetry/instrumentation/google_genai/interactions.py
JacksonWeber: Added `json`, `json_object`, and `json_schema` here and updated the test request params

instrumentation/opentelemetry-instrumentation-google-genai/src/opentelemetry/instrumentation/google_genai/interactions.py
JacksonWeber: Updated to use `math.isfinite`

instrumentation/opentelemetry-instrumentation-google-genai/src/opentelemetry/instrumentation/google_genai/interactions.py
JacksonWeber: Added a check on `stop_sequences`.

instrumentation/opentelemetry-instrumentation-google-genai/src/opentelemetry/instrumentation/google_genai/interactions.py
```diff
@@ -512,10 +518,132 @@ def _maybe_get_tool_definitions(
...
+
+def _apply_interaction_request_attributes(
+    invocation: InferenceInvocation | RemoteAgentInvocation,
+    request: object,
+) -> None:
+    config = _explicit_request_fields(_get_field(request, "generation_config"))
+    for name in ("temperature", "top_p"):
+        value = _get_field(config, name)
+        if isinstance(value, int) and value > float_info.max:
+            continue
```
lmolkova: btw why do we even need to protect from it, if top_p is invalid float, the SDK would throw, why do we need to validate boundaries?

instrumentation/opentelemetry-instrumentation-google-genai/src/opentelemetry/instrumentation/google_genai/interactions.py
```diff
@@ -512,10 +518,132 @@ def _maybe_get_tool_definitions(
...
+        return output_types.JSON.value
+    return None
+
+
+def _apply_interaction_request_attributes(
+    invocation: InferenceInvocation | RemoteAgentInvocation,
+    request: object,
+) -> None:
+    config = _explicit_request_fields(_get_field(request, "generation_config"))
+    for name in ("temperature", "top_p"):
```
lmolkova: why do we have a loop here? please copy over method _coerse_float from some other place in this repo and call it for individual properties.

And the let's use typed invocatio.top_p | temperature setters.

instrumentation/opentelemetry-instrumentation-google-genai/src/opentelemetry/instrumentation/google_genai/interactions.py
```diff
@@ -512,10 +518,132 @@ def _maybe_get_tool_definitions(
...
+        invocation.output_type = _output_type_from_mime_type(
+            _get_field(request, "response_mime_type")
+        )
+
+
 def _start_interactions_invocation(
     telemetry_handler: TelemetryHandler,
     instance: InteractionsResource | AsyncInteractionsResource,
-    kwargs: dict[str, Any],
+    request: object,
```
lmolkova: is there a specific type we can use instead of object?

JacksonWeber:
> Sorry overlooked one small thing, should we add assertions to verify these attributes to vcr tests and conformance test suite?

Added assertions to the sync/async VCR tests and the conformance scenario.

instrumentation/opentelemetry-instrumentation-google-genai/src/opentelemetry/instrumentation/google_genai/interactions.py
JacksonWeber: We convert the value before calling the SDK, so a huge integer could make our instrumentation throw first. This guard just skips recording it and lets the SDK handle the original value. Although it's definitely an unlikely edge case.

instrumentation/opentelemetry-instrumentation-google-genai/src/opentelemetry/instrumentation/google_genai/interactions.py
JacksonWeber: Moved the existing conversion checks into `_coerce_float` and replaced the loop with explicit `invocation.temperature` and `invocation.top_p` assignments

instrumentation/opentelemetry-instrumentation-google-genai/src/opentelemetry/instrumentation/google_genai/interactions.py
JacksonWeber: Changed this to `Mapping[str, object]` and updated the request helpers to match.

lmolkova (approved):

PR #758, opened 2026-09-21
fix(langchain): support explicit context and async context propagation (#513)

### LINKED ISSUE #513: langchain: spans are not nested — every step becomes its own root span/trace (async), and attach/detach raises "was created in a different Context"

lmolkova:
## What does this change do?

Fixes #513 by adding optional `context` and `attach_to_context` parameters to `TelemetryHandler` and passing parent context down in the LangChain callback handler. In async event loops, ambient context attachment is disabled to avoid cross-context detach errors.

## Any known gaps or concerns

Context propagation to nested calls (such as auto-instrumented HTTP clients or database queries within tools) is not supported when using LangChain async API.


instrumentation/opentelemetry-instrumentation-genai-langchain/src/opentelemetry/instrumentation/genai/langchain/callback_handler.py
```diff
@@ -175,11 +176,24 @@ class OpenTelemetryLangChainCallbackHandler(BaseCallbackHandler):
     A callback handler for LangChain that uses OpenTelemetry to create spans for LLM calls and chains, tools etc,. in future.
     """
 
-    def __init__(self, telemetry_handler: TelemetryHandler) -> None:
+    run_inline: bool = True
```
rads-1996: Setting `run_inline = True` means these callbacks now run directly on the application’s event loop. Since they can do heavier work, including calling a synchronous completion hook, an async LangChain call could block other tasks. Is there a way to fix the context issue without running the entire handler inline?

instrumentation/opentelemetry-instrumentation-genai-langchain/src/opentelemetry/instrumentation/genai/langchain/callback_handler.py
```diff
@@ -175,11 +176,24 @@ class OpenTelemetryLangChainCallbackHandler(BaseCallbackHandler):
...
+        telemetry_handler: TelemetryHandler,
+    ) -> None:
         super().__init__()
         self._telemetry_handler = telemetry_handler
         self._invocation_manager = _InvocationManager()
 
+    @staticmethod
+    def _should_attach_to_context() -> bool:
+        try:
+            asyncio.get_running_loop()
```
rads-1996: A running event loop does not necessarily mean LangChain is using its async API. A synchronous `.invoke()` can also be called from an async application. In that case, this check stops the GenAI span from becoming current, so HTTP or database spans created inside the chain may attach to the outer request span instead.

Could we determine whether LangChain is actually using its async callback path?

lmolkova: addressed in https://github.com/open-telemetry/opentelemetry-python-genai/pull/758/changes/f8078f2d6ee089a3b334f03f32b58f0fc811d584, setting it based on handler type now

instrumentation/opentelemetry-instrumentation-genai-langchain/src/opentelemetry/instrumentation/genai/langchain/callback_handler.py
lmolkova: good catch, thanks, updated in https://github.com/open-telemetry/opentelemetry-python-genai/pull/758/changes/f8078f2d6ee089a3b334f03f32b58f0fc811d584

rads-1996 (approved):
LGTM

PR #702, opened 2026-09-13
Codegen most of utils

lmolkova:

PR #765, opened 2026-09-21
Add a `SuppressedInferenceInvocation` class to handle nested `InferenceInvocation` operations

DylanRussell:
# Description

Use class `SuppressedInferenceInvocation` in place of `InferenceInvocation` when an instrumentation starts an inference invocation that has already been started by a instrumentation higher up in the call stack..

Same logic that is in https://github.com/open-telemetry/opentelemetry-python-genai/pull/663 but uses a new child class to represent inner invocations, makes it more clear what inner invocations are responsible for -- mostly just updating the context.

Fixes # (issue)

## Type of change

# How has this been tested?

Unit tests

# Checklist

- [x] Unit tests added
- [x] Documentation updated


lmolkova (commented):
thanks for putting this together!

I'm thinking about an approach where we use strongly typed context data and consolidate invocations at finish time.

**Pros:**
- Strict typing and explicit fields make precedence and validation much easier to reason about.
- Clean and contained lifecycle: publish on inner finish, enrich on outer finish.
- Plays nicely with Weaver codegen since the data container maps directly to schema models.

**Cons:**
- More manual boilerplate to map fields today (Weaver codegen in PR #702 will generate most of this anyway).
- Adding new semconv attributes requires updating the dataclass until codegen is in place.


Here is a proof of concept:
https://github.com/lmolkova/opentelemetry-python-genai/commit/eb38238722db34b9d07f7c4c34a2e5d59048a1a2

lmk what you think

DylanRussell:
That link doesn't work for me. 

Precedence already seems very clear to me -- outer instrumentation always overrides inner. Inner only wins when outer hasn't set an attribute.

Typing what's in the context seems a little overkill to me, it will probably make it harder for instrumentations outside the repo to use it.

I dont think we gain much by typing it within the repo either -- it seems safe to assume the type and value is valid, our own code is typed already and just set the values. ultimately we just merge the two attribute maps and one map overrides the other.

lmolkova:
sorry for the broken link, here's the right one https://github.com/lmolkova/opentelemetry-python-genai/commit/7dcc59df7cca904aa648a00ca42fde5a1b8fd958

can you help me understand why and how instrumentation outside of this repo would care about internal details of suppression without taking dependency on util? It would need to put a very specific dict with very specific keys under internal context object. It's already tight enough coupling that would not let suppression work reliably if one of the libs does not use util. 

What we gain is the same logic for the same properties. request.model becomes attribute on both - spans and metrics, no need for two different places to know that. We also don't need to copy things over between dicts of unknown nature.

DylanRussell:
Both `adk` and `litellm` have native instrumentation that emit inference spans and will be installed alongside libraries we instrument in this repo that also emit inference spans. 

Asking them to take a new dependency (which requires a newish version of the otel api/sdk) is a bigger ask than just reading / writing to the context which i think is straightforward without utils. Maybe they wouldn't do it in either case but we should verify that..

DylanRussell:
I'll ask someone from adk to take a look

DylanRussell:
Merging the attributes back into the class is nice to do

lmolkova:
> Both `adk` and `litellm` have native instrumentation that emit inference spans and will be installed alongside libraries we instrument in this repo that also emit inference spans.
> 
> Asking them to take a new dependency (which requires a newish version of the otel api/sdk) is a bigger ask than just reading / writing to the context which i think is straightforward without utils. Maybe they wouldn't do it in either case but we should verify that..

in order to participate in suppression, they would need to at least have access to the context key, which needs an implicit or explicit dependency.  They can achieve suppression with minimal level of pain, but participating in merging attributes is an extra that I don't think is achievable without proper dependency on util. Having dictionaries with bespoke structure (even if documented) is implicit dependency with nothing guaranteeing any form of compatibility.

I'd prefer to draw a line here - suppression of inner layers can be done easily, merging is a cherry on top.

lmolkova:
@DylanRussell created two side-by-side prototypes of native instrumentation that interacts with utils

1. using typed object
2. using dicts

https://github.com/open-telemetry/opentelemetry-python-genai/commit/97cadb55a2724331a84cb36cb2a43626dd1f24ec

Either way it's possible and relatively easy to do, but with optional / guarded importing it's more obvious what you do - we can keep the API public and enforce back-compat. With dictionaries everything is stringly type, brittle, and hard to enforce

DylanRussell:
Ok i'm coming around to the dataclass idea.. If we do it this way it'd be cool to have a single typed dataclass on InferenceInvocation that holds all it's properties, and also holds its extra span/metric attributes.. The only thing it wont have is the content capture attributes.. I think those should be in a separate dataclass container.. That way publishing to the context and enriching from the context becomes a 1 line operation..

But the downside is we need to add a lot of boilerplate setter / getters on the class, ex:

```
    @property
    def input_tokens(self) -> int | None:
        return self.data.input_tokens
    @input_tokens.setter
    def input_tokens(self, val: int | None) -> None:
        self.data.input_tokens = val
```

I'd prefer if we also update all the instrumentations to reach into the dataclass to get/set things.

I updated the prototype here to use this approach.. Let me know what you think

